PR #8, opened 2023-08-16
use cray-libsci for lapack explicitly with rocm

becker33:
Fixes issues with amg2023/openmp ats4 and amg2023/rocm ats4


alecbcs (approved):
Looks good to me!

PR #9, opened 2023-08-16
rename workspaces to experiments

becker33:


alecbcs (approved):
Looks good to me. I agree I think this'll help reduce confusion.

PR #12, opened 2023-09-14
Add x86 config

scheibelp:
With this you can do:

```
./bin/benchpark saxpy/openmp x86 <workspace dir>
```


configs/x86/spack.yaml
```diff
@@ -0,0 +1,14 @@
+spack:
+  packages:
+    default-compiler:
+      spack_spec: gcc@12.2.0
+    default-mpi:
+      spack_spec: mvapich2@2.3.7-gcc12
```
becker33, line 6: @scheibelp This is the line causing the problem

configs/x86/variables.yaml
```diff
@@ -0,0 +1,7 @@
+variables:
+  mpi_command: 'mpirun -N {n_nodes} -n {n_ranks}'
+  batch_submit: ''
```
douglasjacobsen, line 3: This should be: `batch_submit: '{execute_experiment}'`

pearce8 (commented):
@alecbcs What CI tests can we stand up and enable?

@rfhaque Can you please test this with AMG2023, and then with RAJAPerf branch?

PR #24, opened 2023-10-17
Add/docsnew

### LINKED ISSUE #21: benchpark steps
All steps within the benchpark process should ultimately function interactively, in batch mode, under CI, and in the cloud.  First, lets iron out the steps for a human user.

1. Checkout benchpark and needed infrastructure
> benchpark setup [--spack=spack_dir] [--ramble=ramble_dir]
- Should do everything in step 1-6 in current "running the benchmark"
- once/infrequently
- interactive/on login node
- if point at system spack, what version is it?  will this cause compatibility issues, or lack latest package descriptions?
- Probably want to be able to pull in benchpark core updates, and corresponding spack and ramble versions.  Not clear about pulling in changes to the config files which may conflict with local ones - but may also be necessary....
- Errors: invalid directories specified, unable to clone -- all should be displayed in a shell interactively.  Need to capture in a file if not running interactively (how to tell the difference?)
- Re-entrant?

2. Browse benchpark to see available options for benchmark/programming_model, system, experiments, modifiers
> benchpark list system, benchpark list benchmark, benchpark list modifiers
- interactive/on login node
- Error: interactive only (will not do this step under CI)
- Re-entrant: yes, should have no side effects

3. (opt) Turn on modifiers, make any other changes ((We need to talk about the policy for what we want committed back - vs. what is a local change, and should these be in the same files?))
- edit ramble.yaml, anything else?  If it is possible to add a command to do this, this could be merged with step 3.
- interactive/on login node
- Error? 
- Re-entrant?  I think so, only the user is making changes here

4. Select an available experiment, set up the workspace, build, and generate run scripts.
> benchpark select_experiment --benchmark=benchmark_name/programming_model --system=system_name --workspace=workspace_dir 
- the user can keep coming back to select different experiments to run
- if there are multiple workloads, how and in which step to select one?
> ramble -D . workspace setup    
- Option to run on interactive/on login node or submit to batch
- Error?  From Spack etc.  To fix, would edit the workspace/configs, until fixed.  Then, would need to back port those changes to benchpark, which may or may not be straightforward (depending if benchpark combines ramble.yaml and other files before passing them to ramble, the mapping may be tricky to figure out).
- Re-entrant: Yes, Ramble will overwrite previously generated files.

5. Run
> ramble -D . on
- usually submits to batch
- Error: scheduler (interactive?), launcher (file), or job error (file): Need to make easy for user to find
- Re-entrant: Yes, but may overwrite the previous run or be confusing which of the files are from which run.
- Are there "clean previous run" type calls the user can/should make before re-running?

6. analyze results
> ramble workspace analyze
- interactive or in batch 
- Error?
- Re-entrant?

7. archive results
> ramble workspace archive [-t] [--upload-url <mirror_url>]
- -t creates a tarball
- interactive or in batch
- Error?
- Re-entrant? probably but might create duplicates or overwrite

Please comment on individual steps.

pearce8:
Refactoring documentation to match the pending interface changes.


docs/2-benchpark-list.rst
```diff
@@ -1,15 +1,15 @@
 ===================
-Benchpark workspace
+Benchpark list
 ===================
+To list all available in Benchpark benchmarks and systems::
```
scheibelp: Suggested reword:

> To list all available benchmarks and systems::

(as of now, IMO the Wording is confusing: "all available in...", all available what?)

docs/1-benchpark-setup.rst
```diff
@@ -0,0 +1,27 @@
+=========
+Benchpark setup
+=========
+
+Python 3.6 and git are required to check out and install Benchpark::
```
scheibelp: (minor) IMO it would be clearer to be slightly more specific:

> Git is needed to clone benchpark, and Python 3.6 is needed to run it

docs/1-benchpark-setup.rst
```diff
@@ -0,0 +1,27 @@
...
+
+  git clone git@github.com:LLNL/benchpark.git $benchpark
```
scheibelp: This assumes you have an env variable called `$benchpark`, I'd recommend taking that argument off the end.

docs/1-benchpark-setup.rst
```diff
@@ -0,0 +1,27 @@
+=========
+Benchpark setup
```
scheibelp: I ended up creating a `benchpark setup` subcommand. Which I thought was desired, but if this is the initial section then it creates confusion: are we talking about setting up benchpark itself, or `benchpark setup` (which is used to generate the benchmarks)?

I think this section is how to start from scratch, so I think "Setting up benchpark"  or "Getting started with benchpark" would be a clearer title.

Also worth noting: it is easy for me to rename the subcommand to something like `benchpark create`, or `benchpark create-workspace`.

docs/1-benchpark-setup.rst
```diff
@@ -0,0 +1,27 @@
...
+
+Python 3.6 and git are required to check out and install Benchpark::
+
+  git clone git@github.com:LLNL/benchpark.git $benchpark  
+  benchpark setup [--spack=spack_dir] [--ramble=ramble_dir]
+
+Benchpark uses the following open source projects for specifying configurations:
+
+* `Ramble <https://github.com/GoogleCloudPlatform/ramble>`_ to specify run configurations
+* `Spack <https://github.com/spack/spack>`_ to specify build configurations
```
scheibelp: This repeats lines 25-30 of the base README.rst

docs/2-benchpark-list.rst
```diff
@@ -1,15 +1,15 @@
...
+  benchpark list
  
 Configuration files are organized as follows:: 
 
   $benchpark 
+  ├── bin
+  │  ├── benchpark
   ├── configs 
   │  ├── ${SYSTEM1} 
   │  │  ├── auxiliary_software_files
```
scheibelp: This listing is repeated in https://github.com/LLNL/benchpark/pull/24/files#diff-dcd35bb8e1b7a8850b7180d78da82906aac0c9703c78aa45b4d73c92c7aa7d81 - I think it might as well be deleted here if we're just talking about out to list out available systems/benchmarks (I don't think the user needs to know the config file layout to select/run benchmarks)

I'd say the same for everything after this as well.

docs/1-benchpark-setup.rst
```diff
@@ -0,0 +1,27 @@
+=========
+Benchpark setup
+=========
+
+Python 3.6 and git are required to check out and install Benchpark::
+
+  git clone git@github.com:LLNL/benchpark.git $benchpark  
+  benchpark setup [--spack=spack_dir] [--ramble=ramble_dir]
```
scheibelp: Benchpark doesn't have these options yet

and this also doesn't describe the required options in the command (which you can get with `benchpark setup --help`).

docs/1-benchpark-setup.rst
```diff
@@ -0,0 +1,27 @@
...
+Benchpark uses the following open source projects for specifying configurations:
+
+* `Ramble <https://github.com/GoogleCloudPlatform/ramble>`_ to specify run configurations
+* `Spack <https://github.com/spack/spack>`_ to specify build configurations
+
+If you already have Spack and/or Ramble on your system and you want to use these
+installs, simply point Benchpark at these installs by passing in their locations.
+
+
+What the ``benchpark setup`` script does for you
```
scheibelp: `benchpark` is a script, but `benchpark setup` is a subcommand, and IMO referring to it as a script is confusing; I think it would be clearer to say:

> What ``benchpark setup`` does for you

docs/1-benchpark-setup.rst
```diff
@@ -0,0 +1,27 @@
...
+-----------------------------------------
+
+``benchpark setup`` installs Spack and Ramble. If you are running into any issues with either,
+please see the detailed installation instructions in those projects' docs.
+
+Spack: https://spack.readthedocs.io/en/latest/getting_started.html#installation 
+
+Ramble: https://googlecloudplatform.github.io/ramble/getting_started.html#installation
```
scheibelp: In all, `benchpark setup ...` does the following:

* clones spack and ramble into a new directory
* assembles a ramble workspace with a configuration in that directory (using the specified benchmark and system)
* Adds the application/package definitions as Ramble/Spack repositories (to the respective Ramble/Spack instances that were created in the first step)
* prints out instructions on how to build/run the benchmarks for that ramble workspace

this section only mentions the first step. Also, the following line is somewhat vague:

> If you are running into any issues with either,
please see the detailed installation instructions in those projects' docs.

presumably, we (the Benchpark developers) should take charge of any failures in Benchpark to call out properly to Ramble/Spack. Overall, I think bringing them up in this context is not useful and they are sufficiently referenced in `README`.

docs/3-opt-edit-experiment.rst
```diff
@@ -0,0 +1,61 @@
...
+--------------
+If you would like to use non-default runtime behaviour on your system,
+such as compiling in batch instead of on the login node, 
+you can manually edit the specification in ``variables.yaml``.
+
+ramble.yaml
+--------------
+If you would like to use modifiers on your system,
+such as using Caliper to measure the performance of your experiments, 
+you can manually edit the specification in ``ramble.yaml``.
```
scheibelp: Note, you might want to have this point to `add-a-system`, since that section also explains what all these files are for (i.e. either delete that file, or link to it here and delete the extra explanation here).

docs/3-opt-edit-experiment.rst
```diff
@@ -0,0 +1,61 @@
...
+
+  └── repo 
+     ├── ${BENCHMARK1} 
+     │  ├── application.py 
+     │  └── package.py 
+     └── repo.yaml 
+
+where ``application.py`` is a Ramble specification for the benchmark,
+``package.py`` is a Spack specification for the benchmark, and
+``repo.yaml`` points at those two files.
```
scheibelp: Likewise, this may want to just link to the (now empty) `add-a-benchmark.rst` section.

also, 

> and ``repo.yaml`` points at those two files.

`repo.yaml` doesn't point at individual files. For Spack, it is an indicator that a directory contains package descriptions and serves as an identifier for the collective set of package descriptions stored in that directory. Ramble seems to have adopted the same concept.

Overall, a user adding a new benchmark would not need to change or think about `repo.yaml`, so I suggest omitting it from this section.

docs/3-opt-edit-experiment.rst
```diff
@@ -0,0 +1,61 @@
...
+you may introduce the following code::
+
+  └── repo 
+     ├── ${BENCHMARK1} 
+     │  ├── application.py 
+     │  └── package.py 
+     └── repo.yaml 
+
+where ``application.py`` is a Ramble specification for the benchmark,
+``package.py`` is a Spack specification for the benchmark, and
```
scheibelp: Suggested reword from:

> ``application.py`` is a Ramble specification for the benchmark,``package.py`` is a Spack specification for the benchmark, and

to

> ``application.py`` is a Ramble specification describing the benchmark (what to do, what to measure, etc.), and ``package.py`` is a Spack specification for how to install the benchmark

and/or add links to appropriate sections in Spack/Ramble.

scheibelp (changes requested):

docs/1-benchpark-setup.rst
pearce8: I would like to propose the following change:

clone benchpark
cd benchpark --|
clone spack        |   --- these 4 lines can be bundled in 
clone ramble      |          "benchpark setup"
source files     --|

Then the user can be in $benchpark and do git pull or fork.
cd spack and do git pull or fork
cd ramble and do git pull or fork

Further changes are needed for the workspace to then look in $benchpark/spack and $benchpark/ramble for the files.  I'd like to avoid giving those paths as parameters to creating the workspace since I'd like to keep users from pointing those at the system spack of who knows what vintage.  Thoughts?

docs/1-benchpark-setup.rst
pearce8: I have talked myself out of these options being a good idea.  Removed.

docs/1-benchpark-setup.rst
pearce8: Cleaned up the duplicated text.  This discussion belongs in step 4.

docs/1-benchpark-setup.rst
scheibelp: > I would like to propose the following change:

Is the goal of this to avoid cloning ramble and Spack so much? I can see the benefit of that, as right now we do it for every benchmark/system combo we create (e.g. once for saxpy/openmp, once for saxpy/cuda, etc.). However:

* Each of these benchmark configs likely requires a full reinstall of all packages (e.g. the `+cuda` version of dependencies), which likely takes the bulk of the time
* The distinct-ramble-and-spack-for-each-benchmark approach allows users to make edits for one benchmark without affecting others (a possible advantage)
  * Admittedly this only applies for dependencies of the benchmark (i.e. we change package.py files in the package repo of a specific spack instance, and that doesn't affect the installation of other benchmark/system combos, but if we edit the `saxpy` package.py in `benchpark/repos`, then that *may* affect other benchmark installations)
  * Given the above point, the change you propose is IMO more consistent

I should note that even if you only want Spack/Ramble to be cloned once by benchpark, I can still have the script do that automatically (i.e. we don't have to tell users to do it manually).

> I'd like to avoid giving those paths as parameters to creating the workspace since I'd like to keep users from pointing those at the system spack of who knows what vintage. 

That's fine by me, although users might want to pull the latest benchpark and point it to an old Spack repo to make use of any reusable packages installed there, so I think they might want that in the future (I think that will be clearer though as we flesh out use cases).

pearce8: Ok, so I think from our discussion, lets do the following:
- The default behavior should be to clone out spack and ramble as ``benchpark setup`` does right now.  Please keep save their directories, so you can put the spack/ramble cache in those directories (i.e., /path/to/spack/.spack_cache) to keep their lifetimes to the lifetime of the clone.
- Optionally, ``benchpark setup`` should take a path to spack and/or ramble clone the user provides.  I think this enables the user to have a single instance of Spack/Ramble they can pass in for all of the experiments.  Since these may be system installs, I am not sure where to put the caches in this case?

@scheibelp could you make the above change to the python script, and point me at it, so i can make sure the docs match?

docs/add-a-benchmark.rst
```diff
@@ -0,0 +1,36 @@
...
+where ``application.py`` is a Ramble specification for the benchmark,
+``package.py`` is a Spack specification for the benchmark, and
+``repo.yaml`` points at those two files.
+
+The user can override the benchmark specifications upstreamed to Spack and/or Ramble
+by providing custom specifications in ``$benchpark/repo``, 
+and pointing Spack and Ramble to these custom specifications instead::
+
+  spack repo add --scope=site ${APP_SOURCE_DIR}/repo 
+  ramble repo add --scope=site ${APP_SOURCE_DIR}/repo
```
scheibelp: Note the benchpark script already does this automatically, I'm not sure why a user would do it themselves (i.e. rather than edit `$benchpark/repo/${BENCHMARK}/...`

docs/add-a-benchmark.rst
```diff
@@ -0,0 +1,36 @@
+==================
+Adding a Benchmark
+==================
+
+**Benchmark specification:** Specifications for building and running a specific application/benchmark, independent of the target system. The following is required for each ``${BENCHMARK}``: 
+
+- ``package.py`` is a Spack specification that defines how to build and install the benchmark.
+- ``application.py`` is a Ramble specification that defines the benchmark input and parameters.
+
+By default, upstreamed benchmark specifications provided in the Spack and Ramble repos will be used.
```
scheibelp: Because benchpark adds the `repo/${BENCHPARK}` directory as a repo to the Ramble/Spack instance, the definitions there will override what is available in Ramble/Spack.

Also, the term "upstreamed" is a bit confusing without any context

docs/add-a-system-config.rst
```diff
@@ -0,0 +1,27 @@
...
+  - Minimally (e.g., for the ``openmp`` experiments), you need to define ``default-compiler`` and ``default-mpi``.  These will be used in ``experiments/ramble.yaml`` files.
+  - Benchpark adds `spack.yaml`` to a `scope <https://googlecloudplatform.github.io/ramble/configuration_files.html#configuration-scopes>`_ that is automatically used by Ramble
+
+2. ``variables.yaml`` defines system-specific launcher and job scheduler.
+
+  - `variables.yaml  <https://googlecloudplatform.github.io/ramble/configuration_files.html#variables-section>`_ is also a Ramble configuration file.
+  - Minimally (e.g., for the ``openmp`` experiments), ``variables.yaml`` provides instructions on how to dispatch batch jobs on the system.
+  - For the ``cuda`` experiments, ``variables.yaml`` should also define ``cuda_version`` to use on the system.
+
+3. ``auxiliary_software_files/compilers.yaml`` and ``auxiliary_software_files/packages.yaml`` are used to describe software installed on the system:
```
scheibelp: I never did this originally, but it's worth explicitly stating these are Spack config files.

docs/5-build-experiment.rst
```diff
@@ -0,0 +1,31 @@
...
+    workspace_root/
+        <benchmark>/
+            <system>/
+                spack/
+                ramble/
+                workspace/
+                    configs/
+                        (everything from source/configs/<system>)
+                        (everything from source/experiments/<benchmark>)
+                    experiments/
```
scheibelp, line 27: Everything above this is repeated from `4-benchpark-setup`

Everything at this line and below isn't generated until you run `ramble on`

docs/3-opt-edit-experiment.rst
```diff
@@ -0,0 +1,51 @@
+==============
```
scheibelp, line 1: This now seems covered by `add-a-benchmark`, `add-a-system-config`, and `add-an-experiment` (those sections cover this layout and explain the files in more detail).

scheibelp (changes requested):

scheibelp (approved):

PR #27, opened 2023-10-26
Corrections to instructions, adding benchpark setup parameter check

pearce8:


bin/benchpark
```diff
@@ -153,15 +187,14 @@ def benchpark_setup_handler(args):
...
-    cd {workspace_dir}
+    cd {workspace_dir}/workspace
 
     . {workspace_dir}/spack/share/spack/setup-env.sh
     . {workspace_dir}/ramble/share/ramble/setup-env.sh
 
     export SPACK_DISABLE_LOCAL_CONFIG=1
 
     ramble -D . workspace setup
-    ramble -D . on
```
scheibelp, line 185: Why remove this? This runs the experiments.

bin/benchpark
```diff
@@ -42,20 +42,50 @@ def benchpark_list(subparsers, actions_dict):
 
 def benchpark_list_handler(args):
     source_dir = source_location()
+
+    print("Benchmarks:")
+    experiments_dir = source_dir / "experiments"
+    for x in os.listdir(experiments_dir):
+        for y in os.listdir(experiments_dir / x):
+            print(f"\t{x}/{y}")
```
scheibelp: I recommend abstracting this into a function that maps benchmark identifiers to paths (since this is now copied in two places, consolidating into one function would be helpful).

(same for system)

bin/benchpark
```diff
@@ -86,10 +116,14 @@ def benchpark_setup_handler(args):
...
-
     source_dir = source_location()
-
     debug_print(source_dir)
+    debug_print(benchmark)
+    valid_benchmark = benchpark_check_benchmark(benchmark)
+    debug_print(system)
+    valid_system = benchpark_check_system(system)
+    if not (valid_benchmark and valid_system):
+        return
```
scheibelp, line 163: I recommend having these functions just raise an exception like `BenchmarkSpecificationException`, and we can catch here if we want to print out something cleaner than a stack trace (this strategy works well within Spack at least, and can accommodate less-structured problems that should result in a termination).

bin/benchpark
```diff
@@ -86,10 +116,14 @@ def benchpark_setup_handler(args):
     benchmark = args.benchmark
     system = args.system
     workspace_root = pathlib.Path(os.path.abspath(args.workspace_root))
-
     source_dir = source_location()
-
     debug_print(source_dir)
+    debug_print(benchmark)
```
scheibelp: I recommend providing more context in the message (I'm guilty of not doing that in the original PR where I did this).

scheibelp (changes requested):
Most changes/corrections/extensions look good. I have some requests. Let me know if they don't seem reasonable.

bin/benchpark
pearce8: I removed it because it is in a separate step in the docs (easier to explain as a separate logical step), and also if people were to paste the whole thing, they'd be invoking the run even in the previous line exited with an error...

scheibelp: Fair enough, although it does give a complete picture to include it.

If this stays deleted, the above line should be amended:

> To build/run these benchmarks, do the following:

(remove "run", possibly indicate that further steps are needed to run them)

bin/benchpark
pearce8: Separated out into a function, but not sure what mapping you were interested in, I just have a list.

bin/benchpark
pearce8: I think you have something specific in mind; rather than me trying to guess, do you mind handling this?

scheibelp (approved):

PR #29, opened 2023-10-27
Symlink configs; portable Spack cache

### LINKED ISSUE #28: Fails on macOS due to /dev/shm
When running on macOS, `ramble workspace setup` fails with

```
PermissionError: [Errno 1] Operation not permitted: '/dev/shm'
```

I assume this is becuase that directory doesn't exist on macOS.
I recall us discussing getting rid of the `/dev/shm` dependency, but I suppose this issue is to help track the status of this.

Thanks,
Nate

scheibelp:
* The original benchpark script used `lndir`. This adds a rudimentary Python-based implementation of that to symlink all the configs (when the user edits a config in their ramble workspace, it will change the git-managed files in benchpark)
* Relocate the Spack miscellaneous cache to be inside the per-benchmark Spack instance (originally it was located in a location shared amongst all Spack instances cloned by benchpark)

TODO:

- [ ] It was expressed as a desire to add a `--spack-location` option to `benchpark setup`, so that we could force using an existing spack instance

### DIFF
--- a/bin/benchpark
+++ b/bin/benchpark
@@ -133,12 +133,13 @@ def benchpark_setup_handler(args):
 
     spack_exe = spack_location / "bin" / "spack"
     ramble_exe = ramble_location / "bin" / "ramble"
+    spack_cache_location = spack_location / "misc-cache"
 
     env = {
         "SPACK_DISABLE_LOCAL_CONFIG": "1"
     }
     run_command(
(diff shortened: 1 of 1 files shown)


pearce8:
I realize you still have a TODO to add a --spack-location option to benchpark setup, so that we could force using an existing spack instance; lets handle this in a later commit.

The /dev/shm is actually preventing progress on several issues, so I am going to go ahead and merge this.

pearce8 (approved):

PR #32, opened 2023-10-27
Changes to get x86 config working on cts1

rfhaque:
@pearce8 The x86 config works on cts1 without an accompanying compilers.yaml and packages.aml. Spack uses the requested compiler if it can discover it on the system. It installs other required packages like openmpi


experiments/amg2023/openmp/ramble.yaml
pearce8, line 1: Why did you need to add "- compilers" on line 43?
          experiments:
              matrices:
                - size_threads:
                  - compilers

configs/x86/spack.yaml
pearce8, line 1: lapack is a package required by amg, but not other benchmarks.  It would be good to not have to have it in the system config.  @becker33 thoughts?

configs/x86/spack.yaml
```diff
@@ -1,6 +1,10 @@
 spack:
   packages:
     default-compiler:
-      spack_spec: gcc@12.2.0
+      spack_spec: gcc
     default-mpi:
       spack_spec: openmpi
+    lapack:
+      spack_spec: intel-oneapi-mkl@2022.1.0
+    mpi-{compilers}:
```
pearce8: Why do we have default-mpi and mpi-{compilers}?

pearce8 (changes requested):

configs/x86/spack.yaml
```diff
@@ -1,6 +1,10 @@
 spack:
   packages:
     default-compiler:
-      spack_spec: gcc@12.2.0
+      spack_spec: gcc
     default-mpi:
       spack_spec: openmpi
+    lapack:
+      spack_spec: intel-oneapi-mkl@2022.1.0
```
pearce8: This should really be:

lapack:
  spack_spec: lapack

pearce8 (approved):

PR #41, opened 2023-10-31
use --disable-progress-bar when calling ramble so tqdm not required

### LINKED ISSUE #39: tqdm required by ramble
`==> Error: Module `tqdm` is not found. Ensure requirements.txt are installed.`

Should we be automatically be installing Ramble's package requirements as part of benchpark's setup or let the user get to the setup step of benchpark before hitting this error?

slabasan:
~installs required packages into user's python environment before running ramble commands~

Instead, use `--disable-progress-bar` when calling ramble to avoid having tqdm be a required dependency.

Fixes #39

### DIFF
--- a/bin/benchpark
+++ b/bin/benchpark
@@ -237,9 +237,9 @@ To build these benchmarks, do the following:
 
     export SPACK_DISABLE_LOCAL_CONFIG=1
 
-    ramble -D . workspace setup
+    ramble -P -D . workspace setup
 
-Further steps are needed to run the experiments (ramble -D . on)
+Further steps are needed to run the experiments (ramble -P -D . on)
 """
     print(instructions)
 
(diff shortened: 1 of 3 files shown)


bin/benchpark
```diff
@@ -226,6 +226,9 @@ def benchpark_setup_handler(args):
     run_command(
         f"{ramble_exe} repo add --scope=site {source_dir}/repo"
     )
+    run_command(
+        f"pip install -r {ramble_location}/requirements.txt"
+    )
```
scheibelp: This would modify the user's Python installation without their knowledge (and may not work if they are using a System installation of Python). I recommend changing that:

* Add `--target` to the `pip` invocation
* Tell the user to update their `PYTHONPATH` in the below instructions that are printed

slabasan: Thanks @scheibelp! I was also leery about udpating a user's python installing without them knowing. @rfhaque had an alternative suggestion to add a flag to our ramble calls that disables the progress bar.

pearce8 (approved):

PR #46, opened 2023-11-01
Updates to docs and README.rst

pearce8:


slabasan:
@pearce8 The docs renders a little odd now with the URL in the title, can we move it to a sentence in the section? Something like:
`You can find detailed documentation in [ReadtheDocs](https://software.llnl.gov/benchpark).`

Rendered page here:
Screenshot 2023-11-01 at 9 18 31 PM

bin/benchpark
```diff
@@ -139,6 +139,12 @@ def benchpark_setup(subparsers, actions_dict):
     actions_dict["setup"] = benchpark_setup_handler
 
 
+def benchpark_version(subparsers, actions_dict):
```
slabasan: This `benchpark_version` function should go away after merging with develop.

lib/benchpark/benchpark/__init__.py
slabasan, line 1: This file should also disappear with a merge with develop

bin/benchpark
```diff
@@ -260,9 +260,7 @@ To build these benchmarks, do the following:
 
     export SPACK_DISABLE_LOCAL_CONFIG=1
 
-    ramble -P -D . workspace setup
-
-Further steps are needed to run the experiments (ramble -P -D . on)
+Further steps are needed to build the experiments (ramble -P -D . workspace setup) and run them (ramble -P -D . on)
```
scheibelp, line 263: (no change needed) I think I like this organization since everything above this line is something you'd run once (e.g. when you create a new shell) and you don't have to do it again each time you run `ramble workspace setup` and/or `ramble on`.

docs/5-build-experiment.rst
```diff
@@ -31,3 +26,7 @@ which will build the source code and set up the following workspace directory st
                                 <problem>/
                                     <benchmark>_<ProgrammingModel>_<problem>
                                         execute_experiment
+
+
+If you change any build instructions, rerun the ``ramble -P -D . workspace setup`` command
+to re-setup the workspace and rebuild the binaries as necessary.
```
scheibelp: Perhaps reference the FAQ added in this PR? (as it is, "build instructions" could mean a few different things)

scheibelp (approved):
I have one suggestion about referencing the FAQ somewhere, but this looks good (could merge without that change)

docs/5-build-experiment.rst
pearce8: yes, good catch.

PR #50, opened 2023-11-03
WIP: sketch out source code architecture for apidoc

slabasan:
** **Not ready for merge** ** 

* initial commit: add benchpark version -- Benchpark driver needs to be python to import modules. Spack gets around this by having a top-level bash script that calls main.py which has the driver code.

Screenshot 2024-04-15 at 4 01 16 PM


slabasan:
Think we converged on keeping the `benchpark help` docs page updated (need to update the condition for when docs are re-rendered) instead of adding API docs (too many tokens for RTD to search on). This PR can be closed.

PR #60, opened 2023-11-28
Set of experiments

pearce8:
Documentation for using the same Spack clone and Ramble clone for a set of experiments.


bin/benchpark
```diff
@@ -261,8 +263,6 @@ def benchpark_setup_handler(args):
     instructions = f"""\
 To complete the benchpark setup, do the following:
 
-    cd {workspace_dir}/workspace
-
```
scheibelp, line 265: Without this, the later suggestions `ramble -P -D . workspace setup` and `ramble -P -D . on` are possibly confusing (they wouldnt work until you `cd` into the benchmark workspace dir): you could update them (e.g. `ramble -P -D {ramble_workspace_dir} workspace setup`)

docs/5-build-experiment.rst
```diff
@@ -12,11 +12,11 @@ The next step is setting up the Ramble workspace and building the code::
 Ramble will build the source code and set up the following workspace directory structure::
 
     workspace_root/
+        ramble/
+        spack/
```
scheibelp, line 16: (no change needed) This change by itself looks good, although I see from this review we copy the comment showing the tree in at least 4 places. There should probably only be one (and the rest should refer to that function or doc).

docs/set-of-experiments.rst
```diff
@@ -0,0 +1,36 @@
+==============
+Working with a set of experiments
+==============
+
+To keep the source code consistent while working with 
+a set of ``experiments`` (several ``benchmark/ProgrammingModels`` x ``systems``),
+one can consider working with a single ``Benchpark workspace``, which will ensure the same
+clone of Spack, and the same clone of Ramble, will be used for all of the ``experiments`` in 
+this set.  To work with a set of ``experiments`` based on a single clone of Spack and 
+a single clone of Ramble, set up your experiments in the same /output/path/to/workspace_root::
```
scheibelp: We might want to open with why a user would want to keep the same spack/ramble for multiple experiments

```
You may want to use the same workspace directory when setting up multiple experiments:

* Benchpark only stores one copy of Spack/Ramble per workspace, and does not repeatedly
  download them when reusing a workspace
* Since each experiment in the workspace shares the Spack/Ramble instances, then edits you
  you make to Spack packages will apply to all experiments in the workspace
```

bin/benchpark
```diff
@@ -174,22 +176,22 @@ def symlink_tree(src, dst):
 def benchpark_setup_handler(args):
     """
     workspace_root/
```
scheibelp: We should think of a different names than `workspace_root/` and `workspace/` (or at least, one of those two should change): for example we could have `experiments_basedir/` as the root and `experiment_workspace/` for each Ramble workspace.

That would improve our docs in general, including possibly simplifying the wording of docs added in this PR.

bin/benchpark
```diff
@@ -125,17 +125,19 @@ def benchpark_check_system(arg_str):
...
+    )
+    create_parser.add_argument(
+        "system", type=str, help="The system on which to run the experiment"
     )
-    create_parser.add_argument("system", type=str, help="The system we are on")
     create_parser.add_argument(
         "workspace_root",
         type=str,
-        help="Where to install packages, store results, etc. (should be empty/nonexistent)",
+        help="Where to install packages, store results, etc. (could be empty/nonexistent)",
```
scheibelp: We might want to say here that Benchpark expects to manage this directory (I think the original intent of the comment in "()" was that users should not pick a `workspace_root` that exists and contains files not related to `benchpark`: it should be empty/nonexistent the first time they run `benchpark`).

scheibelp (changes requested):

bin/benchpark
pearce8: The cd is not required for the spack and ramble commands that have the full path.

But, I was wondering what we should have here to help the user with the next steps.  ramble_workspace_dir will be set to a new value with each experiment the user adds to this benchpark_workspace (which we might want to always call benchpark_workspace for clarity), so when these instructions are displayed, they will point at the ramble_workspace_dir for the last experiment the user added.  Any suggestions on how to communicate that:
- There is a ramble_workspace_dir per experiment
- That's what the user should cd into to work with the specific experiment

docs/5-build-experiment.rst
pearce8: Agreed.  I added workspace_directory_structure.rst with the directory structure and tried to include it from 5-build-experiment.rst but the included text is not showing up in preview - @alecbcs how to check if once the docs are built, this will work as intended?

bin/benchpark
pearce8: Yes, I am also convinced that this is confusing.

Ramble uses "workspace" in its docs, so I think we should keep using the term "workspace" to mean Ramble workspace for a single experiment.  It will be nested within the experiment triple, so I am not sure we need to add experiment_.

"Benchpark workspace" or "workspace_root" then need to not have the word "workspace" in it.  I like your suggestion of experiments_basedir, sounds like experiments_root is what we are settling on here.

docs/set-of-experiments.rst
pearce8: Thank you for rewriting this in a much more clear way.  Now we just need to settle on your next point, and circle back to this.

bin/benchpark
pearce8: Agreed.

scheibelp (approved):

PR #61, opened 2023-11-29
Adding GROMACS

pearce8:
Working to add a new problem definition for GROMACS.  Will upstream to Ramble at a later point.

PR #62, opened 2023-11-30
Add gromacs openmp config

rfhaque:
1. Added the gromacs openmp config
2. The config works with for the systems: cts1 and x86 (on cts1) 
3. The config fails for ats4 due to a cmake error within gromacs (this probably needs to be fixed within the gromacs build system)
4. Running gromacs with hwloc on results in a runtime error on cts1 (hence the code is built with hwloc off)

### DIFF
--- a/configs/ats4/auxiliary_software_files/packages.yaml
+++ b/configs/ats4/auxiliary_software_files/packages.yaml
@@ -189,6 +189,11 @@ packages:
     buildable: false
   hypre:
     variants: amdgpu_target=gfx90a
+  hwloc:
+    externals:
+    - spec: hwloc@2.9.1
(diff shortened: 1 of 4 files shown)


pearce8 (approved):

pszi1ard:
> The config fails for ats4 due to a cmake error within gromacs (this probably needs to be fixed within the gromacs build system)

Let's try to repro this with the latest GROMACS release, see my comment on #61.

> Running gromacs with hwloc on results in a runtime error on cts1 (hence the code is built with hwloc off)

Same as above; also note that if CTS1 is [this](https://www.top500.org/system/178971/), hwloc is not critical since we have the internal hardware topology detection as a fallback.

PR #74, opened 2023-12-15
Adding Copyright, enforce license headers on all files

pearce8:
Adding the top level copyright.  Which of the files should have the copyright at the top?
- package.py that our team develops in Benchpark?  What happens to copyright when we push to Spack builtins?
- application.py that our team develops in Benchpark?  What happens to copyright when we push to Ramble builtins?
- system configs?


pearce8:
@slabasan Can you please work through a PR to check the copyright in each file?

.github/workflows/scripts/license.py
```diff
@@ -0,0 +1,184 @@
...
+                # The first line of the license contains the copyright date.
+                # We allow it to be out of date but print a warning if it is
+                # out of date.
+                if i == 0:
+                    if not re.search(strict_date, line):
+                        print("{}: copyright date mismatch".format(path))
+                found.append(i)
+
+    if len(found) == len(license_line_regexes) and found == list(sorted(found)):
+        return
```
scheibelp: This is possibly a nitpick but you might want `found` to be a `set`, since otherwise it looks like you could have `found == [0, 0, 2]` e.g. I can get the license check to pass if I modify `bin/benchpark` like so:

```
#!/usr/bin/env python3
#
# Copyright 2023 Lawrence Livermore National Security, LLC and other
# Copyright 2023 Lawrence Livermore National Security, LLC and other
#

import argparse
import subprocess
import pathlib
...
```

I think only the "middle" line can be omitted in this way (given the date/SPDX checks).

scheibelp (changes requested):
Overall this seems good: removing licenses in various files leads to a failure in the verifier (as desired), and modifying the license identifier also causes a failure (as desired). Constraints on date also appear to work (although error message isn't clear in that case)

1. (minor issue) if I change `configs/ats4/variables.yaml` like so:

```
-# Copyright 2023 Lawrence Livermore National Security, LLC and other
+# Copyright 2021 Lawrence Livermore National Security, LLC and other
```

then the error message is unclear IMO

```
/Users/scheibel1/Desktop/benchmarking/benchpark/configs/ats4/variables.yaml: SPDX license identifier missing
1 improperly licensed files
 files with wrong SPDX-License-Identifier:   0
 files not containing expected license:      1
```

2. (I assume this is fine) if I change `configs/ats4/variables.yaml ` like so

```
-# SPDX-License-Identifier: Apache-2.0
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
```

then I get a failure (which I assume is acceptable, even though mixing these licenses is fine, we don't want to give flexibility on that).

.github/workflows/scripts/license.py
```diff
@@ -0,0 +1,184 @@
+# Copyright 2023 Lawrence Livermore National Security, LLC and other
+# Benchpark Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: Apache-2.0
```
alecbcs:
```suggestion
#!/usr/bin/env python3
#
# Copyright 2023 Lawrence Livermore National Security, LLC and other
# Benchpark Project Developers. See the top-level COPYRIGHT file for details.
#
# SPDX-License-Identifier: Apache-2.0
```

I'd add a shbang line here and modify the name of the file from `license.py` to `license`. Super minor but I'd also go with `.github/workflows/bin/` instead of `scripts/` just to capitalize on well known standards.

.github/workflows/scripts/license.py
slabasan: Thanks @scheibelp for the feedback, I think I'll table this one for a more robust check in a future iteration. Technically, the license header isn't wrong in this case, but certainly has extra lines we don't need.

.github/workflows/scripts/license.py
slabasan: File renamed and scripts renamed to bin, thanks @alecbcs!

PR #76, opened 2023-12-16
Add HPCC mpi-only to benchpark

rfhaque:
Tested on cts1


pearce8:
@rfhaque Please test on its-1 using the x86 config.

configs/cts1/auxiliary_software_files/packages.yaml
pearce8, line 1: Please remove fftw as a handle expected from a system.  We should use Spack to build it instead.

experiments/hpcc/mpi-only/ramble.yaml
pearce8, line 1: environments:
       hpcc:
         packages:


Should not have fftw.

pearce8 (changes requested):

configs/cts1/auxiliary_software_files/packages.yaml
rfhaque: @pearce8 Removed fftw

rfhaque:
> @rfhaque Please test on its-1 using the x86 config.

Tested on cts1 with the x86 config

pearce8 (approved):

PR #80, opened 2023-12-19
Patch cuda arch

pearce8:
Forcing the cuda_arch check to be recursive (apply to all dependencies).

@rfhaque Please test on ats2.


rfhaque (commented):
@pearce8 amg2023, raja-perf and saxy successfully tested on ats2 with this change

rfhaque (approved):

PR #81, opened 2023-12-19
Patch/system names

pearce8:
Giving systems descriptive names, mirroring the Top500 naming.

Defining system_definition.yaml per system.  In a future PR:
- Need a system_definition template
- Make system definitions searchable
- Script to make a table of systems defined for docs


configs/AWS-Hpc7a-EPYC4-EFA/system_definition.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+system_definition:
+  name: AWS-Hpc7a-EPYC4-EFA
+  manufacturer:
+    vendor: AWS
+    name: ParallelCluster3.7.2-Hpc7a
+  processor:
+    vendor: AMD
+    name: EPYC-Zen4
+    ISA: x86_64
```
becker33: I don't think this is specific enough for ISA -- `zen3` and `skylake` do not share an instruction set, even though both are derived from `x86_64`.

What about 

```suggestion
    ISA_family: x86_64
    ISA: zen4
```

configs/AWS-Hpc7a-EPYC4-EFA/system_definition.yaml
```diff
@@ -0,0 +1,28 @@
...
+  interconnect:
+    vendor: AWS
+    name: EFA
+  OS:
+    name: Amazon-Linux-2
+  system-tested:
+    owner: AWS
+    name: ""
+    installation-year: ""
+    description: [aws/hpc7a](https://aws.amazon.com/ec2/instance-types/hpc7a)
```
becker33: Ideally we could get a github gist here with the yaml description for the pcluster setup -- that way it would be reproducible for the cloud workflows.

configs/HPECray-EPYC3-MI250X-Slingshot/system_definition.yaml
```diff
@@ -0,0 +1,28 @@
...
+    vendor: HPE-Cray
+    name: EX235a
+  processor:
+    vendor: AMD
+    name: EPYC-Zen3
+    ISA: x86_64
+  accelerator:
+    vendor: AMD
+    name: MI250X
+    GPU_ISA: gfx90a
```
becker33: Is `GPU_ISA` the right word for these? I would've thought something like `architecture`.

configs/HPECray-EPYC3-MI250X-Slingshot/system_definition.yaml
```diff
@@ -0,0 +1,28 @@
...
+  interconnect:
+    vendor: HPE-Cray
+    name: Slingshot-11
+  OS:
+    name: HPE-Cray-OS
```
becker33: This is not accurate, we ran on TOSS. There's nothing in the config that makes it arch specific though.

configs/IBM-POWER9-V100-Infiniband/system_definition.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+system_definition:
+  name: IBM-POWER9-V100-Infiniband
+  manufacturer:
+    vendor: IBM
+    name: AC922
+  processor:
+    vendor: IBM
+    name: POWER9
+    ISA: POWER-3.0B
```
becker33:
```suggestion
    ISA: power9
    ISA_family: ppc64le
```

This keeps our descriptions in line with what you get from https://github.com/archspec/archspec.git

configs/Penguin-XeonPlatinum-OmniPath/system_definition.yaml
```diff
@@ -0,0 +1,28 @@
...
+    vendor: PenguinComputing
+    name: RelionCluster
+  processor:
+    vendor: Intel
+    name: XeonPlatinum924248C
+    ISA: x86_64
+  accelerator:
+    vendor: ""
+    name: ""
+    GPU_ISA: ""
```
becker33: Is this better than having `accelerator: None`?

I know the example is contrived, but it would distinguish more clearly between no accelerator and a weirdly named one.

configs/x86/system_definition.yaml
```diff
@@ -0,0 +1,28 @@
+# Copyright 2023 Lawrence Livermore National Security, LLC and other
+# Benchpark Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+system_definition:
+  name: x86
```
becker33:
```suggestion
  name: x86_64
```

configs/x86/system_definition.yaml
```diff
@@ -0,0 +1,28 @@
...
+  interconnect:
+    vendor: ""
+    name: ""
+  OS:
+    name: ""
+  system-tested:
+    owner: ""
+    name: ""
+    installation-year: ""
+    description: ""
```
becker33: Is the idea this eventually gets the information for the cloud node we use to test this on?

becker33 (changes requested):
Two big picture comments:

1. I think the names for these systems should include the canonical name for the system on which they were tested. E.g.

`tioga-HPECray-EPICzen3-...`

2. I think we should document the order of fields in this naming scheme.

configs/AWS-Hpc7a-EPYC4-EFA/system_definition.yaml
pearce8: yes!  A cluster setup yaml is definitely something we want to keep.  Do you think we would ask people to add the yaml file in the system definition directory?  Or do you mean something else by GitHub gist?  I do think the system_definition.yaml should point at it - suggestions for how to name the field?

pearce8:
> Two big picture comments:
> 
> 1. I think the names for these systems should include the canonical name for the system on which they were tested. E.g.
> 
> `tioga-HPECray-EPICzen3-...`

I worry about this, because I explicitly want the Frontier and Lumi users to be able to find this.  So, I only name tioga in the system-tested.  

I do wonder if we should allow multiple system-tested (i.e., once they use it on Frontier, and if they didn't need to change it, would add Frontier to the list.

I am also wondering if we need a field to list the systems with this architecture that we know - explicitly saying that Frontier and Lumi have the same hardware, and make that searchable?

While I think this file addresses the hardware questions, I do worry about the software stack.  What do we do if Tioga defaults to a different compiler version than Frontier? or, multiple sites have CI set up, and one of the sites moves up the compiler version?

> 2. I think we should document the order of fields in this naming scheme.

Yes!  We badly need a schema.  And I do want to make these readable in the docs.  So, do you think I should add comments in each of these files, or do the comments go in the docs - and eventually into the schema?

configs/HPECray-EPYC3-MI250X-Slingshot/system_definition.yaml
pearce8: Please see slack for discussion.

configs/HPECray-EPYC3-MI250X-Slingshot/system_definition.yaml
pearce8: The intent here is the "normal" machine definition others recognize, and that list is mostly what you see on the top500.  Technically even tioga came with Cray OS; I didn't mention the OS in the "system-tested".

Do you think we should drop the OS info?  I think it is nice to have it for completeness.  And, maybe we need to indicate the full tested configuration.  While it is nice to have a short system-tested description here, I do worry that it is coupled with the hardware description - where what we really need is the full spec of what was tested, and the date it was tested - because if i tested on tioga a year ago but have since updated these scripts and the machine, this info is obsolete/useless/misleading?

configs/IBM-POWER9-V100-Infiniband/system_definition.yaml
pearce8: Yes, thanks.  Is there a full list of the strings I can look up for capitalization?

But oh no, you are breaking my non-existent schema lol.  ISA then ISA_family in one file, switched order in the next file.  This will get messy quick if we don't write that schema soon.

configs/Penguin-XeonPlatinum-OmniPath/system_definition.yaml
pearce8: My thought was that I will want to make a table of these system descriptions, and each line is a row - which is why I kept all of the rows.  Is that reasonable?

I am happy to have None instead of the "" if that makes it cleaner.

configs/x86/system_definition.yaml
pearce8: I thought about it.  I would need to rename the directory as well - and chance the CI (since that's the only system we are currently testing).

configs/x86/system_definition.yaml
pearce8: probably?  A cloud node could have the specific cluster type (which defines hardware) and the specific cluster build (which defines software stack), but I am not sure which goes here since for other clusters we only define the hardware in this file (and the OS unless we remove it).

configs/HPECray-EPYC3-MI250X-Slingshot/system_definition.yaml
becker33: I thought the idea was that these would be continually tested, so listing the system it is tested on on an ongoing basis would still be meaningful?

configs/Penguin-XeonPlatinum-OmniPath/system_definition.yaml
```diff
@@ -0,0 +1,32 @@
...
+system_definition:
+  name: Penguin-XeonPlatinum-OmniPath
+  manufacturer:
+    vendor: PenguinComputing
+    name: RelionCluster
+  processor:
+    vendor: Intel
+    name: XeonPlatinum924248C
+    ISA: x86_64
+    uArch: IceLake-SP
```
becker33:
```suggestion
    uArch: icelake
```
to match to `archspec` monikers instead of whatever marketing things they come up with.

configs/HPECray-EPYC3-MI250X-Slingshot/system_definition.yaml
becker33: I would be fine with dropping the `OS` information -- it's really only relevant to specs that we use from externals, and those are also recorded. Since different sites can install different OSs on the same system or different software from the OS, leaving it out might be better

docs/add-a-system-config.rst
```diff
@@ -3,14 +3,60 @@ Adding a System Configuration
...
 ``benchpark/configs`` contains a directory for each system specified in Benchpark.
-If the software stack on your system is unlike the available configurations, 
+If your system is unlike the available configurations,
 you can add a new directory with a name which identifies the system.
 
-The following is required for each given system ``benchpark/configs/${SYSTEM}``:
+The naming convention for the systems is as following::
 
-1. ``spack.yaml`` defines default compiler and package names Spack should
+  Integrator-(opt)ClusterType-CPU-(opt)GPU-Network
```
becker33:
```suggestion
  (opt)Integrator-(opt)ClusterType-CPU-(opt)GPU-Network
```
Unless we want to add a "generic" integrator to the generic ones.

docs/add-a-system-config.rst
```diff
@@ -3,14 +3,60 @@ Adding a System Configuration
...
+
+Benchpark has definitions for the following systems:
+- AWS-Hpc7a-EPYC4-EFA
+- HPECray-EPYC3-MI250X-Slingshot	(Frontier, Lumi, Tioga)
+- IBM-POWER9-V100-Infiniband	        (Sierra, Summit)
```
becker33: The config we have would not work for summit

docs/add-a-system-config.rst
```diff
@@ -3,14 +3,60 @@ Adding a System Configuration
...
+- Penguin-XeonPlatinum-OmniPath
+- x86_64                                (generic x86 CPU only platform)
+
+The following files are required for each system ``benchpark/configs/${SYSTEM}``:
+
+1. ``system_definition.yaml``
```
becker33: I think we should add a little bit more description here about what it is and how to fill it out

configs/HPECray-EPYC3-MI250X-Slingshot/system_definition.yaml
pearce8: Continually tested: yes, that would be ideal, and hopefully we figure out a way to pull in "last tested on" from the CI.

Ok, I will remove the OS.

configs/AWS-Hpc7a-EPYC4-EFA/system_definition.yaml
pearce8: resolved.

pearce8 (commented):
I think we've either resolved, or decided on resolving issues in a future PR.

configs/x86_64/system_definition.yaml
```diff
@@ -0,0 +1,30 @@
...
+  interconnect:
+    vendor: Any
+    name: Any
+  system-tested:
+    owner: Any
+    name: Any
+    installation-year: Any
+    description: Any
+  top500-system-instances:
+    - Many
```
becker33, line 30: 😆

configs/HPECray-zen3-MI250X-Slingshot/auxiliary_software_files/packages.yaml
```diff
@@ -1,6 +1,7 @@
 packages:
   all:
-    require: target=x86_64
+    require: target=x86_64:
```
becker33, line 8: For a future PR: this should probably be `target=:zen3`

.github/workflows/run.yml
```diff
@@ -17,7 +17,7 @@ jobs:
 
       - name: Build Saxpy Workspace
         run: |
-          ./bin/benchpark setup saxpy/openmp x86 workspace/
+          ./bin/benchpark setup saxpy/openmp x86_64 workspace/
```
becker33, line 20: Are we going to have to specify these entire names from the CLI? We're going to need some sort of tab completion eventually.

docs/4-benchpark-setup.rst
```diff
@@ -10,7 +10,7 @@ Also choose a directory for your experiment::
 where:
 
 - ``benchmark/ProgrammingModel``: amg2023/openmp | amg2023/cuda | saxpy/openmp (available choices in ``benchpark/experiments``)
-- ``system``: ats2 | ats4 | cts1 (available choices in ``benchpark/configs``)
+- ``system``: x86_64 | HPECray-EPYC3-MI250X-Slingshot | AWS-Hpc7a-EPYC4-EFA (available choices in ``benchpark/configs``)
```
becker33: This needs to update to match the other changes.

docs/add-a-system-config.rst
```diff
@@ -3,14 +3,63 @@ Adding a System Configuration
...
-If the software stack on your system is unlike the available configurations, 
+If your system is unlike the available configurations,
 you can add a new directory with a name which identifies the system.
 
-The following is required for each given system ``benchpark/configs/${SYSTEM}``:
+The naming convention for the systems is as following::
 
-1. ``spack.yaml`` defines default compiler and package names Spack should
+  (opt)Integrator-(opt)ClusterType-CPU-(opt)GPU-Network
+  [INTEGRATOR]-MICROARCHITECTURE[-GPU][-NETWORK]
```
becker33:
```suggestion
  (opt)Integrator-(opt)ClusterType-CPU-(opt)GPU-Network
  [INTEGRATOR]-MICROARCHITECTURE[-GPU][-NETWORK]
```
We should pick which syntax to use to describe this and remove `ClusterType` if we choose the first one

docs/add-a-system-config.rst
```diff
@@ -3,14 +3,63 @@ Adding a System Configuration
...
+
+where::
+
+  COMPANY[_PRODUCTNAME][...]
```
becker33:
```suggestion
  INTEGRATOR = COMPANY[_PRODUCTNAME][...]
```

docs/4-benchpark-setup.rst
pearce8: yeah, some of these need to be automated

.github/workflows/run.yml
pearce8: Yes.  I also hope making suites will help.

becker33 (approved):

PR #83, opened 2023-12-19
Adding Kripke to Benchpark

pearce8:
Adding Kripke, problem definition, and eventually Caliper instrumentation to Kripke.


repo/kripke/package.py
```diff
@@ -0,0 +1,125 @@
...
+    depends_on("adiak@0.4:", when="+caliper")
+    
+    aligned_versions = ["2024.02"]
+    constrain_chai = list(f"^chai@{v}" for v in aligned_versions)
+    requires(*aligned_versions, policy="one_of", msg="Pick a chai version with proper alignment w/RAJA & Umpire")
+    
+    for v in aligned_versions:
+      depends_on(f"raja@{v}~exercises~examples", when=f"^chai@{v}")
+      depends_on(f"umpire@{v}~examples", when=f"^chai@{v}")
+      depends_on(f"chai@{v}~examples+raja", when=f"^chai@{v}")
```
pearce8, line 65: @scheibelp is this how I would specify chai in here?  I do want to build chai ~examples+raja.

Also, in your example you had constrain chai - but just using raja and umpire in the loop.  Is there a reason one was chosen?

Is the list in constrain chai then getting the list of minor versions as well?  I don't understand how the minor versions would work here.

scheibelp: > Also, in your example you had constrain chai - but just using raja and umpire in the loop. Is there a reason one was chosen?

Speaking loosely, these are edge constraints, and two edges covers three nodes

`depends_on(f"raja@{v}~exercises~examples", when=f"^chai@{v}")` for example imposes a version constraint on `raja` based on `chai`.

The third constraint will have no effect (it won't cause a problem, but won't change how anything concretizes).

Note that if you don't force `chai` to be one of the versions in aligned versions, Spack will feel free to pick other versions of `chai` that do not have version-locking constraints you set up here, so

```
constrain_chai = list(f"^chai@{v}" for v in aligned_versions)
requires(*constrain_chai, policy="one_of", msg="Pick a chai version with proper alignment w/RAJA & Umpire")
```

(I'll note there was a typo where I mentioned this to you originally, and the example directly above fixes that)

configs/LLNL-Sierra-IBM-power9-V100-Infiniband/spack.yaml
```diff
@@ -6,9 +6,9 @@
 spack:
   packages:
     default-compiler:
-      spack_spec: xl@16.1.1-2022.08.19-cuda{default_cuda_version}
+      spack_spec: xl@16.1.1-2023.06.28-cuda-{default_cuda_version}-gcc-11.2.1
     default-mpi:
-      spack_spec: spectrum-mpi@2022.08.19
+      spack_spec: spectrum-mpi@2023.06.28-cuda-{default_cuda_version}-gcc-11.2.1
```
rfhaque, line 11: @pearce8 I updated the default compiler on lassen. It'll be good to test at least a few benchmarks with this compiler

rfhaque (approved):

configs/LLNL-Sierra-IBM-power9-V100-Infiniband/spack.yaml
pearce8: yes, please test all that have cuda variants.

PR #89, opened 2023-12-28
AMG without magic compilers

pearce8:
With current AMG package definition and without magic wrappers, we are somehow linking to Fortran and getting the following error:

    configure: error: linking to Fortran libraries from C fails

Possible solutions:
- Disable fortran in amg2023 as in this commit
- However, the likely fix needed is figuring out what library magic was linking to that we need to link to explicitly when not using magic compilers.


rfhaque:
Subsumed by https://github.com/LLNL/benchpark/pull/270

PR #90, opened 2023-12-28
Feature caliper modifier saxpy

rfhaque:
@pearce8 caliper modifier for saxpy. Tested on topdown and cuda modes on quartz and lassen respectively


pearce8 (changes requested):
We need to work towards not requiring libunwind.  This may be a Caliper package change - or a change to how we indicate to build Caliper.

daboehme:
> We need to work towards not requiring libunwind. This may be a Caliper package change - or a change to how we indicate to build Caliper.

Hi @rfhaque - you can either turn off the libunwind option in the Caliper spack spec to build without libunwind, or point spack to the system install of libunwind on Lassen.

repo/saxpy/saxpy.cc
```diff
@@ -110,28 +119,73 @@ int main(int argc, char** argv) {
...
+      adiak_init(&comm);
+      adiak_user();
+      adiak_uid();
+      adiak_launchdate();
+      adiak_executable();
+      adiak_executablepath();
+      adiak_libraries();
+      adiak_cmdline();
+      adiak_hostname();
+      adiak_clustername();
```
daboehme, line 138: If you're using Adiak v0.4.0+ you can use `adiak_collect_all()` to get all the Adiak built-in metadata

repo/saxpy/saxpy.cc
```diff
@@ -110,28 +119,73 @@ int main(int argc, char** argv) {
...
+    #endif
 
     std::cout << "Kernel done (" << rank << "): " << N << std::endl;
 
     delete[] h_x;
     delete[] h_y;
     delete[] h_r;
 
+    #ifdef USE_CALIPER
+      CALI_MARK_END("problem");
```
daboehme, line 179: I guess you can merge this `#ifdef` block into the next. Also looks like `problem` is essentially the same region as `main` so it might be redundant.

daboehme (commented):
Added a few comments in the code. Otherwise looks OK to me.

PR #104, opened 2024-01-06
Run `isort` and `codespell` Against Repo & Add to CI

alecbcs:
- Add `isort` to list of style checks to keep files in line with Spack formatting requirements. (This'll save us later on when we want to upstream stuff to Spack/Ramble.)
- Add `codespell` to style checks to alert us to (likely my) typos.

### DIFF
--- a/bin/benchpark
+++ b/bin/benchpark
@@ -6,11 +6,11 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import argparse
-import subprocess
-import pathlib
 import os
-import shutil
+import pathlib
 import shlex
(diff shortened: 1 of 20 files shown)


bin/benchpark
```diff
@@ -6,11 +6,11 @@
...
 
 import argparse
-import subprocess
-import pathlib
 import os
-import shutil
+import pathlib
 import shlex
+import shutil
+import subprocess
```
pearce8, line 13: What uses subprocess?

pearce8 (approved):

PR #109, opened 2024-01-10
Create script to bundle 3 instructions into 1

scheibelp:
After running `benchpark setup`, it usually tells the user to start with:

```
$ benchpark setup amg2023/rocm HPECray-zen3-MI250X-Slingshot `pwd`/w1
Setting up configs for Ramble workspace
...
To complete the benchpark setup, do the following:

    . .../w1/spack/share/spack/setup-env.sh
    . .../w1/ramble/share/ramble/setup-env.sh

```

With this PR that reduces to

```
$ benchpark setup amg2023/rocm HPECray-zen3-MI250X-Slingshot `pwd`/w1
Setting up configs for Ramble workspace
...
To complete the benchpark setup, do the following:

    . .../w1/setup.sh
```

### DIFF
--- a/bin/benchpark
+++ b/bin/benchpark
@@ -284,8 +284,12 @@ def benchpark_setup_handler(args):
         )
 
         run_command(f"{ramble_exe} repo add --scope=site {source_dir}/repo")
-        run_command(f"{ramble_exe} config --scope=site add \"config:disable_progress_bar:true\"")
-        run_command(f"{ramble_exe} config --scope=site add \"config:spack:global:args:'-d'\"")
+        run_command(
+            f'{ramble_exe} config --scope=site add "config:disable_progress_bar:true"'
+        )
+        run_command(
+            f"{ramble_exe} config --scope=site add \"config:spack:global:args:'-d'\""
+        )
 
     if not initializer_script.exists():
         with open(initializer_script, "w") as f:
--- a/bin/benchpark
+++ b/bin/benchpark
@@ -258,6 +258,8 @@ def benchpark_setup_handler(args):
     ramble_exe = ramble_location / "bin" / "ramble"
     spack_cache_location = spack_location / "misc-cache"
 
+    initializer_script = experiments_root / "setup.sh"
+
     if not spack_location.exists():
         print(f"Cloning spack into {spack_location}")
         run_command(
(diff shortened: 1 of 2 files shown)


bin/benchpark
```diff
@@ -283,13 +285,27 @@ def benchpark_setup_handler(args):
...
+            f.write(
+                f"""\
+if [ -n "${{_BENCHPARK_INITIALIZED:-}}" ]; then
+    return 0
+fi
+
+. {spack_location}/share/spack/setup-env.sh
+. {ramble_location}/share/ramble/setup-env.sh
+
+export SPACK_DISABLE_LOCAL_CONFIG=1
```
pearce8, line 305: @scheibelp Will disabling local config here make Spack run in Verbose mode when called from Ramble?

scheibelp: No. Although to clarify, do you want this to happen? Or not to happen? Spack has a few notions of verbosity  (`spack -d`, and apparently also `spack --verbose`, though I don't actually think the latter has much effect and `-d` is typically what we use to get info when something goes wrong).

pearce8: What I think we want is to get the spack -d output when Ramble calls Spack and the build fails.

So, let me clarify my question:

Will setting this env var make Ramble call spack with -d?

scheibelp: There is currently not an environment variable that does this in Spack (I can see the appeal though).

pearce8: That makes sense, I think we have to set an env var in Ramble to tell Ramble what to do.  I think there are 2 commands that we can run on Ramble after running ramble/setup-env.sh:

1) ramble config --scope=site add “config:disable_progress_bar:true"

----- I think this should make it unnecessary to use "-P" when calling ramble setup and ramble on.

2) ramble config --scope=site add “config:spack:global:args:‘-d’”

------ I think this will make ramble call spack with -d.

I believe right now we can do this if we cd into {ramble_location}/share/ramble/.  I think it would be great for the benchpark setup.sh to make these calls.

@scheibelp Do you think the above is correct?  Can you add the above to this PR?

bin/benchpark
```diff
@@ -282,14 +284,30 @@ def benchpark_setup_handler(args):
         )
 
         run_command(f"{ramble_exe} repo add --scope=site {source_dir}/repo")
+        run_command(f"{ramble_exe} config --scope=site add \"config:disable_progress_bar:true\"")
+        run_command(f"{ramble_exe} config --scope=site add \"config:spack:global:args:'-d'\"")
```
scheibelp: >> Will setting this env var make Ramble call spack with -d?

> There is currently not an environment variable that does this in Spack (I can see the appeal though).

It was pointed out to me that while this isn't possible in Spack, Ramble has a means for this

bin/benchpark
scheibelp: See https://github.com/LLNL/benchpark/pull/109/files#r1454138774 - it looks like we can generally force running Spack in debug mode (when done via Ramble).

pearce8 (approved):

PR #93, opened 2024-01-05
add benchpark define command to create boilerplate system definition

slabasan:

PR #110, opened 2024-01-12
New Benchpark command: "system-create"

### LINKED ISSUE #92: System definition
Updated:

Below the dashed line, we have the *original* system config design.  Things have changed since then, and we need to:
- 

What should the user use as a starter for their system definition?
- [ ] describe the new system definition in the docs (@dyokelson)
- [ ] think about simplifying creation of a new system (@scheibelp)
- [ ] provide instructions on how to find the system with the most similar hardware (search table?  search benchpark?) (@dyokelson)
- [ ] provide instructions on how to define a new system 
- [ ] provide instructions on how to go about software stack definition (compilers, mpi, rocm version...) (@dyokelson)
- [ ] provide instructions (or tests - ie with a small suite of tests - and correct outcomes) on how to verify the system definition is correct.

---------------------------------------------------------
System configs currently contain different types of information, which serves different purposes:

1. Hardware specification
- where defined: system_definition.yaml
- systems it applies to: a class of systems at different sites
- longevity: duration of the system (or class of systems) lifetime
- purpose of record: find a system with the same hardware as my system.  May want to record with the experiment.
2. Software stack: compiler and MPI locations 
- where defined: Optional?!?  compilers.yaml
- systems it applies to: just ours?  can we autodetect?
- longevity: ?
- purpose of record: give the users a starting point to running on their system.  What errors and guidance for mitigation should we give?  Do we want these upstreamed back?  Do we want these recorded in the experiment?
3. Software stack: compiler and MPI versions
- where defined: compilers.yaml
- systems it applies to: different machines could be at different versions
- longevity: new versions can appear any time
- purpose: give the users a starting point, also need to record as part of experiment - and use to debug or compare performance.  Probably want to let users parameterize - and set up versions to use as part of their suites.
4. Scheduler, launcher:
- where defined: variables.yaml
- systems it applies to: many.  Probably need a slurm and a flux schedule definition, auto generated for the user when they tell us which it is (can we autodetect?).  Probably need to define a few launchers and pick one (mpirun, srun, ...) 
- longevity: static, except the queue info is baked in here unfortunately.
- purpose: give the users a starting point.  Probably don't want upstreamed, may not need to record.
5. Software packages we don't want to keep building
- where defined: Optional! packages.yaml
- systems it applies to: probably just ours.  May be able to find using spack external.
- longevity: yeah may want to update versions over time.
- purpose: shorten build time.  We do not want these upstreamed, but we want to be able to record for our own experiments/CI etc.

We should probably define a graded approach for generating these:
- only introduce a new hardware specification if one like it indeed does not exist.
- if hardware specification exists, pick scheduler&launcher, and how to start to define the compilers and MPI to use?
- versions should be parameterized
- optional things can be added later (if desired)

scheibelp:
Add command that can be used to instantiate a new system definition, for example:

```
$ benchpark system-create --from=HPECray-zen3-MI250X-Slingshot cray1
$ ls configs/cray1
auxiliary_software_files  spack.yaml  system_definition.yaml  variables.yaml

$ benchpark system-create cray2 --template=templates/scheduler/slurm/ --template=templates/generic-system/cray/
$ ls configs/cray2
spack.yaml  variables.yaml
```

This will often require an intermediate editing step, in particular to define `auxiliary_software_files`; after that, you could e.g. do:

```
$ benchpark setup amg2023/rocm cray2...
```

Broadly:

* You can copy an existing config with `--from`
* Or you can assembly a generic configuration using a series of `--template` options

If two templates define Ramble configurations with the same name, this command will use a `ramble-python` script to automatically merge them:

```
$ benchpark system-create cray2 --template=templates/scheduler/slurm/ --template=templates/experiment-deps/cray/ --template=templates/catch-all/
```

(where here, `templates/scheduler/slurm/` and `templates/catch-all` both define `variables.yaml`)

Current issues:

- [x] (done as of Jan. 23) Right now, if two templates define the same file name, the last one will overwrite all prior
  - (UPDATE Jan. 17) Work is in progress to handle merging these config files together (I plan to add that here before marking this PR as ready to merge)
- [ ] (New Jan. 23) Right now, I have `benchpark` clone its own Ramble when executing `system-create`, this means that there can be multiple Ramble installations:
  * One for `benchpark` itself (to run `ramble-python`)
  * One for managing each experiments directory (so there already could be multiple, but in general there would now be at least 2)

  It might be better to manage this `ramble` instance as a Git submodule
- [ ] (New Jan. 23) Tests for `system-create` would be good
- [ ] (possibly for later) `bin/benchpark` is now reaching a size where it would be useful to split it up
- [ ] I've added some initial entries to `templates/` to give a notion of how they are supposed to be used in combination with the `system-create` command, but there's not enough there to be actually useful yet


templates/scheduler/flux/variables.yaml
```diff
@@ -0,0 +1,14 @@
+# Copyright 2023 Lawrence Livermore National Security, LLC and other
+# Benchpark Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+variables:
+  gtl_flag: ''  # to be overwritten by tests that need GTL
+  rocm_arch: 'gfx90a'
```
scheibelp: This is a good example of something where we'd want to split `variables.yaml` into multiple files: this is probably the same across systems and is also not scheduler-specific.

slabasan:
@slabasan to migrate boilerplate `system_definition.yaml` from https://github.com/LLNL/benchpark/pull/93 to this PR

bin/benchpark
```diff
@@ -163,7 +215,21 @@ def benchpark_setup(subparsers, actions_dict):
         help="Where to install packages and store results for the experiments. Benchpark expects to manage this directory, and it should be empty/nonexistent the first time you run benchpark setup experiments.",
     )
 
+    system_create_parser = subparsers.add_parser(
+        "system-create", help="Initialize a system config"
```
alecbcs, line 219: It'd be great if we could split this out into `system create` instead and support additional sub-commands such as `system list` and `system edit` in the future.

bin/benchpark
```diff
@@ -163,7 +215,21 @@ def benchpark_setup(subparsers, actions_dict):
...
+    )
+    system_create_parser.add_argument(
+        "--template",
```
alecbcs, line 222: What do folks think about a shorthand `-t` as an additional flag?

bin/benchpark
```diff
@@ -163,7 +215,21 @@ def benchpark_setup(subparsers, actions_dict):
...
+        action="append",
+        help="Choose common config settings in templates/",
+    )
+    system_create_parser.add_argument(
+        "--from", dest="use_existing", type=str, help="Copy an existing system config"
```
alecbcs, line 227: Nice! We'd talked about this prior but I really like the idea of being able to copy from an existing system and just tweak it as necessary.

PR #114, opened 2024-01-16
Feature saxpy caliper support

rfhaque:


repo/saxpy/saxpy.cc
```diff
@@ -132,10 +132,7 @@ int main(int argc, char** argv) {
...
       adiak_executable();
       adiak_executablepath();
-      adiak_libraries();
-      adiak_cmdline();
-      adiak_hostname();
-      adiak_clustername();
+      adiak_collect_all();
       adiak_namevalue("compiler", adiak_general, NULL, "%s", SAXPY_COMPILER_ID);
       adiak_namevalue("compiler version", adiak_general, NULL, "%s", SAXPY_COMPILER_VERSION);
       CALI_MARK_BEGIN("main");
```
rfhaque, line 138: @daboehme The separation between "main" and "problem" is to allow for definition of other problems later on. However, if we have to remove one, can we remove "main" and keep "problem" or vice-versa?

daboehme (commented):
The `adiak_collect_all()` function replaces all of the implicit adiak calls, i.e. you can also delete the user, uid, launchdate, executable, and executablepath calls.

pearce8 (approved):

rfhaque (commented):
~@pearce8 One more change to consolidate adiak calls~
Opened a new PR #115

PR #116, opened 2024-01-19
Separating generic systems and site specific systems

pearce8:
Site specific systems are named site-name-system, and contain site specific details about package versions and locations.

Generic systems define the bare minimum for Benchpark to work, and are intended to be a starting point of a system specification for a specific hardware, that can be built upon to describe a site specific system.


configs/HPECray-zen3-MI250X-Slingshot/variables.yaml
```diff
@@ -1,27 +0,0 @@
...
-  slurm_mpi_command: 'srun -N {n_nodes} -n {n_ranks}'
-  slurm_batch_submit: 'sbatch {execute_experiment}'
-  slurm_batch_nodes: '#SBATCH -N {n_nodes}'
-  slurm_batch_ranks: '#SBATCH -n {n_ranks}'
-  slurm_batch_timeout: '#SBATCH -t {batch_time}:00'
-  mpi_command: '{{scheduler}_mpi_command}'
-  batch_submit: '{{scheduler}_batch_submit}'
-  batch_nodes: '{{scheduler}_batch_nodes}'
-  batch_ranks: '{{scheduler}_batch_ranks}'
-  batch_timeout: '{{scheduler}_batch_timeout}'
```
scheibelp, line 27: Is the intent to remove all of `HPECray-zen3-MI250X-Slingshot`? I get the impression it should, since now there is `generic-HPECray-zen3-MI250X-Slingshot`. If any of it stays, then I think this should stay (or a modified form of it that narrows things down to the scheduler.

(overall though it looks like the `git mv` was not complete for HPECray-zen3-MI250X-Slingshot, since several files in this diff are displayed as entirely new for generic-HPECray-zen3-MI250X-Slingshot)

scheibelp (commented):
The naming system seems reasonable although I'm not sure if lack-of-site means we should use the term "generic". I'd prefer "nosite"

~Also, it looks like the move for `HPECray-zen3-MI250X-Slingshot` is incomplete (more on that in specific comment).~ (edit: I was wrong about that, there is no longer a HPECray-zen3-MI250X-Slingshot directory)

configs/HPECray-zen3-MI250X-Slingshot/variables.yaml
pearce8: I moved the HPECray system to LLNL-Tioga, since I kept that identical (except switching the scheduler to be Flux).  Then I made a (partial) copy for generic-HPECray, excluding site specific packages (spack externals), and using Slurm as the scheduler.

Did I miss anything?

scheibelp: Sorry it's not clear IMO from the diff, but I thought there were remaining files in `HPECray-zen3-MI250X-Slingshot`. I tried pulling this PR and examining it locally, and the organization looks fine, so I'll mark this as resolved (I still think we should not use the term "generic" as part of the name though).

pearce8:
> The naming system seems reasonable although I'm not sure if lack-of-site means we should use the term "generic". I'd prefer "nosite"
> 
> Also, it looks like the move for `HPECray-zen3-MI250X-Slingshot` is incomplete (more on that in specific comment).

I can live with nosite.  I hope we will be able to do per-site directories in the long run; would the generic/nosite specs live in the base directory then, or in the generic/nosite directory?

scheibelp:
> I hope we will be able to do per-site directories in the long run; would the generic/nosite specs live in the base directory then, or in the generic/nosite directory?

The latter: in the nosite directory(which is I think how things are organized now).

If I understand correctly, by the "base" directory you mean something like `HPECray-zen3-MI250X-Slingshot`, which no longer exists in this PR - I think we should remove those (i.e. keep things as they are now in this PR).

pearce8:
> > I hope we will be able to do per-site directories in the long run; would the generic/nosite specs live in the base directory then, or in the generic/nosite directory?
> 
> The latter: in the nosite directory(which is I think how things are organized now).
> 
> If I understand correctly, by the "base" directory you mean something like `HPECray-zen3-MI250X-Slingshot`, which no longer exists in this PR - I think we should remove those (i.e. keep things as they are now in this PR).

Sounds good.  I renamed generic to nosite.  When we can do nested directories, we will plan to have SiteName directories, and a nosite directory for the nosite-* systems.

PR #118, opened 2024-01-19
Adding a Fugaku system config

pearce8:
@pearce8 Updated table, Fugaku config looks good. We might want to increase the content width of the page to reduce left/right scroll

Screenshot 2024-01-29 at 3 36 31 PM


pearce8 (commented):
Looks good, updates only system files and the docs table.

The change to the openmp number of threads might as well happen because 13 is a strange number.

scheibelp (approved):
I think this defines more compilers and packages than required, but I think that can be adjusted in a later PR.

PR #135, opened 2024-02-07
[Exploratory] Finding handles in 1) applications, 2) systems

pearce8:
- Greg's commit is an early prototype (with outstanding to do's) for getting handles from an application.  We need to develop this further.
- Is getting handles from a system similar?


pearce8:
Superseded by current work.

PR #141, opened 2024-02-20
Configs/lumi

pearce8:
To resolve (possibly in the next PR):
- partition (this is Lumi-C)
- allocation (I removed project allocation from this config, but how should users include it if they do not want to make it public?)


configs/CSC-Lumi-HPECray-zen3-MI250X-Slingshot/auxiliary_software_files/compilers.yaml
```diff
@@ -0,0 +1,35 @@
...
+- compiler:
+    spec: cce@16.0.0-rocm5.5.1
+    paths:
+      cc:  /opt/cray/pe/cce/16.0.0/bin/craycc
+      cxx:  /opt/cray/pe/cce/16.0.0/bin/crayCC
+      f77:  /opt/cray/pe/cce/16.0.0/bin/crayftn
+      fc:  /opt/cray/pe/cce/16.0.0/bin/crayftn
+    flags:
+      cflags: -g -O2
+      cxxflags: -g -O2 -std=c++14
```
scheibelp: I may have missed this in other configs: but I'd recommend against putting C++ standard flags in as a universal choice: each package should generally decide on its own whether it needs to enable a given c++ standard (and many Spack packages do this, which could conflict with this setting).

I also think `-g -O2` should be omitted here, but that should generally not be an error.

configs/CSC-Lumi-HPECray-zen3-MI250X-Slingshot/auxiliary_software_files/packages.yaml
```diff
@@ -0,0 +1,17 @@
...
+
+packages:
+  all:
+    require: 'target=x86_64:'
+    variants: amdgpu_target=gfx90a
+  tar:
+    externals:
+    - spec: tar@1.30
+      prefix: /usr
+
```
scheibelp: This should include a `cray-libsci` external entry, and should specify that `blas` and `lapack` are not buildable:

```
package:
  cray-libsci:
    externals:
    - spec: ...
      prefix: ...
  blas:
    buildable: false
  lapack:
    buildable: false
```

configs/CSC-Lumi-HPECray-zen3-MI250X-Slingshot/variables.yaml
```diff
@@ -0,0 +1,18 @@
...
+  rocm_arch: 'gfx90a'
+  batch_time: '02:00'
+  mpi_command: 'srun -N {n_nodes} -n {n_ranks}'
+  batch_submit: 'sbatch {execute_experiment}'
+  batch_nodes: '#SBATCH -N {n_nodes}'
+  batch_ranks: '#SBATCH -n {n_ranks}'
+  batch_timeout: '#SBATCH -t {batch_time}:00'
+  cpu_partition: '#SBATCH -p small'
+  gpu_partition: '#SBATCH -p small-g' 
+  gpu_config: '#SBATCH --gpus-per-task=1'
```
scheibelp: I'm not aware of any of our experiment templates making use of `gpu_config`. I'm not sure this is the way we'll want to allocate resources (I think whether we allocate 1 GPU per task is more relevant to a given experiment than to the whole system config).

configs/CSC-Lumi-HPECray-zen3-MI250X-Slingshot/spack.yaml
```diff
@@ -0,0 +1,26 @@
...
+# Benchpark Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+spack:
+  packages:
+    default-compiler:
+      spack_spec: cce@16
+    default-mpi:
+      spack_spec: cray-mpich@8.1%cce ~gtl
```
scheibelp: `gtl` isn't a variant on `cray-mpich`, and we don't have variants on compilers, so I think all this `+/-gtl` can be removed.

configs/CSC-Lumi-HPECray-zen3-MI250X-Slingshot/auxiliary_software_files/packages.yaml
scheibelp: A less critical issue is that it would be useful to add external entries for all the rocm components here: you can generate that quickly with `spack external find -t rocm`.

scheibelp (changes requested):
I have a few requests: I think the most important one is that this needs to specify an external `cray-libsci` installation in `packages.yaml`.

configs/CSC-Lumi-HPECray-zen3-MI250X-Slingshot/spack.yaml
scheibelp: I should point out that `LLNL-Tioga-HPECray-zen3-MI250X-Slingshot` also does this, and I think I've used that config to successfully generate experiments, so this might not be a breaking issue, but adding `~gtl` should not have any effect here, so keeping it would be misleading IMO.

pearce8: @scheibelp can you please make this change to the Tioga config, and run all the tests on Tioga with it for a sanity check?

If this indeed works on Tioga, please put up a PR to change this on Tioga AND the no-site version of the machine; I believe folks with similar hardware are starting with one of these two configs.

configs/CSC-Lumi-HPECray-zen3-MI250X-Slingshot/auxiliary_software_files/packages.yaml
scheibelp: This also needs an external entry for `cray-mpich`.

configs/CSC-Lumi-HPECray-zen3-MI250X-Slingshot/variables.yaml
pszi1ard: I agree, it can be removed.

pszi1ard:
Note that it is called "LUMI", so I suggest renaming the directory.

I suggest adding to the machine config files' header that this is based on the  configs from`/pfs/lustrep1/appl/lumi/spack/etc/23.09/0.21.0-user` and is tested with spack 0.21 (since benchpark checks out develop I guess it may break in the future).

configs/CSC-Lumi-HPECray-zen3-MI250X-Slingshot/spack.yaml
scheibelp: I just realized that the benchpark Spack package repository defines a `cray-mpich` with a `+gtl` variant, so I was wrong about this.

scheibelp:
> Note that it is called "LUMI", so I suggest renaming the directory.

I think you are referring to `configs/CSC-Lumi-HPECray-zen3-MI250X-Slingshot/`, our general format for these names is: `{organization-acronym}-{name-of-machine}-{generic-type-of-machine}-{target}-{fabric}`. Are you disagreeing with the `CSC` part?

pszi1ard:
> I think you are referring to configs/CSC-Lumi-HPECray-zen3-MI250X-Slingshot/, our general format for these names is: {organization-acronym}-{name-of-machine}-{generic-type-of-machine}-{target}-{fabric}. Are you disagreeing with the CSC part?

No, I was referring to the fact that the official name is all caps "LUMI" not "Lumi", see https://www.csc.fi/en/lumi

scheibelp (approved):

pszi1ard:
@pearce8 @scheibelp I just noticed that a few files are missing from this PR: config.yaml, modules.yaml, and upstream.yaml. Are these intentionally left out?

PR #151, opened 2024-02-27
Move generic constraints out of experiments

scheibelp:
Include a system-independent config that can be used to enforce universal sensible constraints

TODO

- [x] Right now Ramble won't include any file from `auxiliary_software_files` unless it matches the default name for a Spack config section (e.g. `packages.yaml`). So I either need to wait for that to relax, or make sure benchpark combines all sections into `packages.yaml`


scheibelp:
With `https://github.com/GoogleCloudPlatform/ramble/pull/426` and the latest commit to this branch, this works. 

Note: all config that you want to combine has to be put into the one `configs/common/spack_includes.yaml` file that is created here. Putting this file in `configs/common/` might suggest one can add other files, but that is not the case.

pearce8 (approved):

PR #155, opened 2024-03-01
Caliperizing stream

pearce8:


rfhaque:
@pearce8 @scheibelp Tested in openmp mode on quartz, x86, rzwhippet, tioga, lassen and pascal. Ready to merge

PR #156, opened 2024-03-01
Caliperizing hpl, need to modify package here

pearce8:
hpl/openmp works on:
- [X] Quartz
- [X] Lassen
- [X] Tioga


rfhaque:
@pearce8 Works in openmp mode on quartz, lassen and tioga. Can be taken to main

PR #161, opened 2024-03-06
clean up fugaku files and gromacs builds with gcc now with minor spac…

jdomke:
…k modifications to fujitsu-mpi package

### DIFF
--- a/configs/RCCS-Fugaku-Fujitsu-A64FX-TofuD/auxiliary_software_files/compilers.yaml
+++ b/configs/RCCS-Fugaku-Fujitsu-A64FX-TofuD/auxiliary_software_files/compilers.yaml
@@ -5,25 +5,6 @@
 
 compilers:
 - compiler:
-    spec: fj@4.10.0
-    modules: []
-    paths:
-      cc: /opt/FJSVxtclanga/tcsds-1.2.38/bin/fcc
-      cxx: /opt/FJSVxtclanga/tcsds-1.2.38/bin/FCC
-      f77: /opt/FJSVxtclanga/tcsds-1.2.38/bin/frt
-      fc: /opt/FJSVxtclanga/tcsds-1.2.38/bin/frt
-    flags: {}
-    operating_system: rhel8
-    target: aarch64
-    environment:
-      set:
-        fcc_ENV: -Nclang
-        FCC_ENV: -Nclang
-      prepend_path:
-        PATH: /opt/FJSVxtclanga/tcsds-1.2.38/bin
-        LD_LIBRARY_PATH: /opt/FJSVxtclanga/tcsds-1.2.38/lib64
-    extra_rpaths: []
-- compiler:
     spec: fj@4.8.1
     modules: []
     paths:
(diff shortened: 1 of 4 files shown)


configs/RCCS-Fugaku-Fujitsu-A64FX-TofuD/auxiliary_software_files/compilers.yaml
```diff
@@ -128,9 +53,19 @@ compilers:
       cxx: /vol0004/apps/oss/spack-v0.19/opt/spack/linux-rhel8-a64fx/gcc-8.5.0/gcc-12.2.0-sxcx7kmt3qiktffgzzvrj2wmup3g32bc/bin/g++
       f77: /vol0004/apps/oss/spack-v0.19/opt/spack/linux-rhel8-a64fx/gcc-8.5.0/gcc-12.2.0-sxcx7kmt3qiktffgzzvrj2wmup3g32bc/bin/gfortran
       fc: /vol0004/apps/oss/spack-v0.19/opt/spack/linux-rhel8-a64fx/gcc-8.5.0/gcc-12.2.0-sxcx7kmt3qiktffgzzvrj2wmup3g32bc/bin/gfortran
-    flags: {}
+    flags:
+      ldflags: {"-lelf -ldl"}
```
scheibelp, line 57: (potential issue) Des this appear to concretize for you? I'm not familiar with this YAML syntax (it looks like a set of one element), typically one would write:

```
ldflags: "-lelf -ldl"
```

Also, would you be able to mention the error you get without this? I don't want to hold this PR up on it, but I may be able to make build suggestions that avoids the need for this.

configs/RCCS-Fugaku-Fujitsu-A64FX-TofuD/auxiliary_software_files/packages.yaml
```diff
@@ -5,458 +5,206 @@
...
   all:
-    compiler: [fj, gcc]
+    compiler: [fj, clang, gcc]
     providers:
       mpi: [fujitsu-mpi, openmpi, mpich]
       blas: [fujitsu-ssl2, openblas]
       lapack: [fujitsu-ssl2, openblas]
       scalapack: [fujitsu-ssl2, netlib-scalapack]
-      fftw-api: [fujitsu-fftw, fftw, rist-fftw]
+      fftw-api: [fujitsu-ssl2, fftw, rist-fftw]
```
scheibelp, line 14: This should not have an effect: `fujitsu-ssl2` does not provide `fftw-api`. Spack can work around this because the expressions here are taken as preferences by spack (i.e. not hard requirements). In other words, this shouldn't cause anything to fail, but I also think it isn't necessary for anything to succeed.

configs/RCCS-Fugaku-Fujitsu-A64FX-TofuD/spack.yaml
```diff
@@ -6,20 +6,31 @@
 spack:
   packages:
     default-compiler:
-      spack_spec: fj@{default_fj_version}
-    default-mpi:
-      spack_spec: fujitsu-mpi@{default_fj_version}%fj arch=linux-rhel8-a64fx
-    compiler-gcc:
+      #spack_spec: fj@{default_fj_version}
+      #spack_spec: clang@{default_llvm_version}
```
scheibelp, line 10: Can you remove these?

configs/RCCS-Fugaku-Fujitsu-A64FX-TofuD/spack.yaml
```diff
@@ -6,20 +6,31 @@
...
       spack_spec: gcc@{default_gnu_version}
+    default-mpi:
+      spack_spec: fujitsu-mpi@{default_fj_version} arch=linux-rhel8-a64fx
     compiler-fujitsu:
       spack_spec: fj@{default_fj_version}
+    compiler-clang:
+      spack_spec: clang@{default_llvm_version}
```
scheibelp, line 17: (no change needed) This may be required until we invoke Ramble in a way that avoids having it look for additional compilers.

configs/RCCS-Fugaku-Fujitsu-A64FX-TofuD/auxiliary_software_files/compilers.yaml
```diff
@@ -128,9 +53,19 @@ compilers:
...
+    environment:
+      set:
+        OMPI_CC: /vol0004/apps/oss/spack-v0.19/opt/spack/linux-rhel8-a64fx/gcc-8.5.0/gcc-12.2.0-sxcx7kmt3qiktffgzzvrj2wmup3g32bc/bin/gcc
+        OMPI_CXX: /vol0004/apps/oss/spack-v0.19/opt/spack/linux-rhel8-a64fx/gcc-8.5.0/gcc-12.2.0-sxcx7kmt3qiktffgzzvrj2wmup3g32bc/bin/g++
+        OMPI_FC: /vol0004/apps/oss/spack-v0.19/opt/spack/linux-rhel8-a64fx/gcc-8.5.0/gcc-12.2.0-sxcx7kmt3qiktffgzzvrj2wmup3g32bc/bin/gfortran
+        OMPI_F77: /vol0004/apps/oss/spack-v0.19/opt/spack/linux-rhel8-a64fx/gcc-8.5.0/gcc-12.2.0-sxcx7kmt3qiktffgzzvrj2wmup3g32bc/bin/gfortran
```
scheibelp, line 63: (potential issue) When we set `OMPI_*` variables, it is typically to point them at the underlying spack wrappers, it might be preferable to handle this as an update to `fujitsu-mpi`'s `package.py` file in Spack.

scheibelp (commented):
I wanted to check: you can build benchmarks on Fugaku using these changes? A there are a couple spots that look like they might cause problems, but fixing those could be reserved for a later PR if this works for you.

configs/RCCS-Fugaku-Fujitsu-A64FX-TofuD/auxiliary_software_files/compilers.yaml
jdomke: linking against fujitsu's blas lib (SSL2) with gcc results in missing symbols, which is why i need those flags

configs/RCCS-Fugaku-Fujitsu-A64FX-TofuD/auxiliary_software_files/packages.yaml
jdomke: afaik, fujitsu-ssl2 provides fft functions, but for some reason the fftw.h header is missing. i'm also not yet 100% sure how to set it properly, when i compile (forgot) gromacs or hyper i see that it doesnt find the header and then builds fftw from source (which isn't the best approach, IMHO). i don't think 'fujitsu-fftw' exists, so the previous config is definitely wrong.

configs/RCCS-Fugaku-Fujitsu-A64FX-TofuD/spack.yaml
jdomke: in theory yes, but i would like to keep them for now, because i regularly switch compilers to test things and right onw the "default" is gcc because i was able to build amg and gromacs with it. are those comments causing issues or why would you want them to go?

configs/RCCS-Fugaku-Fujitsu-A64FX-TofuD/auxiliary_software_files/compilers.yaml
jdomke: i understand. i can take a look how to fix fujitsu-mpi with an additional PR. for now its just to work around problems i run into with fugaku.

jdomke:
> I wanted to check: you can build benchmarks on Fugaku using these changes? A there are a couple spots that look like they might cause problems, but fixing those could be reserved for a later PR if this works for you.

yes, i was able to build amg2023 and gromacs with gc with those changes. i'll check clang next, and verify that i didnt break anything for fujitsu compiler.
but i had to move to gcc because gromacs was refusing to build with fujitsu (the c++17 support is not 100% working and result in linker errors with some ABI 11 functions, don't ask me why, but for now gromacs+fujitsu is a no-go)

configs/RCCS-Fugaku-Fujitsu-A64FX-TofuD/spack.yaml
scheibelp: I thought this was leftover from debugging, but given that you explain this is useful in the future, it seems fine to keep for now. We should work on providing another mechanism for you to make this switch without having to manually edit.

configs/RCCS-Fugaku-Fujitsu-A64FX-TofuD/auxiliary_software_files/packages.yaml
scheibelp: What I mean is that because Spack's builtin `fujitsu-ssl2` package.py does not have `provides(fftw-api)` spack will never assume it is useful as an fftw provider and will always try to use something else (the logs generated by `ramble workspace setup` should include the concretization, and I assume Spack will have chosen something besides `fujitsu-ssl2` wherever an `fftw` implementation is needed).

scheibelp (approved):
You mentioned this working for you so I think the best approach is to merge this as is and address any concerns in a follow-up.

Thanks!

configs/RCCS-Fugaku-Fujitsu-A64FX-TofuD/auxiliary_software_files/packages.yaml
jdomke: I see, thanks for the clarification

PR #172, opened 2024-03-14
Fix docs output path

alecbcs:
Follow up to #171. The docs Makefile we're now using outputs another layer of directories before getting to the HTML.


pearce8:
@alecbcs @slabasan did this (and previous) PR move the docs/tables/current-system-definitions.csv so that it is not part of the git diff that has been causing conflicts for system PRs until now?

alecbcs:
Yes we pulled out the .csv files since they're dynamically created in CI they shouldn't be tracked in the repo

slabasan (approved):

PR #178, opened 2024-03-15
[WIP] Generate batch submission script at experiment run-time

scheibelp:
On `develop`, Benchpark generates batch scripts directly using Ramble's templating system.

This PR alters `execute_experiment` to generate the sbatch submission: `execute_experiment` now calls out to a Python script which generates the batch script (e.g. for sbatch); this Python script takes as input:

* The resource requests for the experiment (e.g. from `experiments/saxpy/openmp/ramble.yaml`
* The system description (e.g. from `configs/.../variables.yaml`)

and can combine them to generate an appropriate request. For example given:

`./bin/benchpark setup saxpy/openmp nosite-x86_64 `pwd`/test-saxpy-oslic`

`<root>/test-saxpy-oslic/saxpy/openmp/nosite-x86_64/workspace/experiments/saxpy/problem/saxpy_512_1_8_2/execute_experiment` will contain

```
...
result=`<root>/test-saxpy-oslic/saxpy/openmp/nosite-x86_64/workspace/experiments/saxpy/problem/../../../generate-batch`
read invoke_with output_script <<< "$result"

echo "step 3"

chmod +x "$output_script"
"$invoke_with" "$output_script"
```

and `$output_script` contains:

```
#SBATCH -N 1
#Expanded variables: {<AllocOpt.N_TASKS: 1>: '8', <AllocOpt.N_NODES: 2>: '1', <AllocOpt.N_OMP_THREADS_PER_TASK: 5>: '2', <AllocOpt.N_TASKS_PER_NODE: 6>: '8'}
rm -f "<root>/test-saxpy-oslic/saxpy/openmp/nosite-x86_64/workspace/experiments/saxpy/problem/saxpy_512_1_8_2/saxpy_512_1_8_2.out"
touch "<root>/test-saxpy-oslic/saxpy/openmp/nosite-x86_64/workspace/experiments/saxpy/problem/saxpy_512_1_8_2/saxpy_512_1_8_2.out"
. <root>/test-saxpy-oslic/spack/share/spack/setup-env.sh
spack env activate <root>/test-saxpy-oslic/saxpy/openmp/nosite-x86_64/workspace/software/saxpy.problem
export OMP_NUM_THREADS=2
#mpirun -n 8 saxpy -n 512 >> "<root>/test-saxpy-oslic/saxpy/openmp/nosite-x86_64/workspace/experiments/saxpy/problem/saxpy_512_1_8_2/saxpy_512_1_8_2.out
```

### DIFF
--- a/configs/nosite-x86_64/variables.yaml
+++ b/configs/nosite-x86_64/variables.yaml
@@ -6,7 +6,8 @@
 variables:
   batch_time: ''
   mpi_command: 'mpirun'
-  batch_submit: 'sbatch'
+  scheduler: 'slurm'
+  batch_submit: '{execute_experiment}'
   batch_nodes: ''
   batch_ranks: ''
   batch_timeout: ''
--- a/configs/nosite-x86_64/variables.yaml
+++ b/configs/nosite-x86_64/variables.yaml
@@ -5,7 +5,7 @@
 
 variables:
   batch_time: ''
-  mpi_command: 'mpirun -n {n_nodes} -c {n_ranks} --oversubscribe'
+  mpi_command: 'mpirun'
   batch_submit: '{execute_experiment}'
   batch_nodes: ''
   batch_ranks: ''
--- a/configs/nosite-x86_64/variables.yaml
+++ b/configs/nosite-x86_64/variables.yaml
@@ -6,7 +6,7 @@
 variables:
   batch_time: ''
   mpi_command: 'mpirun'
-  batch_submit: '{execute_experiment}'
+  batch_submit: 'sbatch'
   batch_nodes: ''
   batch_ranks: ''
   batch_timeout: ''
--- a/experiments/saxpy/openmp/execute_experiment.tpl
+++ b/experiments/saxpy/openmp/execute_experiment.tpl
@@ -6,15 +6,18 @@
 
 cd {experiment_run_dir}
 
-export N_TASKS={n_ranks}
-export N_NODES={n_nodes}
-export GPUS_PER_NODE={gpus_per_node}
-export EXPERIMENT_RUN_DIR={experiment_run_dir}
+export N_TASKS="{n_ranks}"
+export N_NODES="{n_nodes}"
+export GPUS_PER_NODE="{gpus_per_node}"
+export EXPERIMENT_RUN_DIR="{experiment_run_dir}"
+export BATCH_SUBMIT="{batch_submit}"
 
-output_script=`{workload_run_dir}/../../../generate-batch`
+export SETUP_AND_RUN={experiment_run_dir}/setup_and_run.txt
 
-cat <<EOF >> output_script
+cat <<EOF >> "$SETUP_AND_RUN"
 {command}
 EOF
 
-output_script
+{workload_run_dir}/../../../generate-batch
+chmod +x output_script
+#output_script
--- a/experiments/saxpy/openmp/execute_experiment.tpl
+++ b/experiments/saxpy/openmp/execute_experiment.tpl
@@ -10,6 +10,7 @@ export N_TASKS="{n_ranks}"
 export N_NODES="{n_nodes}"
 export GPUS_PER_NODE="{gpus_per_node}"
 export CORES_PER_TASK="{cores_per_task}"
+export N_OMP_THREADS_PER_TASK="{n_omp_threads_per_task}"
 
 export SCHEDULER="{scheduler}"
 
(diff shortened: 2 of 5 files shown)


experiments/saxpy/openmp/execute_experiment.tpl
```diff
@@ -4,10 +4,30 @@
 #
```
scheibelp, line 4: Note: with the changes in this PR, there would be no more need for a per-experiment `execute_experiment.tpl`. To reach that, most `experiment/.../ramble.yaml` files and `configs/.../variables.yaml` files need to be tweaked, so these changes are localized to make the behavior easy to experiment with until I do that.

experiments/saxpy/openmp/execute_experiment.tpl
```diff
@@ -4,10 +4,30 @@
...
-{batch_nodes}
-{batch_ranks}
-{batch_timeout}
-
 cd {experiment_run_dir}
 
+export N_TASKS="{n_ranks}"
+export N_NODES="{n_nodes}"
+export GPUS_PER_NODE="{gpus_per_node}"
+export CORES_PER_TASK="{cores_per_task}"
```
scheibelp: Note: if these are undefined (e.g. not appearing in `variables.yaml` for a given experiment) then the template filling generates the literal `export CORES_PER_TASK="{cores_per_task}"`

`generate-batch` can then infer that the particular experiment does not set that variable.

experiments/saxpy/openmp/ramble.yaml
```diff
@@ -19,22 +19,19 @@ ramble:
...
-            set:
-              OMP_NUM_THREADS: '{n_threads}'
           variables:
             n_ranks: '8'
           experiments:
-            saxpy_{n}_{n_nodes}_{n_ranks}_{n_threads}:
+            saxpy_{n}_{n_nodes}_{n_ranks}_{n_omp_threads_per_task}:
               variables:
                 processes_per_node: ['8', '4']
                 n_nodes: ['1', '2']
```
scheibelp, line 28: Note: out of `n_nodes`, `processes_per_node`, and `n_ranks`, only 2 of these 3 need to be defined to fully specify what is needed.

pearce8: That makes sense.  Ideas on how we can check that 2 of 3 are defined?  If 3 are defined, we need to check that the math works out.

It looks like n_omp_threads_per_task additionally needs to be defined for OpenMP.  I recommend we replace "task" with "rank" everywhere (they do mean the same thing in this context).

configs/nosite-x86_64/variables.yaml
```diff
@@ -5,7 +5,8 @@
 
 variables:
   batch_time: ''
-  mpi_command: 'mpirun -n {n_nodes} -c {n_ranks} --oversubscribe'
+  mpi_command: 'mpirun'
```
scheibelp, line 8: Notes:

* With this PR (and if it is accepted, going forward), it should be left to `generate-batch` to insert options like `-n`
* For `mpirun` specifically, `-n` and `-c` are the same option (it differs from `srun` in that sense)

scheibelp:
My reasoning behind doing this vs. encoding arithmetic into ramble.yaml configs: I think it has an advantage in the following cases:

* Users change how they ask for resources: I think some users would want to say N_TASKS+N_OMP_THREADS_PER_TASK and some will want to say N_OMP_THREADS+N_OMP_THREADS_PER_TASK - a script could more easily handle blank/null values
* Balancing multiple resources: if we want to specify per-task GPU, CPU, *and* memory limits, the arithmetic becomes quite cumbersome
* A script could avoid running certain requests that might be deemed inappropriate for a machine
    * e.g. a single task that requests 200 cores: we could intercept and not run this
* A script could more easily account for optional parameters:
    * An additional "overloading" factor (doing a job on a fewer number of nodes than should be used)
    * Whether to account for hyperthreading (treating hardware threads as distinct CPUs)
* The arithmetic is trickier than might be anticipated even for fairly straightforward cases
    * e.g. 8 cores per task on a system where nodes have 36 cores, for X tasks, the answer is not the ceiling of (8*X)/CORES_PER_NODE

Use cases (some are given above too, but this section focuses specifically on scenarios we want to enable, vs. comparison):

* Openmp: we probably want to specify a number of omp threads per rank (we don't want just one omp thread for each rank)
    * We might want to specify the number of omp threads and number of ranks, and then have something automatically divide for us
* Cuda: we want to specify GPUs, and make sure that we allocate enough nodes to get the number of GPUs we want
* Mixed GPU/omp cases (do we want to support this?): each rank might want X GPUs and Y omp threads
    * What about cases where one or both don't align with per-node resources?

scheibelp:
@douglasjacobsen pointed out 

* this may be better handled with modifiers (and I think @becker33 understood this suggestion and liked it as an alternative as well).
* with the approach in this PR, Ramble queries like "give me all the jobs that run on 4 nodes" will not work (because with this approach, the number of nodes is not determined until `ramble on` time)

I added some of the cases that made this PR appealing to me in https://github.com/LLNL/benchpark/pull/178#issuecomment-2008221116

PR #190, opened 2024-03-26
Add modifier docs

### LINKED ISSUE #174: Add documentation on caliper modifier
Should be a sub-bullet under "Edit experiments".
Should reference Caliper - possibly the front page of Caliper docs.
Should reference Ramble modifiers page.
Should provide instructions on how to turn on the caliper modifier.

pearce8:
- [ ] (later) modifiers for setting up the resources etc
- [x] Modifiers section - generic definition of a modifier, link to Ramble modifiers
- [x] Caliper modifiers section - generic “Caliper modifier” definition, how to use them, link to Caliper.   Followed with a list of specific Caliper modifiers available, where they are applicable (the initial one is platform independent, then its x86 for top-down, and cuda GPUs for the cuda one), and what metrics they collect (so maybe a table with 3 columns we can add to?)

Screenshot 2024-04-04 at 10 48 45 PM


pearce8:
@slabasan Thanks for the table!  I think we can put the modifier name in the left column, so this would be:

caliper
caliper-topdown
caliper-cuda

and would match the string one would put after --modifier="one of the strings in the list above".  Do you think this you can make this clear?

bin/benchpark
```diff
@@ -147,21 +157,31 @@ def benchpark_list_handler(args):
...
+                        print(f"\t{system}")
+                else:
+                    if sublist == "modifiers":
+                        print("Modifiers:")
+                        for modifier in modifiers:
+                            print(f"\t{modifier}")
+                    else:
+                        raise ValueError(
+                            f'Invalid benchpark list "{sublist}" - must choose [experiments], [systems], [modifiers] or leave empty'
+                        )
```
scheibelp: (suggestion) I recommend "flattening" this:

```
if sublist == "systems":
    ... # same as before
elif sublist == "modifiers":
    ... # dedent, but otherwise same as before
else:
    raise ValueError(...)
```

bin/benchpark
```diff
@@ -311,32 +344,34 @@ def benchpark_setup_handler(args):
...
-    valid_benchmark = benchpark_check_benchmark(benchmark)
+    debug_print(f"specified experiment (benchmark/ProgrammingModel) = {experiment}")
+    valid_experiment = benchpark_check_experiment(experiment)
     debug_print(f"specified system = {system}")
     valid_system = benchpark_check_system(system)
-    if not (valid_benchmark and valid_system):
+    debug_print(f"specified modifier = {modifier}")
+    valid_modifier = benchpark_check_modifier(modifier)
+    if not (valid_experiment and valid_system and valid_modifier):
         raise ValueError(
```
scheibelp: (minor) Note the individual `benchpark_check*` functions already throw an exception for each variable being checked, so I think this extra `raise` would be redundant. That was true before this PR, and these changes make the logic consistent with the new modifier check.

scheibelp (approved):
I have a couple suggestions but don't see any major issues.

PR #193, opened 2024-04-03
caliper top-down modifier

rfhaque:


pearce8:
@rfhaque Can you please test on all x86 machines on-site we have defined in benchpark?

rfhaque:
@pearce8 Tested on quartz and pascal

slabasan:
@rfhaque Can you include here how you turn on this caliper modifier? I just have `--modifier=caliper`, but will we need more specifics to select between top-down, cuda, etc?

pearce8:
We need to impose a constraint on the spack spec as this should only be used on Intel's x86.  If a user tries to use this modifier on a different architecture, we should make sure they get an error indicating that the modifier does not work on this architecture.  Doug provided this example: https://github.com/GoogleCloudPlatform/ramble/blob/develop/lib/ramble/ramble/language/modifier_language.py#L285

rfhaque:
in `benchpark setup`, we specify `--modifier=caliper-topdown`

pearce8 (approved):

PR #194, opened 2024-04-03
caliper cuda modifier

rfhaque:


pearce8:
@rfhaque Can you please test on all cuda machines on-site we have defined in benchpark?

rfhaque:
@pearce8 Its already tested on pascal and lassen.

pearce8:
We need to impose a constraint on the spack spec as this should only be used on NVIDIA GPUs (where CUDA is available).  If a user tries to use this modifier on a different architecture, we should make sure they get an error indicating that the modifier does not work on this architecture.  Doug provided this example: https://github.com/GoogleCloudPlatform/ramble/blob/develop/lib/ramble/ramble/language/modifier_language.py#L285

pearce8:
@rfhaque Also, how would one turn on this modifier?  @slabasan needs to know for the docs.

rfhaque:
in `benchpark setup`, we specify `--modifier=caliper-cuda`

rfhaque:
> We need to impose a constraint on the spack spec as this should only be used on NVIDIA GPUs (where CUDA is available). If a user tries to use this modifier on a different architecture, we should make sure they get an error indicating that the modifier does not work on this architecture. Doug provided this example: https://github.com/GoogleCloudPlatform/ramble/blob/develop/lib/ramble/ramble/language/modifier_language.py#L285

I'll take a look at this

pearce8 (approved):

PR #195, opened 2024-04-03
Use Ramble modifier to fill in allocation variables

### LINKED ISSUE #221: Set up dry-run tests for experiments in benchpark
This is how this works in Ramble:
https://googlecloudplatform.github.io/ramble/tutorials/2_running_a_simple_gromacs_experiment.html#setting-up-the-experiments
  ramble repo add <benchpark_app_repo>
  ramble unit-test -k known_applications

Near-term: perform the ramble unit-test in a benchpark workspace, with a list of applications (perhaps the x86 versions to start)
Mid-term: create a list of applications (perhaps the x86 versions to start) and set them up to run in CI 
Longer-term: dry-run test when a new experiment is contributed via a PR

scheibelp:
Closes https://github.com/LLNL/benchpark/pull/178
Fixes #221

Given an experiment that requests resources (nodes, cpus, gpus, etc.) and a system description (cpus-per-node, gpus-per-node, etc.) this intends to generate an appropriate scheduler request for resources. In some cases that ends up determining things like how many nodes are desired for a given benchmark.

https://github.com/LLNL/benchpark/pull/178#issuecomment-2008221116 brings up some more-interesting examples like these, and this PR is an alternative approach.

This requires a newer Ramble than what Benchpark currently uses by default (right now I'm using `https://github.com/GoogleCloudPlatform/ramble/pull/452`).

```
./bin/benchpark setup saxpy/openmp nosite-x86_64 `pwd`/test-saxpy
```

### Remaining work:

- [x] remove all experiment-specific `execute_experiment.tpl` files
- [x] Implement scheduler definition function for Sierra and Fugaku
- [x] Update all experiment files and all system config files (currently just `experiments/saxpy/openmp` and `configs/nosite-x86_64` are changed to demonstrate the organization)
- [ ] All `experiment/*/*/ramble.yaml` files have been translated, but need further updates to actually describe system resources (e.g. number of GPUs on each node etc.)
  - (April 22 2024) All LLNL systems are now updated with # of CPUs/GPUs per-node (in the latter case, only for systems that have them)
  - (April 23 2024) All Systems except Eiger are now updated (note that LUMI and Daint have partitions with different types of nodes, and currently the variables only describe one type)
- [x] (May 14 2024) Update CI to do a `ramble workspace setup --dry-run` of some configs and experiments: this actually runs the modifier defined here to generate batch scripts etc. with all resource requests filled in

### Testing:

You can run any one of the following on any system

```
./bin/benchpark setup saxpy/openmp nosite-x86_64 <basedir>
./bin/benchpark setup amg2023/cuda LLNL-Sierra-IBM-power9-V100-Infiniband <basedir>
./bin/benchpark setup amg2023/cuda LLNL-Pascal-Penguin-broadwell-P100-OmniPath <basedir>
```

For the `ramble workspace setup` command it tells you to run, just append `--phases make_experiments` to the end of it (that will skip the concretize/install steps).

### Oddities:

* (April 23 2024) LUMI/Daint nodes have different characteristics based on what partition you request. For now, I only describe one type of node. I think we can handle this in the future by creating different configs based on what partition the user wants to submit to.
* (April 12 2024) The GROMACS `execute_experiment.tpl` files are slightly different than the others: they have an extra `{experiment_setup}`; everything sets that variable to `''` though, so I don't see a problem with removing them as well
* (EDIT: now resolved) ~Some values must be defined before the modifier runs, e.g. n_ranks. I've arbitrarily decided the placeholder value for these is "7" (they must be positive integers, so I decided to choose a number that was (a) unlikely to be explicitly chosen and (b) small (in case they percolate to actual requests)~


modifiers/allocation/modifier.py
```diff
@@ -0,0 +1,103 @@
...
+SENTINEL_UNDEFINED_VALUE = 7
+
+
+def defined_allocation_options(expander):
+    """For each possible allocation option, check if it was filled in by
+       Ramble: in that case it will be an integer.
+    """
+    defined = {}
+    for alloc_opt in AllocOpt:
+        var_def = expander.expand_var(f"{{{alloc_opt.name.lower()}}}")
```
scheibelp: This appears to work for some variables, like `n_threads`, but not ones like `n_cores_per_task`/`n_ranks_per_node` (the modified `experiments/saxpy/openmp/ramble.yaml` in this PR):

```
  applications:
    saxpy:
      workloads:
        problem:
          env_vars:
            set:
              OMP_NUM_THREADS: '{n_threads}'
          variables:
            n_cores_per_task: '1'
          experiments:
            saxpy_{n}_{n_nodes}_{n_ranks}_{n_threads}:
              variables:
                n_ranks_per_node: ['8', '4']
                n_nodes: ['1', '2']
                n_threads: ['2', '4']
                n: ['512', '1024']
  applications:
    saxpy:
      workloads:
        problem:
          env_vars:
            set:
              OMP_NUM_THREADS: '{n_threads}'
          variables:
            n_cores_per_task: '1'
          experiments:
            saxpy_{n}_{n_nodes}_{n_ranks}_{n_threads}:
              variables:
                n_ranks_per_node: ['8', '4']
                n_nodes: ['1', '2']
                n_threads: ['2', '4']
                n: ['512', '1024']

```

configs/nosite-x86_64/variables.yaml
```diff
@@ -4,9 +4,10 @@
...
-  batch_nodes: ''
-  batch_ranks: ''
-  batch_timeout: ''
+  mpi_command: 'mpirun -n {n_ranks} --oversubscribe'
+  n_ranks: '7'
+  n_nodes: '7'
+  n_threads: '7'
+  scheduler: "slurm"
+  batch_submit: "placeholder"
+  mpi_command: "placeholder"
```
scheibelp, line 14: n_ranks, n_nodes, ~n_threads~, batch_submit, and mpi_command must all be defined (for the first three, they must be positive integers), so I'm thinking each system can define them, and individual experiments can skip defining any that they want the modifier to handle.

experiments/saxpy/openmp/execute_experiment.tpl
```diff
@@ -4,9 +4,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-{batch_nodes}
-{batch_ranks}
-{batch_timeout}
+{allocation_directives}
```
scheibelp: This PR can now generate appropriate headers and invocations, so the only thing a system needs to specify is a `scheduler` variable, at which point the modifier will set:

* {allocation_directives}
* {mpi_command}
* {batch_submit}

Moreover, with this PR there only needs to be one `execute_experiment.tpl` for all experiments

scheibelp:
Example script generated from `experiments/saxpy/openmp` and `configs/nosite-x86_64` (tweaked to assume slurm, for a more interesting output):

```
#SBATCH -n 8
#SBATCH -N 1
#SBATCH --time 120

cd <benchpark-prefix>/test-saxpy-oslic-new/saxpy/openmp/nosite-x86_64/workspace/experiments/saxpy/problem/saxpy_512_1_2

rm -f "<benchpark-prefix>/test-saxpy-oslic-new/saxpy/openmp/nosite-x86_64/workspace/experiments/saxpy/problem/saxpy_512_1_2/saxpy_512_1_2.out"
touch "<benchpark-prefix>/test-saxpy-oslic-new/saxpy/openmp/nosite-x86_64/workspace/experiments/saxpy/problem/saxpy_512_1_2/saxpy_512_1_2.out"
export OMP_NUM_THREADS="2";
. <benchpark-prefix>/test-saxpy-oslic-new/spack/share/spack/setup-env.sh
spack env activate <benchpark-prefix>/test-saxpy-oslic-new/saxpy/openmp/nosite-x86_64/workspace/software/saxpy.problem
srun -n 8 -N 1 saxpy -n 512 >> "<benchpark-prefix>/test-saxpy-oslic-new/saxpy/openmp/nosite-x86_64/workspace/experiments/saxpy/problem/saxpy_512_1_2/saxpy_512_1_2.out"
```

configs/nosite-x86_64/variables.yaml
```diff
@@ -4,9 +4,10 @@
 # SPDX-License-Identifier: Apache-2.0
 
 variables:
-  batch_time: ''
-  mpi_command: 'mpirun -n {n_nodes} -c {n_ranks} --oversubscribe'
-  batch_submit: '{execute_experiment}'
-  batch_nodes: ''
-  batch_ranks: ''
-  batch_timeout: ''
+  mpi_command: 'mpirun -n {n_ranks} --oversubscribe'
```
pearce8: Launcher command (mpirun, flux run, srun) always takes n_nodes and n_ranks.

configs/nosite-x86_64/variables.yaml
```diff
@@ -4,9 +4,10 @@
...
+  n_ranks: '7'
+  n_nodes: '7'
```
pearce8: lets leave default n_nodes as 1 (single node).

configs/nosite-x86_64/variables.yaml
```diff
@@ -4,9 +4,10 @@
...
+  n_threads: '7'
```
pearce8: On CPU, n_threads is always n_threads_per_node - should probably make this clear here.

configs/nosite-x86_64/variables.yaml
pearce8: I think we need n_cores_per_node.  Then n_threads_per_node can default to n_cores_per_node, or people can indicate n_threads_per_node = 2 * n_cores_per_node if they have hyper threads.

n_ranks_per_node can default to n_cores_per_node (or n_gpus_per_node on accelerated systems).

n_ranks = n_nodes * n_ranks_per_node.

modifiers/allocation/modifier.py
```diff
@@ -0,0 +1,204 @@
...
+import math
+from enum import Enum
+from ramble.modkit import *
+
+
+class AllocOpt(Enum):
+    # Experiment resource requests
+    N_RANKS = 1
+    N_NODES = 2
+    N_CORES_PER_TASK = 3
```
pearce8: I think N_CORES_PER_TASK should be N_CORES_PER_NODE

modifiers/allocation/modifier.py
```diff
@@ -0,0 +1,204 @@
...
+    N_THREADS = 4  # number of OMP threads per task
+    N_RANKS_PER_NODE = 5
+    N_GPUS = 6
+
+    # Descriptions of resources available on systems
+    GPUS_PER_NODE = 100
```
pearce8: we might want to default this to 2

modifiers/allocation/modifier.py
```diff
@@ -0,0 +1,204 @@
...
+    CPUS_PER_NODE = 101
```
pearce8: we might want to default this to 8

modifiers/allocation/modifier.py
```diff
@@ -0,0 +1,204 @@
...
+
+    # Scheduler identification and other high-level instructions
+    SCHEDULER = 200
```
pearce8, line 28: I think you are using some sort of numbering system here.  How does this work?

modifiers/allocation/modifier.py
```diff
@@ -0,0 +1,204 @@
...
+    TIMEOUT = 201  # This is assumed to be in minutes
```
pearce8, line 29: the scheduler needs a few more pieces of info:
- batch_time
- batch queue (we have pdebug and pbatch usually, but other centers may name these different.  Often, special batches are created to give priority to a project etc.)
- account/bank (a bank/allocation we want to 'charge' for what we are working on - which is a time allocation for our project.  This is common across centers)

modifiers/allocation/modifier.py
```diff
@@ -0,0 +1,204 @@
...
+    def determine_allocation(self, v):
+        if not v.n_ranks:
+            if v.n_ranks_per_node and v.n_nodes:
+                v.n_ranks = v.n_nodes * v.n_ranks_per_node
+
+        if not v.n_nodes:
+            if v.n_ranks:
+                cpus_request_per_task = 1
+                multi_cpus_per_task = v.n_cores_per_task or v.n_threads or 0
+                cpus_request_per_task = max(multi_cpus_per_task, 1)
```
pearce8: what is "task" here?

configs/nosite-x86_64/variables.yaml
scheibelp: Sorry I should have set this to "placeholder", because it is now set by the modifier. But regarding the ability to control nodes/processes in mpirun:

I don't think `mpirun` has an option to request nodes (the way you can provide `-N` to srun) - see: https://www.open-mpi.org/doc/current/man1/mpirun.1.php 

* For example `-N` in mpirun specifies the number of processes to launch per node
* Originally this was 'mpirun -n {n_nodes} -c {n_ranks} --oversubscribe', but from those docs `-c` and `-n` are synonyms

The only mechanism I see for controlling which/how-many nodes is by creating a host file

configs/nosite-x86_64/variables.yaml
scheibelp: Wherever I put "7" in this file, it means "ignore this value", any other value is taken to mean "this value is set, do not change it". I assume "1" is a value we would want.

configs/nosite-x86_64/variables.yaml
scheibelp: I thought Ramble was treating this specially but I think I'm wrong. I meant it to represent OMP_NUM_THREADS and I agree a more-descriptive name would be better (I'm thinking num_omp_threads_per_task)

configs/nosite-x86_64/variables.yaml
scheibelp: Sorry here I'm talking about what Ramble needs defined up front.

I agree we should also require that users specify the number of CPUs per node.

Everything under "# Descriptions of resources available on systems" in `class AllocOpt(Enum):` should probably be defined on each system (unless e.g. that system doesn't have GPUs).

modifiers/allocation/modifier.py
scheibelp: I think we need both:

* N_CORES_PER_TASK, this is similar to NUM_OMP_THREADS and refers to the possibility that a task might create its own threads and wants more CPU resources based on this. You can also use both like `N_CORES_PER_TASK=1/NUM_OMP_THREADS=2` to imply that both of the OpenMP threads should be sharing a core
* N_CORES_PER_NODE is I think what you mean to be the system description: how many CPUs are on each node. I listed that as CPUS_PER_NODE below

As a general policy for naming I prefixed *requests* with `N_` and the resources *availability* variables don't have that. Maybe a different prefix should be chosen (although in general I think it's helpful to distinguish using a prefix).

modifiers/allocation/modifier.py
scheibelp: These are enum codes. There's nothing required about them other than that they are distinct. I also broadly divide them into categories for ease of reading.

modifiers/allocation/modifier.py
scheibelp: * I used the more generic term "timeout" to account for "batch_time". Right now, this can specify a value in minutes in any experiment/system config file, and the modifier can map that to an appropriate request format on slurm/flux/etc.
* Agree
* Agree

For the last two, we didn't have that before and I haven't added it yet in this PR but I think it would be straightforward to do.

modifiers/allocation/modifier.py
scheibelp: I think you mean in `configs/nosite-x86_64/variables.yaml` - yes I can do that

(although by default I was originally going to leave this blank we could that and through an error for any experiment that requests GPUs on `nosite-x86_64` - the systems that have GPUs we should set for sure)

modifiers/allocation/modifier.py
scheibelp: I use "task" and "rank" interchangeably.

In either case I mean "a process that represents a running instance of a program" and I thought task was slightly more generic (could account for cases where we want to run processes w/o MPI).

configs/nosite-x86_64/variables.yaml
pearce8: Would it be easier to find the values that weren't set if we set them to -1 initially (an invalid value by our definition of "positive integers")?

configs/nosite-x86_64/variables.yaml
pearce8: Thanks, that distinction between "what the machine has" and "what ramble needs to schedule the job" is more clear now.

modifiers/allocation/modifier.py
pearce8: Unfortunately "task" is an overloaded term in HPC, where "rank" is unambiguous, so we should just use "rank" everywhere.

configs/nosite-x86_64/variables.yaml
scheibelp: I was likewise hoping I could set them to something that was more obviously wrong, but Ramble expects them to be positive integers. I'll double check though to make sure.

modifiers/allocation/modifier.py
scheibelp: That is now done

configs/nosite-x86_64/variables.yaml
scheibelp: I have updated with new logic so we don't have to use some random number as a placeholder:

* There is now a `max_request` variable (which defaults to 1000 if unset)
* Requests above this are considered placeholders

Later on, the logic which fills in all unset variables also makes use of `max_request` to make sure it didn't accidentally request too much.

This particular thread I'll mark as resolved, because I think it's less confusing now.

modifiers/allocation/modifier.py
scheibelp: I think the renaming of variables (generally using "rank" instead of "task") has made this less confusing (and I added both variables to clarify that each is useful).

configs/CSC-LUMI-HPECray-zen3-MI250X-Slingshot/variables.yaml
```diff
@@ -6,12 +6,13 @@
...
-  batch_time: '02:00'
-  mpi_command: 'srun -N {n_nodes} -n {n_ranks}'
-  batch_submit: 'sbatch {execute_experiment}'
-  batch_nodes: '#SBATCH -N {n_nodes}'
-  batch_ranks: '#SBATCH -n {n_ranks}'
-  batch_timeout: '#SBATCH -t {batch_time}:00'
-  cpu_partition: '#SBATCH -p small'
-  gpu_partition: '#SBATCH -p small-g' 
+  timeout: '120'
+  scheduler: "slurm"
```
pearce8, line 10: I think we should preface scheduler-related names with "batch_", so "timeout" would become "batch_timeout".

We need at least 2 more scheduler variables:
- batch_partition_name   (on LC, this would be 'pbatch')
- batch_bank (on LC, lets set this to 'guest')

Ideally, the user should be able to overwrite the latter two from their experiments script - is that possible?

configs/nosite-x86_64/variables.yaml
```diff
@@ -4,9 +4,12 @@
...
 variables:
-  batch_time: ''
-  mpi_command: 'mpirun -n {n_nodes} -c {n_ranks} --oversubscribe'
-  batch_submit: '{execute_experiment}'
-  batch_nodes: ''
-  batch_ranks: ''
-  batch_timeout: ''
+  n_ranks: '1000001'  # placeholder value
+  n_nodes: '1000001'  # placeholder value
+  scheduler: "sierra"
```
pearce8: The scheduler on Sierra is LSF (the full name is IBM's Spectrum LSF).

Docs here if you need/are interested:
https://hpc.llnl.gov/documentation/tutorials/using-lc-s-sierra-systems#Monitor

experiments/gromacs/cuda/ramble.yaml
```diff
@@ -18,13 +18,11 @@ ramble:
...
               OMP_PLACES: cores
-              OMP_NUM_THREADS: '{omp_num_threads}'
           variables:
             experiment_setup: ''
-            n_ranks: '{processes_per_node} * {n_nodes}'
-            processes_per_node: '2'
+            n_ranks_per_node: '2'
             n_nodes: '4'
-            omp_num_threads: '10'
+            n_threads_per_proc: '10'
```
pearce8, line 28: So, there is only 2 ways in current use to launch multiple threads on the CPUs:
- pthreads (super low level, rarely if ever used by the application - although the tools might use these)
- OpenMP threads, which is what is used here.  OMP_NUM_THREADS is an environment variable everyone would immediately recognized, and I think we can reasonably assume that this is the only type of threads people will use on HPC CPUs.  I wonder if it is worth it to stick with the convention here for familiarity, and keep the omp_num_threads name.

pearce8 (changes requested):

configs/CSC-LUMI-HPECray-zen3-MI250X-Slingshot/variables.yaml
scheibelp: > We need at least 2 more scheduler variables:
> batch_partition_name (on LC, this would be 'pbatch')

I used the term "queue" for this (i.e. I translate the `queue` variable to `-p` for `srun`, if `queue` was set), do you want it to in fact be `batch_partition_name`? Or would `batch_queue` work? (any are fine with me, I'm just curious which is the more general term)

> batch_bank (on LC, lets set this to 'guest')

Does this translate to `-A` for `srun`? I don't need to know all the translations, but I'm not clear where to figure out what they are (and I assume we'll want to be able to do this for all the schedulers)

experiments/gromacs/cuda/ramble.yaml
scheibelp: I think users will need to look up the list of experiment variables they need to set either way; I figured when they read through this, that `n_threads_per_proc` would be clear. And as you say, my main intent was to not constrain the term to one threading model.

FWIW I don't think anything is stopping me from adding the possibility of using `omp_num_threads` as an alias of `n_threads_per_proc` i.e. allow both, if we want the user to be able to guess what variables they set.

configs/RCCS-Fugaku-Fujitsu-A64FX-TofuD/variables.yaml
```diff
@@ -4,13 +4,15 @@
...
-  batch_submit: 'pjsub {execute_experiment}'
-  batch_nodes: '#PJM -L "node={n_nodes}"'
-  batch_ranks: '#PJM --mpi proc={n_ranks}'
-  batch_timeout: '#PJM -L "elapse={batch_time}:00" -x PJM_LLIO_GFSCACHE="/vol0001:/vol0002:/vol0003:/vol0004:/vol0005:/vol0006"'
   default_fj_version: '4.8.1'
   default_llvm_version: '15.0.3'
   default_gnu_version: '12.2.0'
-
+  timeout: "120"
+  scheduler: "slurm"
```
pearce8: Fugaku uses the Fujitsu scheduler, PJM.

configs/CSC-LUMI-HPECray-zen3-MI250X-Slingshot/variables.yaml
scheibelp: > Ideally, the user should be able to overwrite the latter two from their experiments script - is that possible?

You mean a customized `execute_experiment.tpl`? 

* FWIW, once I fold in the logic to combine multiple `variables.yaml` files, the appropriate `*_instructions` function can turn `batch_partition/bank` into arguments for the system, and there could be a single file associated with the user that sets these (and benchpark could combine it with each the system `variables.yaml`)
* But I could also add the option for a customized `execute_experiment.tpl` file (e.g. if there's an option that should be applied for all users of the system)

I think either/both of these would be good follow-up PRs (I want to add the understanding of these variables here, but split up the configuration later).

experiments/gromacs/cuda/ramble.yaml
scheibelp: I've added `omp_num_threads` as an alias (and added a mechanism for alias processing, but if we decide we just want one name for everything, I can remove `n_threads_per_proc` and the alias logic).

modifiers/allocation/modifier.py
```diff
@@ -0,0 +1,415 @@
...
+        # TODO: this might have to be an option on the batch_submit vs.
+        # a batch directive
+        if v.queue:
+            batch_opts.append(f"-q {v.queue}")
+        if v.timeout:
+            batch_opts.append(f"-W {TimeFormat.as_hhmm(v.timeout)}")
+
+        batch_directives = list(f"#BSUB {x}" for x in batch_opts)
+
+        v.mpi_command = f"lrun {' '.join(cmd_opts)}"
```
nhanford, line 387: Unfortunately `lrun` is a Livermore-Specific tool. Under the covers, it runs `jsrun`, which should work on all IBM Spectrum LSF-scheduled machines with CSM and JSM enabled.
You can see the underlying `jsrun` command generated by `lrun` with the `-v` flag.
Because of the complexity of our `jsrun` calls on Lassen, it may make sense to leave this in and just make note that it is Livermore-specific.
There is some `jsrun` docmentation [here](https://www.ibm.com/docs/en/SSWRJV_10.1.0/jsm/jsm104_pdf.pdf)

scheibelp: > Because of the complexity of our jsrun calls on Lassen, it may make sense to leave this in and just make note that it is Livermore-specific.

That would be my preference: I think it's fine to offer command/batch generators for Livermore-specific systems within this file.

configs/CSC-LUMI-HPECray-zen3-MI250X-Slingshot/variables.yaml
pearce8: yes, lets go with:
batch queue (-p for srun, pbatch as the default value)
batch bank (-A for srun, guest as the default value on LC)

pearce8 (approved):

PR #196, opened 2024-04-08
Benchpark: use requirements.txt from Ramble

scheibelp:
Update ramble version used by Benchpark (primarily to get `https://github.com/GoogleCloudPlatform/ramble/pull/452`); append requirements.txt for Ramble to benchpark requirements.txt

This means that users will automatically get everything they need to run Ramble instances created by Benchpark as part of installing the `requirements.txt` for Benchpark itself. If users explicitly update a Ramble instance to a later version, they may need to install additional dependencies.

As long as Ramble doesn't use deprecated library methods, the latest requirements.txt from Ramble should be sufficient for running older Ramble versions as well (if that starts happening, we'll need to change our strategy to maintain a Python venv for each Ramble instance that we use).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,6 +48,7 @@ jobs:
               - '.github/**'
               - 'bin/**'
               - 'configs/**'
+              - 'checkout-versions.yaml'
               - 'experiments/**'
               - 'repo/**'
             license:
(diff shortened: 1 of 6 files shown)


scheibelp:
The CI pre-checks assumed that CI did not need to run. From testing this manually I see there is an issue - I've marked this as draft in the meantime

scheibelp:
Ready now, and some of the changes I made forced a CI run. As it is, specifically changing the requirements.txt might not prompt a rerun.

scheibelp:
> As it is, specifically changing the requirements.txt might not prompt a rerun.

I think I fixed that with 39fbbe6

pearce8 (approved):

PR #199, opened 2024-04-15
Initial commit of MPI-only Laghos with Sedov blast wave experiment

wdhawkins:
This PR adds an mpi-only version of Laghos (https://github.com/CEED/Laghos) with a Sedov blast wave experiment.  

Although all of the experiments run successfully, FOM and success criteria are not working correctly. I'm currently looking into this.


repo/laghos/package.py
```diff
@@ -0,0 +1,82 @@
...
+    # See lib/spack/spack/build_systems/makefile.py
+    def check(self):
+        with working_dir(self.build_directory):
+            make("test", *self.build_targets)
+ 
+    def install(self, spec, prefix):
+        mkdirp(prefix.bin)
+        install("laghos", prefix.bin)
+
+    @run_after("install")
```
pearce8: @scheibelp Is run_after("install") the correct way to copy data files after installing the binary?

scheibelp: This should work fine. Is there a failure at this step?

Using `cache_extra_test_sources` assumes the files live in `.../stage/spack-src/data` and that you want them in `<install-prefix>/data`

wdhawkins: I tested this on a fresh `benchpark` install.  It looks to me like the bit of code that I hoped would  copy the `data` directory to `install-prefix/data` actually copied it to `install-prefix/.spack/test/data`.

scheibelp: Whoops sorry yes: you'll need to use `install_tree` then (which allows setting up a `src` and a `dst`):

```
install_tree("data", prefix.data)
```

should do it

wdhawkins: Thanks @scheibelp! That fixed the problem.

pearce8 (approved):

PR #209, opened 2024-04-25
Adding docs on using Ramble workspaces

### LINKED ISSUE #176: Improve documentation on benchpark/repo
Currently a sub-bullet under "Add a benchmark" - but it is hidden by default, so maybe we need to make it more prominent?  Users are clearly not finding it.

It is a little dense at the moment, perhaps can write more clearly?

Explain the staging:  package.py and/or application.py in benchpark/repo overwrite Spack and Ramble built-ins.  Intent: getting things working.

Once working, the goal is to upstream package.py to Spack (and remove from Benchpark once available in Spack), and application.py to Ramble.

pearce8:


docs/5-build-experiment.rst
```diff
@@ -19,17 +19,16 @@ Ramble will build the source code and set up the following workspace directory s
...
-                        configs/
-                            (everything from source/configs/<system>)
-                            (everything from source/experiments/<benchmark>)
+        <Benchmark/ProgrammingModel>/
+            <System>/
+                workspace/
+                    configs/
+                        (everything from source/configs/<System>)
+                        (everything from source/experiments/<Benchmark/ProgrammingModel>)
                         experiments/
```
scheibelp: (issue) This `experiments` dir (and all subdirs of it) should be dedented one level (to be aligned with `configs/`.

docs/add-a-benchmark.rst
```diff
@@ -7,27 +7,34 @@
...
+so that you can test them together with your experiment specification on a given system::
 
   benchpark
-  └── repo 
-     ├── ${BENCHMARK1} 
-     │  ├── application.py 
-     │  └── package.py 
-     └── repo.yaml 
+  └── repo
+     ├── ${Benchmark1}
```
scheibelp, line 31: (recommendation) Might be useful to tie this to an example invocation like

```
benchpark setup ${Benchmark1}/${ProgrammingModel1} ${System1} /output/path/to/experiments_root
```

overall, I think this specification will be clearest if it leads by discussing that you can override Ramble/Spack descriptions; the rewording indicates to me that you're trying to avoid that, but to manipulate these at all, the user will have to be familiar with them anyhow.

scheibelp (changes requested):
One minor issue with updated indentation. I also have a recommendation in the `add-a-benchmark` section.

scheibelp (approved):

PR #216, opened 2024-04-29
ci: rebuild docs on changes to bin/benchpark

alecbcs:
Adds a rule to rebuild the docs when `bin/benchpark` (or any future file in bin) changes in a PR.

Fixes #210

PR #217, opened 2024-04-29
Tags command should use benchmarks not experiments

alecbcs:
Bug introduced in #190 queried experiments not benchmarks to generate tags which resulted in invalid key errors when running the command and manifested as an empty table in the docs.


alecbcs:
#216 should be merged first to trigger a rebuild of the docs when this merges.

pearce8 (approved):

Issue #207, opened 2024-04-22
intermittent xz/openssl fetch issues on LLNL Poodle with nosite-x86_64 saxpy/openmp 


rountree:
**Big picture:**

For the past several days I've hit intermittent issues with xv and openssl downloads when trying to build a nosite-x86_64 saxpy/openmp instance on LLNL's Poodle cluster.

I've been able to reproduce the issue in an isolated clone of spack.  Adjusting the connection timeout might have helped, but in digging further into the logs I wasn't able to nail down what might be taking so long (`fetch` for `xz` was taking over a minute whether it succeeded or failed, and it's not that big).  Cloning another spack instance into `/dev/shm` (to remove filesystem lag from the equation) resulted in ssl certificate errors.  Note that multiple other packages were being fetched and built correctly.

**From the top:**

I sourced the following into my environment.

```
module load python/3.11.5
export PATH=${HOME}/.local/bin:${PATH}
export BPROOT=${PWD}/benchpark
export PATH=${BPROOT}/bin:${PATH}
export WORKSPACE_DIR=${PWD}/workspace
export BPSITE=nosite-x86_64
export BPEXPR=saxpy/openmp
export SPACK_DISABLE_LOCAL_CONFIG=1
alias bp="benchpark"
```

I executed the following manually.
```
mkdir ${WORKSPACE_DIR}
git clone git@github.com:LLNL/benchpark.git
cd benchpark
pip install -r requirements.txt
bp setup ${BPEXPR} ${BPSITE} ${WORKSPACE_DIR}
. ${WORKSPACE_DIR}/setup.sh
ramble -P -D ${WORKSPACE_DIR}/${BPEXPR}/${BPSITE}/workspace workspace setup
```

Output of the `ramble` command:
```
rountree@poodle18 ~/w/poodle/bp/benchpark$ ramble -P -D ${WORKSPACE_DIR}/${BPEXPR}/${BPSITE}/workspace workspace setup
==> Warning: The following config sections are deprecated and ignored:
==> Warning:     spack:concretized
==> Warning: Please remove from your configuration files.
==> Streaming details to log:
==>   /g/g24/rountree/w/poodle/bp/workspace/saxpy/openmp/nosite-x86_64/workspace/logs/setup.2024-04-22_13.37.45.out
==>   Setting up 8 out of 8 experiments:
==> Experiment #1 (1/8):
==>     name: saxpy.problem.saxpy_512_1_8_2
==>     root experiment_index: 1
[...]

PR #219, opened 2024-04-30
Update spack version

rfhaque:
Update the spack version ussed by benchpark

### DIFF
--- a/checkout-versions.yaml
+++ b/checkout-versions.yaml
@@ -5,4 +5,4 @@
 
 versions:
   ramble: bb664f142b2cbdb2b2ea39e70a8535c9f27c1179
-  spack: 31de670bd26beca979ebd75dcb0ce90c535a78c4
+  spack: c2eef8bab26adb00b250992e29d697b4706356a0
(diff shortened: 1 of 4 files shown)


pearce8 (approved):

pearce8 (changes requested):
@scheibelp @alecbcs this PR appears to hit the same issue in CI that @rountree sees with his tests.

rfhaque:
@pearce8 Is this a bug in the concretizer or is it due to spack having deprecated a specific cmake version

scheibelp:
> scheibelp alecbcs this PR appears to hit the same issue in CI that rountree sees with his tests.

You're talking about https://github.com/LLNL/benchpark/issues/207? That is a fetcher error, this appears to be a [concretization error](https://github.com/LLNL/benchpark/actions/runs/8900373130/job/24441784894?pr=219)

> Is this a bug in the concretizer or is it due to spack having deprecated a specific cmake version

If we are explicitly requesting `cmake@3.23.1`, then that will fail with the latest spack (it was deprecated as of March 19 2024)

UPDATE: we are in fact requesting that version of `cmake` for `saxpy` right now: https://github.com/LLNL/benchpark/blob/develop/experiments/saxpy/openmp/ramble.yaml

pearce8 (approved):

PR #240, opened 2024-05-08
Streamlining terminology in docs

### LINKED ISSUE #213: Need to document spack and ramble version pinning
This can come up when adding or modifying a benchmark - and the package.py is already in Spack, but not yet in Benchpark.

- how to find the versions we are on
- potential side effects (e.g., version compatibility issues (e.g., package in tip of develop vs. our pinned version))
- what might happen if people manually get the tip of develop (and how they could do that if they really needed to)

pearce8:
Standardizing organization and naming of the subcomponents of Benchpark:
- Benchmarks
- System specifications (originally called system configs)
- Experiments

Reorganizing how-to information into a set of FAQ sub-pages.


README.rst
```diff
@@ -20,33 +20,39 @@ Benchpark is an open collaborative repository for reproducible specifications of
...
 
-* source repo and version
-* build (Spack) configuration
-* run (Ramble) configuration
+* Source repo and version
+* Build configuration (with `Spack <https://github.com/spack/spack>`_)
+* Run configuration (with `Ramble <https://github.com/GoogleCloudPlatform/ramble>`_)
 
-3. **Experiment specifications**
+3. **Experiment Specifications** (specific benchmark experiment on a specific system)
```
scheibelp: These don't explicitly specify a system. For now though they do implicitly encode system details because they record things like `compiler-gcc`.

README.rst
```diff
@@ -20,33 +20,39 @@ Benchpark is an open collaborative repository for reproducible specifications of
...
 
-* programming models to use for benchmarks on a given system type
-* valid experiments for benchmarks on a given system (scientific parameter studies, performance parameter studies, etc.)
+* Programming model (e.g., OpenMP, CUDA, ROCm) for the benchmark on a given system
+* Parameters for individual runs in a study
 
 Dependencies
 ------------
-Benchpark uses the following open source projects for specifying configurations:
+Benchpark uses the following open source packages for specifying configurations:
```
scheibelp: I would recommend using "project" over "package": "package" relates to spack terminology and implies we might be installing these with Spack.

README.rst
```diff
@@ -20,33 +20,39 @@ Benchpark is an open collaborative repository for reproducible specifications of
...
 
-* `Ramble <https://github.com/GoogleCloudPlatform/ramble>`_ to specify run configurations
-* `Spack <https://github.com/spack/spack>`_ to specify build configurations
+* `Spack <https://github.com/spack/spack>`_ for build configurations
```
scheibelp: I would recommend "for building benchmark and dependencies" over "for build configurations": "configurations" implies something else is doing the actual build.

It might be worth mentioning that Spack currently only exists as a dependency of Ramble: we don't drive it directly.

docs/add-a-site-specific-system-config.rst
```diff
@@ -4,7 +4,7 @@
    SPDX-License-Identifier: Apache-2.0
 
 ======================================
-Adding a Specific System Configuration
+Adding a Specific System Specification
```
scheibelp: Mildly awkward wording (a "specific specification"). "configuration" seemed more appropriate, or "Adding a System Specification".

docs/index.rst
```diff
@@ -17,13 +17,19 @@
...
+   FAQ-spack-ramble-dependence
+   FAQ-benchpark-repo
 
 .. toctree::
    :maxdepth: 1
    :caption: Catalogue
 
-   catalogue-of-benchmarks
    available-system-specs
    benchmark-list
```
scheibelp, line 34: This file was renamed, does the reference need to be updated?

docs/FAQ-benchpark-repo.rst
```diff
@@ -0,0 +1,27 @@
+.. Copyright 2023 Lawrence Livermore National Security, LLC and other
+   Benchpark Project Developers. See the top-level COPYRIGHT file for details.
+
+   SPDX-License-Identifier: Apache-2.0
+
+=================================
+Benchmark Specs in benchpark/repo
```
scheibelp: IMO any subsection title within an FAQ section should be a question, or it should be phrased like "how to do X". As phrased, I don't think it's clear to users what they could achieve by reading through these `FAQ-*` sections.

scheibelp (changes requested):
IMO it would be worth putting the high-level intent in the PR description: at least one of the high-level goals seems to be standardizing organization and naming of the subcomponents of benchpark:

* Benchmarks
* System specifications (originally called system configs)
* Experiments

Another appears to be reorganizing how-to information into a set of FAQ sub-pages (I have a comment that we might want to change the titles of these pages).

PR #242, opened 2024-05-10
Separate CUDA spec in ramble.yaml

rfhaque:
In some cases, spack does not pick the correct compiler if the cuda spec is specified as a part of the amg2023 spack spec in ramble.yaml.


rfhaque:
Resolved in https://github.com/LLNL/benchpark/pull/243

PR #248, opened 2024-05-22
remove step to remove gcc 13

scheibelp:
Trying to resolve CI failures like https://github.com/LLNL/benchpark/actions/runs/9130896803/job/25108882636?pr=246  in https://github.com/LLNL/benchpark/pull/246

### DIFF
--- a/.github/workflows/run.yml
+++ b/.github/workflows/run.yml
@@ -6,10 +6,6 @@ jobs:
   saxpy:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


pearce8 (approved):

alecbcs:
This looks good to me. @scheibelp did you check to see what versions of GCC are present on the latest Ubuntu image?

I'd excluded `gcc@13` before since they didn't have a corresponding `g++` for it at the time. We might want to check to make sure there's a compatible `g++` for all of the versions of `gcc` in the image now.

PR #258, opened 2024-06-01
adding pre and post processing options which are unrelated to the ind…

jdomke:
…ividual benchmark

the issue is that fujitsu-mpi does not redirect to stdout, and on fugaku a usual workaround via -of|-std flags is currently disabled by the ops team and therefore we can at best redirect to a merged file of stdout+strerr for each rank via -std-proc

so, the general ramble approach of `mpi_command exec >> log_file` does not work; pre_cmd and post_cmd don't work either, because they are part of the benchmark or part of a modifier (which isnt the case for an mpi)

currently, only way i see to add this is via these pre_process and post_process variables which might be useful for others in the future, too

redirect mpi output on fugaku


scheibelp:
I'm currently working on a PR to address the conflicts here as well as add a few other conveniences on top of #195. I should have it ready by end of Monday (June 10th)

pearce8:
@jdomke Please test to check if the newly merged PR addresses the issue you were solving in this PR. (https://github.com/LLNL/benchpark/pull/276)

PR #261, opened 2024-06-05
[WIP] Add command: `benchpark system create`

scheibelp:
Closes https://github.com/LLNL/benchpark/pull/110
Based on https://github.com/LLNL/benchpark/pull/195 (not strictly needed, but defines variables only actually used in that PR)

(note: ~500 of the +/- of this diff is from moving `bin/benchpark` to `lib/main.py` - I recommend collapsing these to make the diff more-readable)

Example:

```
./bin/benchpark system create --basedir=test-system-create --type=aws --set instance-type=c4.xlarge
```

Creates a directory in `test-system-create`, and writes a `variables.yaml` file to it:

```
variables:
  scheduler: "mpi"
  sys_cores_per_node: "4"
  sys_mem_per_node: 7.5
  ...
```

TODOs

- [ ] `--basedir` argument can generate an entirely new system config directory, but `benchpark setup` can't actually use it yet
- [ ] This should accommodate the possibility of `packages`/`compilers`config
- [ ] This has a `--from` option to allow taking an existing config, but doesn't support it yet. I think it would be useful to combine this option with `--set` to override settings from that file


scheibelp:
@douglasjacobsen FYI

scheibelp:
This has moved to https://github.com/LLNL/benchpark/pull/298 (uses a branch from within Benchpark vs. my fork of it)

PR #265, opened 2024-06-06
Update Spack and Ramble versions to 6/5/2024 develop

pearce8:


scheibelp (approved):
Looks like this doesn't need to update requirements.txt, looking at https://github.com/GoogleCloudPlatform/ramble/blame/develop/requirements.txt

Ramble's spack.yaml `spack:concretized` was deprecated in https://github.com/GoogleCloudPlatform/ramble/commit/c9b8c49b286adfd6812273f23b1bf2c29cb684e8 (and removed recently in  https://github.com/GoogleCloudPlatform/ramble/commit/7de4373542441a6383ebee63fe8389f7251bbd25#diff-04c5d2bb0120186d2c53c6caa17e537cc0a5cd2bffd15557831d7bea293eff77)

It is my assessment that for the operations we do, we do not have to add anything after removing `spack:concretized:true` because the operations we tell users to run at https://github.com/LLNL/benchpark/blob/develop/docs/FAQ-what-to-rerun.rst  would already do this.

PR #274, opened 2024-06-10
Avoiding Spack error

pearce8:


scheibelp:
This is to avoid Benchpark configs running into https://github.com/spack/spack/issues/44533, once that issue is resolved, we can advance the pinned Spack commit.

PR #278, opened 2024-06-13
Removing invalid -g option from BSUB

pearce8:


modifiers/allocation/modifier.py
```diff
@@ -370,9 +370,6 @@ def lsf_instructions(self, v):
             cmd_opts.append(f"-n {v.n_ranks}")
         if v.n_nodes:
             batch_opts.append(f"-nnodes {v.n_nodes}")
-        if v.n_gpus:
-            gpus_per_rank = self.gpus_as_gpus_per_rank(v)
-            batch_opts.append(f"-g {gpus_per_rank}")
```
scheibelp, line 375: If this is removed with nothing added, then GPU experiments aren't expressing any request for GPUs when using this scheduler, e.g. on Sierra)

I think the simplest thing to add here is to append this to `cmd_opts` instead of `batch_opts`, that means it will be an argument to lrun (from looking at `man lrun`, it appears to support this option).

I assumed it would translate directly as a batch directive, but the format is different: https://www.ibm.com/docs/en/spectrum-lsf/10.1.0?topic=jobs-example-gpu-job-submissions. 

With a bit more work, it looks like this could in fact be expressed as a batch directive like:

`f"num={v.n_gpus}:mode=exclusive_process"`

pearce8: The "exclusive process" has to do with sharing the GPU among different processes.  The technologies to do so are far from portable or stable at the moment.  Since we are currently not sharing GPUs between CPU processes, lets not open that can of worms.

scheibelp (approved):

PR #287, opened 2024-06-14
Adding more OSU microbenchmarks workloads

pearce8:
@gracenansamba should this PR be closed?


gracenansamba:
not yet, I would like to work on OSU benchmark some more

PR #298, opened 2024-06-25
Add command: `benchpark system init`

scheibelp:
Closes https://github.com/LLNL/benchpark/pull/110

(note: ~500 of the +/- of this diff is from moving `bin/benchpark` to `lib/main.py` - I recommend collapsing these to make the diff more-readable)

Add a new command `benchpark system init`, that can generate a system config (a collection of files in a directory) that you can then pass to `benchpark setup` like:

```
benchpark setup saxpy/openmp <the-directory-created-by-benchpark-system-init> workspace/
```

### Example 1:

`./bin/benchpark system init --basedir=test-system-create aws instance_type=c4.xlarge`
(old) ~`./bin/benchpark system create --basedir=test-system-create --type=aws --set instance-type=c4.xlarge`~

Creates a directory in `test-system-create`, and writes a `variables.yaml` file to it:

```
variables:
  scheduler: "mpi"
  sys_cores_per_node: "4"
  sys_mem_per_node: 7.5
  ...
```

This doesn't tie in with `boto` yet, so doesn't provide a means to generate a description for all instance types. IMO that would be better to handle in a separate PR.

### Example 2:

`./bin/benchpark system init --basedir=test-system-create tioga rocm=551 compiler=cce ~gtl`
(old) ~`./bin/benchpark system create --basedir=test-system-create --type=tioga`~

will assemble a `packages.yaml` config such that:

* there is only one set of ROCm packages (either all 5.4.3, or all 5.5.1, but not a mix of them)
* the MPI external chosen matches the compiler choice

(Update August 3rd: done) ~This doesn't yet generate a `variables.yaml` or a compiler config, so the result is incomplete.~

### TODOs

- [x] `--basedir` argument can generate an entirely new system config directory, but `benchpark setup` can't actually use it yet
- [x] This should accommodate the possibility of `packages`/`compilers`config
  - As of [296e0e1](https://github.com/LLNL/benchpark/pull/298/commits/296e0e15fc9891ee8fe83393e3796a54a45c1bc9), this provides a means to assemble a set of individual package configs into a single config
- [x] ~This has a `--from` option to allow taking an existing config, but doesn't support it yet. I think it would be useful to combine this option with `--set` to override settings from that file~
  - (update August 6th: this is no longer relevant: you can now create system classes that fully replicate an existing directory in `configs/`)


lib/benchpark/sys_repo/systems/tioga/system.py
```diff
@@ -0,0 +1,21 @@
...
+
+class Tioga(System):
+    def __init__(self):
+        super().__init__()
+
+        self.scheduler = "flux"
+        self.sys_cores_per_node = "64"
+        self.sys_gpus_per_node = "4"
+
+        base = pathlib.Path(os.path.abspath(__file__)).parents[0]
```
scheibelp: Currently 

`./bin/benchpark system create --basedir=test-system-create --type=aws --set instance-type=c4.xlarge`

works, but not

`./bin/benchpark system create --basedir=test-system-create --type=tioga`

does not, because `__file__` is not available in this context: I'll need to figure out how to make that available.

lib/benchpark/test_repo/experiments/saxpy/experiment.py
```diff
@@ -0,0 +1,68 @@
...
+    )
+
+    def compute_applications_section(self):
+        variables = {}
+        matrix = {}
+
+        # GPU tests include some smaller sizes
+        n = ["512", "1024"]
+        matrix = ["size"]
+        if self.spec.satisfies("openmp"):
```
becker33:
```suggestion
        if self.spec.satisfies("programming_model=openmp"):
```

bin/benchpark
```diff
@@ -5,546 +5,16 @@
...
-    elif args.application:
-        if benchpark_check_benchmark(args.application):
-            benchpark_experiments_tags = helper_experiments_tags(ramble_exe, benchmarks)
-            print(benchpark_experiments_tags[args.application])
-    else:
-        benchpark_experiments_tags = helper_experiments_tags(ramble_exe, benchmarks)
-        print("All tags that exist in Benchpark experiments:")
-        for k, v in benchpark_experiments_tags.items():
-            print(k)
+    basedir = pathlib.Path(os.path.abspath(__file__)).parent.parent
```
becker33, line 15: For a follow-on PR, we should make this bilingual like we did for the equivalent Spack script

lib/benchpark/error.py
```diff
@@ -0,0 +1,81 @@
...
+
+import inspect
+import sys
+
+#: at what level we should write stack traces or short error messages
+#: this is module-scoped because it needs to be set very early
+debug = 0
+
+
+class BenchparkError(Exception):
```
becker33, line 11: I think we can delete the entire body of this class if we inherit from `ramble.error.RambleError`. Alternatively, I also don't think we're using any of the complex error handling from Spack/ramble that requires `Error.long_message`, so we could just make this a simple class that inherits from `Exception` and adds no logic. I'm leaning towards the latter.

lib/benchpark/experiment.py
```diff
@@ -0,0 +1,196 @@
...
+
+import ramble.language.language_base
+import ramble.language.language_helpers
+
+bootstrapper = benchpark.runtime.RuntimeResources(benchpark.paths.benchpark_home)
+bootstrapper.bootstrap()
+
+
+class Experiment(SpecTemplate):
+    """This is the superclass for all benchpark experiments.
```
becker33, line 24: Looks like we only got halfway through this docstring in terms of updating it to be benchpark instead of spack.

lib/benchpark/experiment.py
```diff
@@ -0,0 +1,196 @@
...
+        the given spec.
+
+        If ``self.spec`` is not concrete, this returns whether this package may
+        extend the given spec.
+        """
+        if spec.name not in self.extendees:
+            return False
+        s = self.extendee_spec
+        return s and spec.satisfies(s)
+
```
becker33:
```suggestion
```

This is all leftover from Spack that we aren't using anywhere

lib/benchpark/experiment.py
```diff
@@ -0,0 +1,196 @@
...
+import benchpark.paths
+import benchpark.repo
+import benchpark.runtime
+import benchpark.variant
+
+import ramble.language.language_base
+import ramble.language.language_helpers
+
+bootstrapper = benchpark.runtime.RuntimeResources(benchpark.paths.benchpark_home)
+bootstrapper.bootstrap()
```
becker33:
```suggestion
bootstrapper = benchpark.runtime.RuntimeResources(benchpark.paths.benchpark_home)
bootstrapper.bootstrap()

import ramble.language.language_base
import ramble.language.language_helpers
```
Need to bootstrap before we import from ramble

lib/benchpark/experiment.py
```diff
@@ -0,0 +1,196 @@
...
+
+from typing import Dict, Tuple
+import yaml  # TODO: some way to ensure yaml available
+
+from benchpark.directives import SpecTemplate
+import benchpark.spec
+import benchpark.paths
+import benchpark.repo
+import benchpark.runtime
+import benchpark.variant
```
becker33, line 14:
```suggestion
import benchpark.paths
import benchpark.runtime
```
Some of these imports either aren't used or won't be once we delete vestigial code

lib/benchpark/directives.py
```diff
@@ -0,0 +1,203 @@
...
+            raise DirectiveError(directive, msg.format(pkg.name, name))
+
+        pkg.variants[name] = benchpark.variant.Variant(
+            name, default, description, values, multi, validator, sticky
+        )
+
+    return _execute_variant
+
+
+class SpecTemplate(metaclass=DirectiveMeta):
```
becker33: I would call this `ObjectTemplate` instead of `SpecTemplate`, since it's the template for the objects that the spec describes, rather than for the specs themselves.

lib/benchpark/directives.py
```diff
@@ -0,0 +1,203 @@
...
+    @classproperty
+    def package_dir(cls):
```
becker33:
```suggestion
    def code_dir(cls):
```

lib/benchpark/directives.py
becker33: Throughout this class we need to update comments to not be spack-specific

lib/benchpark/repo.py
```diff
@@ -0,0 +1,201 @@
...
+
+# Experiments
+def _exprs():
+    """Get the singleton RepoPath instance for Ramble.
+
+    Create a RepoPath, add it to sys.meta_path, and return it.
+
+    TODO: consider not making this a singleton.
+    """
+    experiments_repo = _base_path() / "test_repo"
```
becker33: We should fix this to properly configure a path for the repo and put it in a more appropriate place

lib/benchpark/runtime.py
```diff
@@ -0,0 +1,147 @@
...
+    initial_dir = os.getcwd()
+    try:
+        os.chdir(location)
+        yield
+    finally:
+        os.chdir(initial_dir)
+
+
+def git_clone_commit(url, commit, destination):
+    run_command(f"git clone -c feature.manyFiles=true {url} {destination}")
```
becker33, line 28: This should probably add `--depth=2` to go faster. Depth 2 is better than depth 1 because git sets some unfriendly configuration up automatically when depth=1.

lib/benchpark/runtime.py
```diff
@@ -0,0 +1,147 @@
+from contextlib import contextmanager
+import os
+import pathlib
+import shlex
+import subprocess
+import sys
+import yaml
+
+DEBUG = True
```
becker33: This seems vestigial

lib/benchpark/sys_repo/systems/tioga/externals/rocm/00-version-543-packages.yaml
```diff
@@ -0,0 +1,73 @@
+packages:
+  hipfft:
```
becker33: Should these configs have `buildable: false` for the rocm components?

lib/benchpark/sys_repo/systems/tioga/compilers/rocm/01-rocm-543-compilers.yaml
```diff
@@ -0,0 +1,26 @@
+compilers:
```
becker33: We can combine this into the `packages.yaml` section now that we're using a Spack version that supports that.

lib/benchpark/sys_repo/systems/tioga/system.py
```diff
@@ -0,0 +1,130 @@
...
+        "rocm",
+        default="551",
+        values=("543", "551"),
+        description="ROCm version",
+    )
+
+    variant(
+        "compiler",
+        default="rocm",
+        values=("gcc", "cce"),
```
becker33:
```suggestion
        default="rocm",
        values=("rocm", "gcc", "cce"),
```

lib/benchpark/system.py
```diff
@@ -0,0 +1,177 @@
...
+from benchpark.directives import SpecTemplate
+import benchpark.repo
+from benchpark.runtime import RuntimeResources
+
+# Duplicate imports from experiment.py
+from typing import Dict, Tuple
+import benchpark.spec
+import benchpark.variant
+import ramble.language.language_helpers
+import ramble.language.shared_language
```
becker33: These need to go after the bootstrap

lib/benchpark/test_repo/repo.yaml
```diff
@@ -0,0 +1,2 @@
+repo:
```
becker33: We need to move this repo to a reasonable place now that it works

lib/test/benchpark/directives.py
```diff
@@ -0,0 +1,4 @@
+# Copyright 2023 Lawrence Livermore National Security, LLC and other
```
becker33: Does this do anything?

becker33 (changes requested):
Most of these are pretty simple/mechanical changes.

The only big picture issue is the `benchpark system create` name. I think it would be better UX to name this `benchpark system init` -- in the long run I think users are unlikely to think of the yaml on disk as the "system" that is created, since they'll be interacting with the python interface. So I think it's better to preserve the `benchpark system create` name to create the python files later (which is also analogous to Spack and Ramble).

We should also eventually provide a mechanism to init the system as part of the setup command, but that's a separate PR.

lib/benchpark/sys_repo/systems/tioga/externals/rocm/00-version-543-packages.yaml
scheibelp: I did that in the shared part: `base/00-packages.yaml` (https://github.com/LLNL/benchpark/pull/298/files#diff-1eb6ee20863e93ac4e58d78e98984b493b337d7732af2b9bc7ff4dc7824753b0) it combines with a config that sets the versions.

lib/test/benchpark/directives.py
scheibelp: Not that I know of: it came from https://github.com/LLNL/benchpark/pull/299/files#diff-ce8ab53a2b3008e27988e961de694d655e49cc693f625e6d82c19a437715cb19

(I will remove it)

lib/benchpark/sys_repo/systems/tioga/system.py
scheibelp: I actually don't handle `rocm` at all as a compiler choice (although like I said I'm not 100% clear on what compiler mixtures are good), so I removed that as the default choice.

lib/benchpark/experiment.py
scheibelp: I removed the parts that seemed spack-specific  (essentially the last 2/3rds of it) and finished off this docstring in a generic (but relevant) way.

lib/benchpark/repo.py
scheibelp: I moved this to `<base>/var/exp_repo`

lib/benchpark/test_repo/repo.yaml
scheibelp: (I think this is resolved along with https://github.com/LLNL/benchpark/pull/298#discussion_r1714436173, but IMO that comment tracks this same issue if it is not resolved)

lib/benchpark/runtime.py
scheibelp: (not addressed yet) Does the `--depth=2` setting make it easy to check out other branches and pull full history if needed? If not, I'd prefer to keep the full depth.

lib/benchpark/directives.py
scheibelp: I've replaced all such references ("package" is replaced with "System/Experiment").

lib/benchpark/experiment.py
scheibelp: Our style checker should detect these (and generally has as I refactored things). Some of the references are non-obvious, e.g. `import benchpark.variant` is only used by the type enforcement logic, but the import is still needed for that.

lib/benchpark/directives.py
scheibelp: (not addressed yet) I'm a bit leery of calling anything `Object*`. IMO it makes sense to call `Experiment`s/`System`s "spec templates" because the `system.py` lays out all the variables that you freeze to get a final spec. Overall though I figured it was going to be challenging to find an agreeable term that abstracts this aspect of packages/system/experiments, so I'm definitely interested in suggestions (I prefer `SpecTemplate` to `ObjectTemplate` though).

lib/benchpark/directives.py
scheibelp: I ended up calling it `template_dir` (in keeping w/ my reasoning at https://github.com/LLNL/benchpark/pull/298#discussion_r1716143929).

lib/benchpark/sys_repo/systems/tioga/compilers/rocm/01-rocm-543-compilers.yaml
scheibelp: (not addressed yet) I wasn't sure if:

* Every feature of `compilers.yaml` was supported by `packages.yaml`
* Putting these entries into `packages.yaml` would be seamless in terms of Ramble integration

Are both of these obviously not an issue? I was thinking that moving these was the right idea, but I wanted to handle it after this PR.

scheibelp:
@becker33 I have renamed `benchpark system create` to `benchpark system init` and

* all comments that resolved trivially I took the liberty of collapsing
* If I responded but made no change (e.g. I liked your suggestion but wanted to punt it) I put "not addressed yet" at the front
* Other comments that I think I resolved but involved some discretion, I responded to but did not collapse

Also I don't think I mentioned it yet, but the style checker was giving me trouble trying to block this on style issues in the system/experiment repos. I had some trouble getting it to ignore those, so I just fixed those issues for now. IMO we shouldn't apply the same stringency to system/experiment definitions as we do to the core (but if you disagree, then this PR is in the right place).

lib/benchpark/runtime.py
becker33: Yes, that's specifically why I suggested `depth=2` instead of `depth=1`. With `depth=1` git sets things that make it impossible to checkout other branches without modifying config. With `depth=2` we just need to fetch (to get the missing refs) before we can checkout other branches.

lib/benchpark/sys_repo/systems/tioga/compilers/rocm/01-rocm-543-compilers.yaml
becker33: I'm fine with doing this in a next PR as long as we're going to move it long-term. But yes, this should be seamless and the two interfaces have feature parity.

alecbcs:
@scheibelp can we move `lib/benchpark/system_cmd.py` -> `lib/benchpark/cmd/system.py` ?

lib/benchpark/directives.py
scheibelp: This is now called `ExperimentSystemBase` as discussed.

lib/benchpark/runtime.py
scheibelp: Do you mind if this happens in a later PR?

scheibelp:
@becker33 `SpecTemplate` is renamed
@alecbcs `system_cmd` was moved

The only remaining comment is https://github.com/LLNL/benchpark/pull/298#discussion_r1722342188, which I'd like to address in a different PR.

lib/benchpark/runtime.py
becker33: Yep fine with that

becker33 (approved):

PR #317, opened 2024-07-22
Progress on adding some benchmarks to benchpark

august-knox:
Just checking in on some of my progress. I changed quicksilver to run on multiple nodes, and created an implementation of the ior benchmark (not too sure about the FOMs though). I also added remhos, although it is unable to run on the LLNL-Tioga-HPECray-zen3-MI250X-Slingshot system configuration due to issues with the external mvapich package. This seems like a known issue though.


pearce8:
Attempting to capture the status on Ruby, Lassen, Tioga.  OpenMP versions for 4 benchmarks:
- **Quicksilver**: working on Ruby, Lassen, Tioga.  Added strong and weak scaling.
- **IOR**: working on Ruby, Lassen, Tioga.  Writing strong and weak scaling.
- **Branson**: working on Ruby and Tioga.  MPI-related error on Lassen?  @rfhaque, please take a look
- **Remhos**: working on Ruby and Lassen.  Hypre compiler error on Tioga?  @rfhaque, please take a look

Question about Branson input file:
- In order to change the load, you must change the number of photons, which is a value specified within the input file. Would you know of any ways to potentially edit the file for each individual run?  (if there isn't a natural way to do this, we might ask Branson developers for a parameterized solution).

experiments/quicksilver/openmp/ramble.yaml
```diff
@@ -0,0 +1,75 @@
...
+                X: '32'
+                Y: '32'
+                Z: '16'
+
+                I: ['2','2','4','4']
+                J: ['2','2','2','4']
+                K: ['1','2','2','2']
+  software:
+    packages:
+      mpi:
```
rfhaque: @august-knox @pearce8 The mpi package spec needs to be removed. We treat mpi as an external package that must be available on the target system

repo/quicksilver/package.py
```diff
@@ -0,0 +1,64 @@
...
+    homepage = "https://codesign.llnl.gov/quicksilver.php"
+    url = "https://github.com/LLNL/Quicksilver/tarball/V1.0"
+    git = "https://github.com/LLNL/Quicksilver.git"
+
+    maintainers("richards12")
+
+    version("master", branch="master")
+    version("1.0", sha256="83371603b169ec75e41fb358881b7bd498e83597cd251ff9e5c35769ef22c59a")
+
+    variant("openmp", default=True, description="Build with OpenMP support")
```
rfhaque: @august-knox Set the default openmp variant to be False. We will need to support other modes like (+cuda, +rocm etc)

experiments/remhos/openmp/ramble.yaml
```diff
@@ -0,0 +1,51 @@
...
+             execute_{n_nodes}:
+              variants: 
+                package_manager: spack  
+              variables:
+                n_nodes: ['1','2']
+  software:
+    packages:
+      hypre:
+        pkg_spec: hypre
+        compiler: compiler-gcc
```
rfhaque: @august-knox Does the hypre build specifically depend on the gcc compiler for this code?

experiments/ior/openmp/ramble.yaml
```diff
@@ -0,0 +1,56 @@
...
+              variants:
+                package_manager: spack 
+              variables:
+                n_nodes: ['1','2','4']
+                b: '268435456/{n_nodes}'
+                t: '{b}/256'
+                #t: '{b}/({sys_cores_per_node}*4)'
+  software:
+    packages:
+      mpi:
```
rfhaque: @august-knox @pearce8 The mpi package spec needs to be removed. We treat mpi as an external package that must be available on the target system

rfhaque: @august-knox This ior implementation should be mpi-only instead of openmp

pearce8:
> Question about Branson input file:
> 
> * In order to change the load, you must change the number of photons, which is a value specified within the input file. Would you know of any ways to potentially edit the file for each individual run?  (if there isn't a natural way to do this, we might ask Branson developers for a parameterized solution).

It sounds like Branson developers use sed.  Can you try something like this:

```
executable('setup_experiment',
           template=[
               'cp {input_path}/* {experiment_run_dir}/.',
               'sed -i "s|<pattern to replace>|<new_pattern>|g" {experiment_run_dir}/<file_name>'
           ])
```

here’s an example:
https://github.com/GoogleCloudPlatform/ramble/blob/develop/var/ramble/repos/builtin/applications/lammps/application.py#L86-L93
alternative option (I am not sure what is better):
There is a filter_file method that Ramble and Spack share that is basically a python implementation of sed (but with much better UI). If there’s not a better way to do it, we can parametrize the file in the application.py file and use the filter_file method to apply the parameters.  I’d probably inject a new phase into the application definition that calls filter_file after make_experiments and handle generating the parameterize inputs there rather than as part of the experiment.  Something like this:

```
register_phase('parameterize_inputs', pipeline='setup', run_after=['make_experiments'])
def _parameterize_inputs(self, workspace, app_inst=None):
  # Copy input files from the inputs directory to '{experiment_run_dir}'
  filter_file( .... )
```

rfhaque:
> Attempting to capture the status on Ruby, Lassen, Tioga. OpenMP versions for 4 benchmarks:
> 
>     * **Quicksilver**: working on Ruby, Lassen, Tioga.  Added strong and weak scaling.
Spun off as #364 
>     * **IOR**: working on Ruby, Lassen, Tioga.  Writing strong and weak scaling.
Spun off as #365 
>     * **Branson**: working on Ruby and Tioga.  MPI-related error on Lassen?  @rfhaque, please take a look
Getting the benchmark to work on lassen requires changes to its build script. We can handle that later when moving to `experiment.py`
>     * **Remhos**: working on Ruby and Lassen.  Hypre compiler error on Tioga?  @rfhaque, please take a look
```
diff --git a/configs/LLNL-Tioga-HPECray-zen3-MI250X-Slingshot/auxiliary_software_files/compilers.yaml b/configs/LLNL-Tioga-HPECray-zen3-MI250X-Slingshot/auxiliary_software_files/compilers.yaml
index 173623d..0d02875 100644
--- a/configs/LLNL-Tioga-HPECray-zen3-MI250X-Slingshot/auxiliary_software_files/compilers.yaml
+++ b/configs/LLNL-Tioga-HPECray-zen3-MI250X-Slingshot/auxiliary_software_files/compilers.yaml
@@ -15,6 +15,7 @@ compilers:
       cflags: -g -O2
       cxxflags: -g -O2 -std=c++17
       fflags: -g -O2 -hnopattern
+      ldflags: -ldl
     operating_system: rhel8
     target: x86_64
     modules: []
diff --git a/experiments/remhos/openmp/ramble.yaml b/experiments/remhos/openmp/ramble.yaml
index 895fe32..fa54de4 100644
--- a/experiments/remhos/openmp/ramble.yaml
+++ b/experiments/remhos/openmp/ramble.yaml
@@ -31,7 +31,7 @@ ramble:
   software:
     packages:
       hypre:
-        pkg_spec: hypre
+        pkg_spec: hypre@2.31.0 +mpi+openmp+mixedint~fortran{modifier_spack_variant}
         compiler: default-compiler
       remhos:
         pkg_spec: remhos@1.0 +metis{modifier_spack_variant} 
```
This patch fixes the tioga issue

rfhaque (approved):

PR #339, opened 2024-08-22
`benchpark experiment init`

### LINKED ISSUE #331: Experiment init
## Problem/Opportunity Statement

With the completion of #298 experiments may now be generated from python objects similar to systems. However, there isn't an interface attached to this functionality which obscures the functionality from users.

## What would success / a fix look like?

To mirror the functionality of `benchpark system init` we should implement a similar CLI for experiments, allowing them to be generated from the command line. For example,

```
benchpark experiment init --output-dir=</path/to/output> <experiment-spec>
```

Closed by #339.

alecbcs:
Adding in an initial `experiment init` subcommand to render experiments.  Closes #331.

### DIFF
--- a/lib/benchpark/cmd/experiment.py
+++ b/lib/benchpark/cmd/experiment.py
@@ -66,4 +66,6 @@ def command(args):
     if args.experiment_subcommand in actions:
         actions[args.experiment_subcommand](args)
     else:
-        raise ValueError(f"Unknown subcommand for 'system': {args.system_subcommand}")
+        raise ValueError(
+            f"Unknown subcommand for 'experiment': {args.experiment_subcommand}"
+        )
--- a/lib/benchpark/cmd/experiment.py
+++ b/lib/benchpark/cmd/experiment.py
@@ -37,7 +37,11 @@ def experiment_init(args):
 
 
 def experiment_list(args):
-    raise NotImplementedError("'benchpark experiment list' is not available")
+    experiments = benchpark.repo.all_object_names(
+        benchpark.repo.ObjectTypes.experiments
+    )
+    # TODO: prettier printing
+    print("    ".join(experiments))
 
 
 def setup_parser(root_parser):
--- a/lib/benchpark/cmd/experiment.py
+++ b/lib/benchpark/cmd/experiment.py
@@ -1,0 +1,68 @@
+# Copyright 2023 Lawrence Livermore National Security, LLC and other
+# Benchpark Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import os
+import shutil
+import sys
+
+import benchpark.experiment
+import benchpark.spec
+
+
+def experiment_init(args):
+    experiment_spec = benchpark.spec.ExperimentSpec(" ".join(args.spec))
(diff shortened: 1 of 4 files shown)


lib/benchpark/cmd/experiment.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+def setup_parser(root_parser):
+    system_subparser = root_parser.add_subparsers(dest="experiment_subcommand")
+
+    init_parser = system_subparser.add_parser("init")
+    init_parser.add_argument("--dest", help="Place all system files here directly")
+    init_parser.add_argument(
+        "--basedir", help="Generate a system dir under this, and place all files there"
+    )
```
becker33, line 54: Why are these two different arguments?

lib/benchpark/cmd/experiment.py
```diff
@@ -0,0 +1,68 @@
...
+        sysdir = experiment.experiment_id()
+        destdir = os.path.join(base, sysdir)
+    elif args.dest:
+        destdir = args.dest
+    else:
+        raise ValueError("Must specify one of: --dest, --basedir")
+
+    try:
+        os.mkdir(destdir)
+        experiment.generate_description(destdir)
```
becker33:
```suggestion
        experiment.write_ramble_dict(destdir)
```

lib/benchpark/cmd/experiment.py
```diff
@@ -0,0 +1,68 @@
...
+        print(f"Abort: experiment description dir already exists ({destdir})")
+        sys.exit(1)
+    except Exception:
+        # If there was a failure, remove any partially-generated resources
+        shutil.rmtree(destdir)
+        raise
+
+
+def experiment_list(args):
+    raise NotImplementedError("'benchpark experiment list' is not available")
```
becker33:
```suggestion
    experiments = benchpark.repo.all_object_names(object_type="experiments").
    print(experiments)
```
We'll probably want prettier printing at some point, but that's the gist of it.

lib/benchpark/cmd/experiment.py
```diff
@@ -0,0 +1,68 @@
...
+
+def command(args):
+    actions = {
+        "init": experiment_init,
+        "list": experiment_list,
+    }
+    if args.experiment_subcommand in actions:
+        actions[args.experiment_subcommand](args)
+    else:
+        raise ValueError(f"Unknown subcommand for 'system': {args.system_subcommand}")
```
becker33:
```suggestion
        raise ValueError(f"Unknown subcommand for 'experiment': {args.experiment_subcommand}")
```

lib/main.py
```diff
@@ -282,7 +283,12 @@ def init_commands(subparsers, actions_dict):
     """
     system_parser = subparsers.add_parser("system", help="Initialize a system config")
     benchpark.cmd.system.setup_parser(system_parser)
+
+    experiment_parser = subparsers.add_parser("experiment", help="Interact with experiments")
```
becker33, line 287: Not for this PR, but we need to move the rest of the commands to `cmd/` and then turn this into a for-loop

becker33 (changes requested):

becker33 (approved):

PR #342, opened 2024-08-22
qws: rewrite as experiment object

alecbcs:
Rewriting `qws` as an experiment object.


pearce8 (changes requested):
Doesn't pass the dry run

PR #344, opened 2024-08-22
experiment class for kripke

becker33:
Implements all functionality from existing kripke experiment in new interface


pearce8:
@becker33 I would like to be able to add the following functionality:

- [ ] Require a definition of a single node experiment (for CI)
- [ ] For weak and strong scaling, I'd like to specify:
 ./bin/benchpark experiment init --dest=kripke-rocm kripke programming_model=rocm scaling=weak start_proc=4 end_proc=32     

I added in the Kripke experiment what I think this could look like.  I think for 3d problems, start_proc may need to be 3d, and it is possible it is easier to say how many times to double the problem/resources.  We will also want to be able to pass in the starting problem size (also 3-dimensional), that way if we need one problem size for one CPU or GPU, and another for another one, we can do this without modifying the experiment file (I would like for the parameters to go into the experiment suite).  It is possible we need to have a 1-dimensional experiment (linear), and a 3-dimensional (3D) experiment to accommodate the difference in parameters.  For 3D, this would be start_proc_x, start_proc_y, start_proc_z, start_size_x, start_size_y, start_size_z, times_to_double_experiment.

Other things I'd like to enforce using base class(es):

- [ ] Problem size and scaling parameters need to be specified orthogonally to the cuda/rocm/programming model, so maybe in a separate function/section?
- [ ] I wonder if we can do something about the nested return section - this is where the experiment writer currently has to play the "guess the pattern" game currently (and I am not sure we can give them tests to check if they got it right).

becker33:
@pearce8 do you want those as part of this PR, or should we merge this as-is and do the scaling as part of #375 ?

pearce8:
@becker33  Just agreed with @rfhaque to merge this, #352 (AMG), and #379 first - he will work on #375 from there.

pearce8 (approved):

PR #283, opened 2024-06-13
[Draft] ScaleML-UNet3D benchmark

pearce8:
## Description

Current priorities:
- [ ] package.py - test
- [ ] application.py
- [ ] cuda experiment - a single experiment
- [ ] test the 3 above, add output here

In a future PR:
- [ ] rocm experiment
- [ ] expand experiments to a set
- [ ] enable Caliper

Adding a note here for later (and possibly for a later PR).  It will soon be possible to use both spack and pip to install dependencies, so lets leave the input generation library to deal with then as we shouldn't need a spack package for it.
Spack package manager: https://github.com/GoogleCloudPlatform/ramble/blob/develop/var/ramble/repos/builtin/package_managers/spack/package_manager.py
Pip package manager: https://github.com/GoogleCloudPlatform/ramble/blob/develop/var/ramble/repos/builtin/package_managers/pip/package_manager.py

Package definitions can have multiple specs that are specific to each package manager. As an example::

packages:
  saxpy:
    spack_pkg_spec: saxpy+openmp
    pip_pkg_spec: "saxpy[openmp]"

Package managers can be parameterized: variants:package_manager:'{experiment_package_manager}'

@Thionazin We have merged the ability to use different package managers, its in benchpark develop now.  First, you need to look at other packages in benchpark to update to use this capability.  Second, here's an example of how to use pip to manage some of your packages:
https://github.com/GoogleCloudPlatform/ramble/blob/develop/var/ramble/repos/builtin/applications/pip-test/application.py
and you would use it from experiments like this::
```
ramble:
  variants:
    package_manager: pip
  applications:
    pip-test:
      workloads:
        import:
          experiments:
            test:
              variables:
                n_ranks: '1'
                mpi_command: ''
                batch_submit: '{execute_experiment}'
  software:
    packages: {}
    environments: {}
```

In order for this to work, you will need to use Ramble concretize:
https://ramble.readthedocs.io/en/latest/workspace.html#concretizing-a-workspace

Dependencies: FIXME:Add a list of any dependencies.

Fixes issue(s): FIXME:Add list of relevant issues.

## Type of Change

- { } Adding a system, benchmark, or experiment
- { } Modifying an existing system, benchmark, or experiment
- { } Documentation update
- { } Build/CI update
- { } Benchpark core functionality

## Checklist:

If adding/modifying a system:
- { } Create a new directory for the system and a new `system.py` file
- { } Add a new dry run unit test in `.github/workflows`
[...]

PR #345, opened 2024-08-23
rename section to match new ramble interface

becker33:
Ramble interface changed `spack` to `software` in one spot

Question: do we update our internals to match, or only update our outputs.

This PR only does the outputs, but we could update the internals as well


pearce8:
Re: do we update our internals to match.  We will need to be able to use pip for some of the packages (incomplete example: #283).  Will that force us to update our internals? @becker33

pearce8 (approved):

becker33:
No -- currently we have a method called `compute_spack_section`, and what it computes is put into a section named `software`. We can rename the method to `compute_software_section` but I just don't like the new ramble name, so I didn't immediately jump on doing that.

pearce8:
I am fine with leaving it as `compute_spack_section`.

PR #349, opened 2024-08-23
Cts system.py (ruby and magma)

dyokelson:
We can add pascal/dane to this same config in a separate PR if we decide to go that direction probably.


lib/benchpark/system.py
```diff
@@ -150,6 +150,7 @@ def variables_yaml(self):
                 raise ValueError(f"Missing required info: {attr}")
 
         optionals = list()
+        optionals_as_cfg = ""
```
scheibelp, line 153: (no change needed) Thanks for fixing that!

scheibelp (approved):

PR #350, opened 2024-08-23
osu-microbenchmarks: rewrite as experiment.py file

august-knox:
Adding osu-microbenchmarks experiment.py file.


becker33 (commented):
What is the purpose of the `programming_model` variant for this experiment?

I think the scaling nodes could also be turned into a variant, maybe @nhanford can chime in on whether that's worth doing.

Otherwise this looks good to me.

pearce8:
Also need to add a dry-run.

pearce8 (changes requested):
Need to add a dry-run.

pearce8 (approved):

pearce8 (changes requested):
It looks like indentation errors in the dry run test prevent the worflow pipeline from executing.

PR #351, opened 2024-08-24
New System class: Sierra

scheibelp:
This adds a new system definition you can select with `benchpark system init` to generate a refined version of https://github.com/LLNL/benchpark/tree/develop/configs/LLNL-Sierra-IBM-power9-V100-Infiniband, e.g.

`benchpark system init --dest=sierra-system sierra cuda=10-1-243 compiler=xl`

The system has two variants:

* one to select CUDA version for components (either 10.1.243 or 11.8.0 for now)
* one to select compiler (4 different options)

The resulting config will contain a single CUDA version, and a compatible compiler selection.

Not all combinations are supported.


pearce8 (changes requested):
Please add a dry run.

pearce8 (approved):

pearce8 (approved):

PR #356, opened 2024-08-27
Feature/new modifier implementation

rfhaque:
New modifier implementation using experiment.py


var/exp_repo/experiments/caliper/experiment.py
```diff
@@ -0,0 +1,94 @@
...
+        modifier_list = []
+        if not self.spec.satisfies("caliper=none"):
+            for var in list(self.spec.variants["caliper"]):
+                if var != "time":
+                    caliper_modifier_modes = {}
+                    caliper_modifier_modes["name"] = "caliper"
+                    caliper_modifier_modes["mode"] = var
+                    modifier_list.append(caliper_modifier_modes)
+            # Add time as the last mode
+            modifier_list.append({"name": "caliper", "mode": "time"})
```
scheibelp, line 41: (question) When you mentioned having to order things yourself, is this what you mean?

In a separate conversation, you mention that this is so "In this case, the time mode is always added last to the modifier list which adds the CALI_CONFIG  variable after each of the CALI_CONFIG_MODE variables have been defined. "

I'm not clear on how this achieves that. Is `time` the only modifier mode that defines `CALI_CONFIG`?

var/exp_repo/experiments/caliper/experiment.py
```diff
@@ -0,0 +1,94 @@
...
+                papi_support = True  # check if target system supports papi
+                if papi_support:
+                    package_specs["caliper"]["pkg_spec"] += "+papi"
+                else:
+                    raise NotImplementedError(
+                        "Target system does not support the papi interface"
+                    )
+            elif self.spec.satisfies("caliper=cuda"):
+                cuda_support = (
+                    self.spec.satisfies("caliper=cuda") and True
```
scheibelp, line 75: `and True` looks vestigial

var/exp_repo/experiments/caliper/experiment.py
```diff
@@ -0,0 +1,94 @@
...
+        system_specs["compiler"] = "default-compiler"
+        if self.spec.satisfies("programming_model=cuda"):
+            system_specs["cuda_arch"] = "{cuda_arch}"
+        if self.spec.satisfies("programming_model=rocm"):
+            system_specs["rocm_arch"] = "{rocm_arch}"
+
+        # set package spack specs
+        package_specs = {}
+
+        if not self.spec.satisfies("caliper=none"):
```
scheibelp, line 60: I get the impression that some combinations of `caliper=...` are not allowed. I don't know if we support `conflict` yet, but FWIW you can raise an exception here for forbidden combinations.

var/exp_repo/experiments/caliper/experiment.py
```diff
@@ -0,0 +1,94 @@
...
+# Benchpark Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+
+from benchpark.directives import variant
+from benchpark.experiment import Experiment
+
+
+class Caliper(Experiment):
```
scheibelp: (suggestion, not a problem) IIRC you mentioned that `caliper` would never be instantiated as an experiment itself. FWIW, this could move from `var/exp_repo/experiments/caliper/` to `lib/benchpark`, possibly into a new subdir like `lib/benchpark/experiment_addons`

var/exp_repo/experiments/amg2023/experiment.py
```diff
@@ -0,0 +1,226 @@
...
+                                "exclude": excludes,
+                                m_tag: matrices,
+                            }
+                        }
+                    }
+                }
+            }
+        }
+
+    def compute_modifiers_section(self):
```
scheibelp, line 105: I have a suggestion for refactoring this to make it easier to add analogs to `caliper`:

- `Experiment.__init__` (in the base class `Experiment`) calls `Experiment.initialize`
- Move all functionality in `caliper.Caliper` except for variant declaration into `caliper.Helper`
- `Amg2023.initialize` sets `helpers = [caliper.Helper()]`
- `Experiment` (the base class) handles tacking on extra modifiers from all `.helpers`
  - It can also take care of `caliper_package_specs`

The benefit of this organization is that you can add more functionality (that defines other modifiers/specs) without having to modify functions in this class other than `.initialize`.

scheibelp (commented):
I have a suggestion on refactoring `caliper.Caliper` into two classes to make it easier to assemble modifier/spec components.

var/exp_repo/experiments/caliper/experiment.py
scheibelp: I think the failure at https://github.com/LLNL/benchpark/actions/runs/11446987606/job/31847369138?pr=356 will be resolved if you make it so that `caliper` does not inherit `Experiment` (which I think is what you wanted to do). If that doesn't work I'll likely have to investigate it (but that failure in the dry runs does in fact look legitimate, so we'll want to address it before merging this).

var/exp_repo/experiments/amg2023/experiment.py
scheibelp: Note: this is implemented in https://github.com/LLNL/benchpark/pull/404

var/exp_repo/experiments/caliper/experiment.py
scheibelp: This could also be addressed by undoing https://github.com/LLNL/benchpark/pull/356/commits/2a0a077d6f85ee4861e0a7bf07290a96ca3bf138 I think (the part that rearranges the order of declared superclasses)

scheibelp (approved):

PR #362, opened 2024-09-06
experiment class for laghos

august-knox:
bin/benchpark experiment init --dest laghos

Don't believe there's any variants that need to be specified here


repo/laghos/application.py
```diff
@@ -19,10 +19,33 @@ class Laghos(ExecutableApplication):
...
+        
+    workload_variable('pt', default='',
+            description='partitioning ratio',
+            workloads=['laghos'])
+
+    workload_variable('rs', default='',
+            description='number of serial refinements',
+            workloads=['laghos'])
+    
+    workload_variable('rp', default='0',
```
dyokelson, line 38: Let's keep this "example" default at what it was before so default value should be 5, since we can pass in different values now via the scaling variants in experiment.

repo/laghos/application.py
```diff
@@ -19,10 +19,33 @@ class Laghos(ExecutableApplication):
...
+            description='number of parallel refinements',
+            workloads=['laghos'])
+    workload_variable('ms', default='4',
```
dyokelson: Likewise let's update this default value back to 500

var/exp_repo/experiments/laghos/experiment.py
```diff
@@ -0,0 +1,62 @@
...
+        "experiment",
+        default="example",
+        values=("weak", "strong","example"),
+        description="weak or strong scaling",
+    )
+
+    def compute_applications_section(self):
+        variables = {}
+            
+        if self.spec.satisfies("scaling=weak"):
```
dyokelson: Let's start with the scaling assuming a single node, but change problem size per rank instead, once we know we have this right we can add more nodes or make updates. So for weak scaling:
rs values: 1
num processes: 8
rp values (increases workload per rank): 1,2,3,4

Another option if you can also implement this is rs values 1,2,3 and mpi processes 1,8,64

var/exp_repo/experiments/laghos/experiment.py
```diff
@@ -0,0 +1,62 @@
...
+            variables["rs"] = ["5","6"]
+            variables["n_nodes"] = ["1","8"]
+        elif self.spec.satisfies("scaling=strong"):
+            variables["n_nodes"] = ["1","8"]
```
dyokelson: Instead of increasing nodes for strong scaling let's keep it on one node and increase MPI ranks, 1,8,64. We should lower the rs value to 1 or 2 here as well, depending on how long this takes to solve, these would ideally be less than 5 minutes to solve.

var/exp_repo/experiments/laghos/experiment.py
```diff
@@ -0,0 +1,62 @@
...
+            variables["rs"] = "5"
+        else:
+            variables["n_nodes"] = ["1","2","4","8","16","32","64","128"]    
+        variables["n_ranks"] = "{sys_cores_per_node} * {n_nodes}"
```
dyokelson: Since we're setting ranks explicitly above let's move both n_nodes and n_ranks into the else, and set this to one node, 8 ranks since this will be for the example run.

var/exp_repo/experiments/laghos/experiment.py
```diff
@@ -0,0 +1,62 @@
...
+
+        return {
+            "laghos": {  # ramble Application name
+                "workloads":{
+                    # TODO replace with a hash once we have one?
+                    "problem": {
+                       #"variables": variables,
+                        "experiments": {
+                            "laghos_{n_nodes}_{n_ranks}": {
```
dyokelson: Can you add the other parameters into the naming scheme here, so basically anything that gets passed in (rs, rp, etc) should be in here, see the naming for the kripke experiment in https://github.com/LLNL/benchpark/pull/344/files

dyokelson (changes requested):
Nice job so far! Some updates requested, let me know if you have any questions.

var/exp_repo/experiments/laghos/experiment.py
pearce8: @august-knox I enabled a dry-run test and it is failing.  Please diagnose/fix.

var/exp_repo/experiments/laghos/experiment.py
```diff
@@ -0,0 +1,74 @@
+from benchpark.directives import variant
```
rfhaque: @pearce8 @august-knox We need to remove this file. I just kept it around for comparison

.github/workflows/run.yml
```diff
@@ -219,22 +219,23 @@ jobs:
             --disable-logger \
             workspace setup --dry-run
 
-      - name: Dry run laghos/mpi-only on LLNL-Magma-Penguin-icelake-OmniPath with allocation modifier
+      - name: Dry run dynamic laghos/mpi-only on LLNL-Magma-Penguin-icelake-OmniPath with allocation modifier
         run: |
-          ./bin/benchpark setup laghos/mpi-only LLNL-Magma-Penguin-icelake-OmniPath workspace/
+            ./bin/benchpark experiment init --dest=laghos-mpi-only laghos
```
slabasan: @august-knox Can you fix this indentation? It's causing yaml syntax errors.

pearce8:
@rfhaque this runs into the zlib problem.

pearce8 (changes requested):
Dry run fails, looks like a zlib issue.

PR #375, opened 2024-09-25
Feature/experiment scaling

rfhaque:
Draft implementation of the scaling functionality for benchmarks. Currently only strong scaling is implemented. Works for amg2023/openmp 
`bin/benchpark experiment init --dest ./experiments kripke programming_model=openmp scaling=strong strong-scaling-factor=2 strong-scaling-num-exprs=3`

and kripke/openmp
`bin/benchpark experiment init --dest ./experiments amg2023 programming_model=openmp scaling=strong strong-scaling-factor=2 strong-scaling-num-exprs=4`


var/exp_repo/experiments/scaling/experiment.py
```diff
@@ -0,0 +1,65 @@
...
+        "weak-scaling-factor",
+        default="2",
+        description="Weak-scaling factor (factor by which to increase resources and problem sizes)",
+    )
+
+    variant(
+        "weak-scaling-num-exprs",
+        default="4",
+        description="Number of weak-scaling experiments",
+    )
```
becker33: We need to implement conditional variants so we can make this clearer. This could be

```
variant("scaling", default="none", values=("none", "strong", "weak"), description="...")

variant("scaling-factor", default=2, values=lambda x: x >= 2, when="scaling=weak", description="...")
variant("scaling-factor", default=2, values=lambda x: x >= 2, when="scaling=strong", description="...")

variant("scaling-iterations", default=4, values=lambda x: x > 1, when="scaling=weak", description="...")
variant("scaling-iterations", default=4, values=lambda x: x > 1, when="scaling=strong", description="...")
```
And with a little additional syntax beyond that we could get rid of the duplicate lines.

var/exp_repo/experiments/scaling/experiment.py
```diff
@@ -0,0 +1,65 @@
+from benchpark.directives import variant
+
+
+class Scaling(object):
```
becker33: I would call this `ScalingExperiment` instead of just `Scaling`, I think it's a bit clearer that way

var/exp_repo/experiments/scaling/experiment.py
```diff
@@ -0,0 +1,65 @@
...
+        resource_list = [[x] for x in initial_resource_list]
+
+        while num_exprs > 0:
+            for idx in round_robin_order:
+                for i, r in enumerate(resource_list):
+                    r.append(r[-1]*scaling_factor if i == idx else r[-1])
+                num_exprs=num_exprs-1
+                if not num_exprs:
+                    break
+        return resource_list
```
becker33: I think this method can be simplified slightly

```suggestion
    def generate_strong_scaling_parameters(self, initial_resource_list: list):
        scaling_factor = int(self.spec.variants["strong-scaling-factor"][0])
        num_exprs = int(self.spec.variants["strong-scaling-num-exprs"][0])
        round_robin_order = self.compute_round_robin_order(initial_resource_list)
        resource_list = [[x] for x in initial_resource_list]

        for i in range(num_exprs - 1):
            idx = (i + round_robin_order[0]) * len(initial_resource_list)
            for r_idx, resource in enumerate(resource_list):
                next_value = resource[-1] * scaling_value if r_idx == idx else resource[-1]
                resource.append(next_value)

        return resource_list
```

var/exp_repo/experiments/scaling/experiment.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def generate_weak_scaling_parameters(self, initial_resource_list: list, initial_problem_size_list: list):
```
becker33: I don't think it's necessarily safe for us to assume that the resource requirements and the problem size are expressed in such neatly compatible ways. I think we should probably assume these could be different dimensions (e.g. resource list is a scalar, problem size is 3D) and round-robin them each separately.

becker33:
The variant logic can be simplified once we merge #379

var/exp_repo/experiments/scalingexperiment/experiment.py
```diff
@@ -0,0 +1,124 @@
+from benchpark.directives import variant
+
+
+class ScalingExperiment(object):
```
scheibelp: (no change needed) I originally thought it would be problematic that this defines `variant()`s but does not inherit `Experiment`, but I think that as long as any package that inherits this also inherits `Experiment`, there will be no issue with this.

That being said, IIRC one of your arguments about not inheriting `Experiment` was a concern with overriding methods in `Experiment`, which is technically still possible regardless of whether `ScalingExperiment` inherits `Experiment`.

var/exp_repo/experiments/scalingexperiment/experiment.py
```diff
@@ -0,0 +1,124 @@
...
+
+    variant(
+        "scaling-iterations",
+        default="4",
+        values=int,
+        description="Number of experiments to be generated",
+    )
+
+    # input parameters:
+    # 1. input_variables: dict[str, int | tuple(str), list[int]]. Input variables
```
scheibelp: If this is another way to express:

> input_variables key value pairs are either of type str: int or tuple(str): list(int)

I can say this mypy-like syntax of expression is confusing (e.g. if `list(int)` values *always* have `tuple(str)` keys and never `str` keys, this isn't how it can be expressed in `mypy`). I recommend replacing what you have on this line with the quoted explanation that you used elsewhere.

If that's not the intended meaning, can you explain how this input differs?

var/exp_repo/experiments/scalingexperiment/experiment.py
```diff
@@ -0,0 +1,124 @@
...
+    # For the first value in input_variables, if the value is a list, select the
+    # index of its smallest element, 0 otherwise
+    # Beginning with this index, generate a list of indexes of length equal to
+    # the number of dimensions in an (ascending) round-robin order
+    #
+    # output:
+    # scaling_order: list[int]. list of num_exprs values, one for each dimension,
```
scheibelp: `num_exprs` is not an input to this particular function, I think this is a list of indices.

var/exp_repo/experiments/scalingexperiment/experiment.py
```diff
@@ -0,0 +1,124 @@
...
+    #
+    # 2. scaling_factor: int. Factor by which to scale the variables. All entries in
+    # input_variables are scaled by the same factor
+    #
+    # 3. num_exprs: int. Number of experiments to be generated
+    #
+    # output:
+    # output_variables: dict[str, int | list[int]]. num_exprs values for each
+    # dimension of the input variable scaled by the scaling_factor according to the
+    # scaling policy
```
scheibelp: I think example inputs and outputs would be useful in the comments/doc for this function, which you could adapt from https://github.com/LLNL/benchpark/blob/develop/experiments/amg2023/openmp/ramble.yaml, for example:

```
input:
  [[10], [10], [10]]
  [[2], [2], [2]]
output:
  nx: ['10', '20', '20']
  ny: ['10', '10', '20']
  nz: ['10', '10', '10']
  px: ['2', '4', '4']
  py: ['2', '2', '4']
  pz: ['2', '2', '2']
```

var/exp_repo/experiments/scalingexperiment/experiment.py
```diff
@@ -0,0 +1,124 @@
...
+                if len(param) == 1:
+                    param[0].append(param[0][-1] * scaling_factor)
+                else:
+                    for p_idx, p_val in enumerate(param):
+                        p_val.append(
+                            p_val[-1] * scaling_factor
+                            if p_idx
+                            == scaling_order_index[exp_num % len(scaling_order_index)]
+                            else p_val[-1]
+                        )
```
scheibelp: IMO a comment like:

> Take initial parameterized vector for experiment, for each experiment after the first, scale one dimension of that vector by the scaling factor; cycle through the dimensions in round-robin fashion.

would be useful here

var/exp_repo/experiments/amg2023/experiment.py
```diff
@@ -18,106 +19,143 @@ class Amg2023(Experiment):
...
+        else:
+            self.workload = "problem2"
+
+        px = "px"
+        py = "py"
+        pz = "pz"
+        nx = "nx"
+        ny = "ny"
+        nz = "nz"
+        num_procs = f"{{{px}}} * {{{py}}} * {{{pz}}}"
```
scheibelp: (minor) I'm assuming you are using `px` as an abbreviation for `"px"`, and not because it's value might change: in that case`"{px} * {py} * {pz}"` is simpler

var/exp_repo/experiments/scalingexperiment/experiment.py
```diff
@@ -0,0 +1,124 @@
...
+            if isinstance(param, list):
+                n_dims = len(param)
+                break
+
+        # starting with the minimum value dim of the ordering parameter
+        # compute the remaining n_dims-1 values in a round-robin manner
+        val = input_variables[next(iter(input_variables))]
+        min_dim = val.index(min(val)) if isinstance(val, list) else 0
+
+        return [(min_dim + i) % n_dims for i in range(n_dims)]
```
scheibelp: (suggestion) Since this doesn't sort (e.g. doesn't make sure that the second-lowest dimension is handled second), IMO this would be simpler if it just returned `min_dim`.

var/exp_repo/experiments/scalingexperiment/experiment.py
```diff
@@ -0,0 +1,124 @@
...
+        values=int,
+        description="Factor by which to scale values of problem variables",
+    )
+
+    variant(
+        "scaling-iterations",
+        default="4",
+        values=int,
+        description="Number of experiments to be generated",
+    )
```
scheibelp: #379 was merged, but these are not conditioned on `when=` (as mentioned in https://github.com/LLNL/benchpark/pull/375#discussion_r1775946142). I think you can effectively turn scaling off by setting `scaling-iterations=1`, although it still might be clearer to use `when`.

scheibelp (commented):
I don't see any major issues, but have a few suggestions

lib/benchpark/experiment.py
```diff
@@ -48,10 +49,141 @@ class Experiment(ExperimentSystemBase):
...
+        description="Factor by which to scale values of problem variables",
+    )
+
+    variant(
+        "scaling-iterations",
+        default="4",
+        values=int,
+        description="Number of experiments to be generated",
+    )
+
```
rfhaque, line 65: @scheibelp I have incorporated your earlier suggestions into the scaling methods. I also moved the entire scaling implementation directly into the base Experiment class. Please review and comment this is a good approach. Especially the fact that the base Experiment class will have two variants `scaling-factor` and `scaling-iterations` defined in it

lib/benchpark/experiment.py
```diff
@@ -48,10 +49,153 @@ class Experiment(ExperimentSystemBase):
...
+    # specified, the scaling order is defined using the first element in input_variables
+    #
+    # output:
+    # scaling_order: list[int]. list of indices, with one value for each dimension,
+    # output:
+    # output_variables: dict[str, int | list[int]]. num_exprs values for each
+    # dimension of the input variable scaled by the scaling_factor according to the
+    # scaling policy
+    def scale_experiment_variables(
+        self, input_variables, scaling_factor, num_exprs, scaling_variable=None
```
rfhaque, line 121: Hi @dyokelson, I've added the option to specify a scaling_variable to the experiment (all variables in `input_variables` will be scaled according to the ordering defined on `scaling_variable`

var/exp_repo/experiments/amg2023/experiment.py
```diff
@@ -18,106 +18,147 @@ class Amg2023(Experiment):
...
+                variables[nz] = initial_n[2]
+            elif self.spec.satisfies("experiment=weak"):
+                scaling_variable = (px, py, pz)
+                input_params[scaling_variable] = initial_p
+                input_params[(nx, ny, nz)] = initial_n
+            variables |= self.scale_experiment_variables(
+                input_params,
+                int(self.spec.variants["scaling-factor"][0]),
+                int(self.spec.variants["scaling-iterations"][0]),
+                scaling_variable,
```
rfhaque, line 177: @dyokelson Usage of the `scaling_variable`

var/exp_repo/experiments/amg2023/experiment.py
rfhaque: @scheibelp This will be removed once I add variants for all `px`, `py` and `pz`

rfhaque:
@pearce8 @scheibelp If the changes in the PR are acceptable, we can take this to main. I can make further changes in a separate PR

var/exp_repo/experiments/scaling/experiment.py
pearce8: While the resource list is a scalar, there are a few concepts here that need to be purposefully coupled, in D dimensions (where D is 1, 2, or 3):
- Problem size (in D dimensions)
- Decomposition (in D dimensions)

They are coupled to keep the D dimensional size of the problem per resource the same shape.  In 2D, you are aiming for squares, in 3D for cubes, to minimize the surface to volume ratio.

Of course someone could be studying the impact of decomposition on solution time - but in that case they will be trying who knows what decompositions, so I do not think it is tractable to provide all things to all people.

var/exp_repo/experiments/amg2023/experiment.py
pearce8: As discussed on Tuesday, we should not be adding ``px``, ``py``, and ``pz`` as variants.

pearce8 (approved):

pearce8 (changes requested):
I think this is still not working.  We need:
- [x] By default, we should get a single node experiment, and no scaling code should be called nor extra parameters to the experiment specified
- [x] Add a dry run test that tests the scaling functionality

lib/benchpark/experiment.py
```diff
@@ -48,10 +49,153 @@ class Experiment(ExperimentSystemBase):
...
+        values=int,
+        description="Factor by which to scale values of problem variables",
+    )
+
+    variant(
+        "scaling-iterations",
+        default="4",
+        values=int,
+        description="Number of experiments to be generated",
+    )
```
scheibelp, line 64: (question) Were there plans to make these variants conditional? I'm not strictly for or against, but I thought I'd cover the why/why-not:

Without `when=...` these two variants are sometimes irrelevant (e.g. when an experiment subclass sets `scaling=none`). That's not strictly an error, but it can lead to confusing configurations.

OTOH, `when=` would be hard to implement because the variant related to scaling lives in the subclass

* We could probably standardize on some values for `experiment` that would indicate no scaling
* But we can't express `when!=those-values` (in particular when the values outside the agreed-upon set may differ between experiment subclasses)
  * and the experiment values you would want to do scaling for change from experiment to experiment

var/exp_repo/experiments/kripke/experiment.py
```diff
@@ -119,28 +104,64 @@ def compute_applications_section(self):
...
+            system_specs["cuda_version"] = "{default_cuda_version}"
+            system_specs["cuda_arch"] = "{cuda_arch}"
+        if self.spec.satisfies("programming_model=rocm"):
+            system_specs["rocm_arch"] = "{rocm_arch}"
+
+        # set package spack specs
+        package_specs = {}
+        if self.spec.satisfies("programming_model=cuda"):
+            package_specs["cuda"] = {
+                "pkg_spec": "cuda@{}+allow-unsupported-compilers".format(
```
scheibelp, line 129: You should be able to say just `cuda` here: the system will always handle generating the appropriate external cuda spec.

pearce8 (approved):

lib/benchpark/experiment.py
```diff
@@ -48,10 +49,153 @@ class Experiment(ExperimentSystemBase):
         Dict[str, benchpark.variant.Variant],
     ]
 
+    variant(
+        "scaling-factor",
+        default="2",
+        values=int,
+        description="Factor by which to scale values of problem variables",
+    )
```
scheibelp, line 57: I'm wondering if it might be more generally applicable for a variant that exists across all experiments that expresses the maximum system resources you want to test at (e.g. "do strong scaling on up to 100 nodes"). Is it common to specify scaling factors other than 2? IMO "keep doubling until you hit X nodes" would be a bit "safer" than "double X times"

scheibelp (commented):
Leaving some comments post merge for discussion

PR #384, opened 2024-10-02
Add test suite

becker33:
Test suite uses pytest.

Includes a Spack environment that can be activated with direnv to build the required python environment for benchpark

Future work:
- Run test suite in CI
- Use codecov in CI
- Add more tests

I do not think we should hold up this PR for any of those future work items.


pearce8 (dismissed):
Looks good to me.  

- [x] @becker33 please verify that the lint fix I committed is reasonable.  
- [ ] I don't understand why the workspace directory doesn't get created - please pull in @alecbcs or @scheibelp to help if you are not able to get to this before the holidays.

scheibelp:
> I don't understand why the workspace directory doesn't get created 

If you are referring to the failed dry run: the workspace is created, but this PR changes the hashing of specs, which means that the dynamically generated system dir changes (because it is based on the system spec hash).

For example, you can update the dry run to replace `Tioga-d34a754` with`Tioga-975af3c` to get it to pass (the other dry run using a dynamically generated system will also fail, and the same type of fix will work for that).

scheibelp (approved):

PR #400, opened 2024-10-14
[experimental] Adding a CudaExperiment, and variables to Base and CudaExperiment

pearce8:
- [ ] Variables for Application and Spack section to go into Experiment
- [ ] CudaExperiment to hold cuda-specific versions for the above
- [ ] Application-specific experiment.py would then be ``class Amg2023(Experiment,CudaExperiment):``
- [ ] Can/should we define experiment.py for hypre, and use it in experiment.py for amg2023?

### DIFF
--- a/lib/benchpark/cuda.py
+++ b/lib/benchpark/cuda.py
@@ -20,7 +20,7 @@ import ramble.language.language_base  # noqa
 import ramble.language.language_helpers  # noqa
 
 
-class Experiment(ExperimentSystemBase):
+class CudaExperiment(ExperimentSystemBase):
     """This is the superclass for all benchpark experiments.
 
     ***The Experiment class***
@@ -37,6 +37,10 @@ class Experiment(ExperimentSystemBase):
          essentially a collection of files defining an experiment in a
          Ramble workspace.
     """
+
+    variant("cuda", default=False, description="Build and run with CUDA")
+
+
 
     #
     # These are default values for instance variables.
@@ -71,8 +75,7 @@ class Experiment(ExperimentSystemBase):
     def compute_applications_section(self):
         # TODO: is there some reasonable default?
         variables = {}
-        variables["n_ranks"] = num_procs  # TODO: num_procs will be defined in the child...
-        n_resources = num_procs
+        variables["n_gpus"] = num_procs  # TODO: num_procs will be defined in the child...
 
         raise NotImplementedError(
             "Each experiment must implement compute_applications_section"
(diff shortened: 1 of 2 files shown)


lib/benchpark/cuda.py
```diff
@@ -0,0 +1,131 @@
...
+    There are two main parts of a Benchpark experiment:
+
+      1. **The experiment class**.  Classes contain ``directives``, which are
+         special functions, that add metadata (variants) to packages (see
+         ``directives.py``).
+
+      2. **Experiment instances**. Once instantiated, an experiment is
+         essentially a collection of files defining an experiment in a
+         Ramble workspace.
+    """
```
scheibelp, line 26: This docstring is copied from `Experiment` and doesn't explain anything new or specific to this class.

lib/benchpark/cuda.py
```diff
@@ -0,0 +1,131 @@
...
+
+    #
+    # These are default values for instance variables.
+    #
+
+    # This allows analysis tools to correctly interpret the class attributes.
+    variants: Dict[
+        "benchpark.spec.Spec",
+        Dict[str, benchpark.variant.Variant],
+    ]
```
scheibelp, line 37: You should not have to instantiate this here. Why does this inherit from `ExperimentSystemBase` vs. `Experiment`? If this is related to MRO concerns, see my later comment about using composition.

lib/benchpark/cuda.py
```diff
@@ -0,0 +1,131 @@
...
+        raise NotImplementedError(
+            "Each experiment must implement compute_applications_section"
+        )
+
+    def compute_spack_section(self):
+        # TODO: is there some reasonable default based on known variable names?
+        system_specs = {}
+        system_specs["cuda_version"] = "{default_cuda_version}"
+        system_specs["cuda_arch"] = "{cuda_arch}"
+        system_specs["blas"] = "cublas-cuda"  #TODO: can we define this in Lapack/Blas CudaExperiment?
```
scheibelp: In my experience this differs from system to system, and it's best handled by saying here just that you need `blas`; the system can then register whatever `blas` external is appropriate (see for example in Sierra definition: https://github.com/LLNL/benchpark/blob/develop/var/sys_repo/systems/sierra/externals/cuda/00-version-10-1-243-packages.yaml).

lib/benchpark/cuda.py
```diff
@@ -0,0 +1,131 @@
...
+        package_specs["hypre"]["pkg_spec"] += "+cuda cuda_arch={}".format(
+            system_specs["cuda_arch"]
+        )
+        package_specs[app_name]["pkg_spec"] += "+cuda cuda_arch={}".format(
+            system_specs["cuda_arch"]
+        )
+
+        raise NotImplementedError(
+            "Each experiment must implement compute_spack_section"
+        )
```
scheibelp: With an inheritance scheme, this would have to return something, e.g.

```
        return {
            "packages": {k: v for k, v in package_specs.items() if v},
            "environments": {app_name: {"packages": list(package_specs.keys())}},
        }
```

(copied from Amg2023) at which point, a subclass would want to do something like:

```
class FooExperiment(CudaExperiment):
  def compute_spack_section(self):
    config = super().compute_spack_section()
    # now add all the things specific to Foo
```

Using inheritance is somewhat inflexible, because everything you want to mix together would need to be part of a single inheritance chain (which is why I suggested composition in https://github.com/LLNL/benchpark/pull/356#discussion_r1783579262, for example).

lib/benchpark/experiment.py
```diff
@@ -70,12 +70,43 @@ def compute_modifiers_section(self):
 
     def compute_applications_section(self):
         # TODO: is there some reasonable default?
+        variables = {}
+        variables["n_ranks"] = num_procs  # TODO: num_procs will be defined in the child...
```
scheibelp: If that is so, this will need to be `self.num_procs`, and that will be a new required method for all subclasses (which can be checked by `benchpark audit`, FWIW).

lib/benchpark/experiment.py
```diff
@@ -70,12 +70,43 @@ def compute_modifiers_section(self):
...
+
+        # TODO: is there a way to generically do this?
+        package_specs[app_name] = {
+            "pkg_spec": f"{app_name}@{app_version} +mpi",
+            "compiler": system_specs["compiler"],
+        }
+
         raise NotImplementedError(
             "Each experiment must implement compute_spack_section"
         )
```
scheibelp, line 103: If this does anything, it can't raise `NotImplementedError`. FWIW `benchpark audit` currently makes sure that the lowest-level definition (e.g. what lives in `exp_repo`) defines this method.

lib/benchpark/experiment.py
```diff
@@ -70,12 +70,43 @@ def compute_modifiers_section(self):
...
         # TODO: is there some reasonable default based on known variable names?
+        system_specs = {}
+        system_specs["compiler"] = "default-compiler"
+        system_specs["mpi"] = "default-mpi"
+        system_specs["lapack"] = "lapack"   #TODO: can we define this in lapack experiment?
+
+        package_specs = {}
+        package_specs[system_specs["mpi"]] = (
+            {}
+        )  # empty package_specs value implies external package
```
scheibelp, line 93: This seems like a good candidate for a method in `Experiment` like

```
def needs_external(pkgs_dict, system_specs, pkg_name):
  pkgs_dict[system_specs[pkg_name]] = {}
```

(mainly so this comment doesn't have to be repeated).

scheibelp (commented):
Overall my comments fall into two categories

* Some discuss how you could potentially make this inheritance-based approach work
* Some highlight the potential issues with an inheritance-based approach and discuss an alternative

PR #410, opened 2024-10-23
Printing message in bootstrap

pearce8:


lib/benchpark/runtime.py
```diff
@@ -82,7 +82,9 @@ def __init__(self, dest):
         self.spack_location = self.dest / "spack"
 
     def bootstrap(self):
+        print("Hold tight, Benchpark is bootstrapping itself.")
```
scheibelp, line 85: I notice that e.g. `_install_ramble` has this already, but as a debug print statement:

```
    def _install_ramble(self):
        debug_print(f"Cloning Ramble to {self.ramble_location}")
```

IMO it would be preferable to promote that to a `print()` statement (likewise for spack)

scheibelp (approved):
It occurs to me we might want to hide status messages like this behind a verbosity option to `benchpark` - that would be useful any future `benchpark` command were to generate output that someone might want to use as part of there scripts. We don't have any such cases now, so this is an improvement given how long it can take to get Spack.

PR #416, opened 2024-10-29
move experiments/systems

scheibelp:
Doesn't yet update docs

* Move all static system/experiment definitions into `legacy` dir
* Move `var/exp_repo/experiments` to `experiments`
  * Likewise for `var/sys_repo/systems`
* Update code references

The actual code changes are small

PR #419, opened 2024-10-31
Move experiments/systems: Reapply changes from #416

scheibelp:
Closes #416

On top of latest develop


pearce8 (approved):

dyokelson:
Pulled this locally, checked the tables in the docs, benchpark list command, as well as tried out the "legacy" way of setting up a system and experiment and everything seems to work as expected!

PR #435, opened 2024-11-14
System variant: no scheduler

pearce8:
Enable a way to generate run scripts without scheduler instructions, for interactive runs or composing CI runs.  

```
./bin/benchpark system init --dest=tioga-system tioga rocm=551 compiler=cce ~gtl scheduler=mpi
```

- [x] Base class only has the no scheduler instructions option (`mpi`).  
- [ ] The AWS system doesn't need to define a new scheduler; should the scheduler code be removed from it?
- [ ] The AWS system needs a dry run (should be addressed in response to #455)
- [ ] Derived class still specifies the scheduler on the system, which will be used by default (when the scheduler is not specified during `system init`).
- [ ] Is there a way to clean up the code in the system-specific system.py?  ``values=("flux", "mpi")`` with "flux" being the default option seems brittle.
- [ ] Is there a way to make the scheduler a binary variant?  i.e., scheduler=True uses flux, scheduler=False uses mpi.


systems/tioga/system.py
```diff
@@ -33,8 +33,8 @@ class Tioga(System):
 
     def initialize(self):
         super().initialize()
-
-        self.scheduler = "flux"
+        if self.spec.variants["scheduler"][0]=="yes":
```
pearce8: if self.spec.satisfies("schedule=yes"):

becker33: variant("scheduler", values=("flux", "mpi"), default="flux", sticky=True, description="...")

lib/benchpark/system.py
```diff
@@ -82,7 +90,7 @@ def initialize(self):
         self.sys_cores_per_node = None
         self.sys_gpus_per_node = None
         self.sys_mem_per_node = None
-        self.scheduler = None
+        self.scheduler = "mpi"
```
becker33:
```suggestion
        self.scheduler = self.spec.variants["scheduler"].value
```

lib/benchpark/system.py
```diff
@@ -72,6 +73,13 @@ class System(ExperimentSystemBase):
         Tuple["benchpark.variant.Variant", "benchpark.spec.ConcreteSystemSpec"],
     ]
 
+    variant(
+        "scheduler",
+        default="yes",
+        values=("yes", "none"),
+        description="Use the scheduler on the system, or generate scripts without the scheduler commands",
+    )
```
becker33:
```suggestion
    variant(
        "scheduler",
        default="mpi",
        values=("mpi",),
        description="Use the scheduler on the system, or generate scripts without the scheduler commands",
    )
```

pearce8: I reverted the changes from @dyokelson.  @becker33 the dry run now shows the error: the value of the variant is 'mpi' when the concretizer expects 'None'.  @scheibelp if you are able to help fix this, please do.

lib/benchpark/system.py
```diff
@@ -72,6 +73,13 @@ class System(ExperimentSystemBase):
         Tuple["benchpark.variant.Variant", "benchpark.spec.ConcreteSystemSpec"],
     ]
 
+    variant(
+        "scheduler",
+        default="mpi",
+        values=("mpi"),
```
scheibelp: This needs a comma: `values = ("mpi",)`

lib/benchpark/system.py
```diff
@@ -82,7 +90,7 @@ def initialize(self):
         self.sys_cores_per_node = None
         self.sys_gpus_per_node = None
         self.sys_mem_per_node = None
-        self.scheduler = None
+        self.scheduler = self.spec.variants["scheduler"].value
```
scheibelp: Once the comma is added to the definition, this will need to access via index:  `self.spec.variants["scheduler"][0]`

(this is a departure from how variants are referenced in Spack, but we don't seem to use `.value` anywhere in Benchpark)

scheibelp:
Now that Tioga has a new variant, the hash location of the Ramble workspace has changed (sorry - https://github.com/LLNL/benchpark/pull/450 ought to help with this)

pearce8:
As per issue #485, we will do this differently.

PR #440, opened 2024-11-19
variants: improve handling of booleans

becker33:
boolean variants now store values as bool
all other variants still store values as str

### DIFF
--- a/lib/benchpark/openmp.py
+++ b/lib/benchpark/openmp.py
@@ -13,7 +13,7 @@ class OpenMPExperiment:
 
     class Helper(ExperimentHelper):
         def get_helper_name_prefix(self):
-            return "openmp" if self.spec.satisfies("openmp") else ""
+            return "openmp" if self.spec.satisfies("+openmp") else ""
 
         def get_spack_variants(self):
-            return "+openmp" if self.spec.satisfies("openmp") else "~openmp"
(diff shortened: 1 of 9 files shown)


lib/benchpark/openmp.py
```diff
@@ -9,16 +9,11 @@
...
-        default="non",
-        values=("oui", "non"),
-        description="Build and run with OpenMP",
-    )
+    variant("openmp", default=False, description="Build and run with OpenMP")
 
     class Helper(ExperimentHelper):
         def get_helper_name_prefix(self):
-            return "openmp" if self.spec.satisfies("openmp=oui") else ""
+            return "openmp" if self.spec.satisfies("openmp") else ""
```
scheibelp: This would have to be `.satisfies("+openmp")` wouldn't it?

(if this is true, then FYI this correction needs to be made in several files)

pearce8:
I turned OpenMP, CUDA, ROCm variants into binary variants to test @becker33's code.  

@scheibelp please review/merge.

@rfhaque please take a look, and start a new PR to fix up the other variants that can now be binary.

scheibelp (approved):
I also pulled this and checked for any remaining references to `openmp/cuda/rocm=oui` and didn't see any. So this leaves:

* `single_node`
* `strong`
* `weak`
* `throughput`

PR #446, opened 2024-11-24
lammps experiment.py

rfhaque:


repo/lammps/package.py
```diff
@@ -9,10 +9,22 @@
 
 class Lammps(BuiltinLammps):
 
+  depends_on("kokkos+openmp", when="+openmp")
```
pearce8, line 12: Is this not in the upstreamed package?  Did ORNL not need this?

systems/tioga/compilers/rocm/00-rocm-551-compilers.yaml
```diff
@@ -20,8 +20,8 @@ compilers:
     modules: []
     environment:
       prepend_path:
-        LD_LIBRARY_PATH: /opt/cray/pe/cce/16.0.0/cce/x86_64/lib
-    extra_rpaths: [/opt/cray/pe/cce/16.0.0/cce/x86_64/lib/, /opt/cray/pe/gcc-libs/]
+        LD_LIBRARY_PATH: "/opt/cray/pe/cce/16.0.0/cce/x86_64/lib:/opt/rocm-5.5.1/lib"
```
pearce8: So this works on both the static and dynamic tioga?

pearce8 (changes requested):
- [ ] Please add a dry run
- [ ] Does this work on any system other than Tioga at this point?

repo/lammps/package.py
rfhaque: I think there is a standalone openmp mode in lammps where we disable the kokkos pacakge and just enable the openmp one (~kokkkos+openmp-package) With this spack spec, we always enable kokkos with the desired backend (+kossos+openmp)
We can ask ORNL which one of `+kokkos+openmp` or `+openmp-package+openmp` is preferable.

systems/tioga/compilers/rocm/00-rocm-551-compilers.yaml
rfhaque: I will not work on static tioga. I'll need to add the `LD_LIBRARY_PATH`

rfhaque:
> * [ ]  Please add a dry run
Will do that
>     * [ ]  Does this work on any system other than Tioga at this point?
The code works on tioga(openmp, rocm) and ruby(openmp)

pearce8 (approved):

Issue #443, opened 2024-11-22
Add a stand alone function to get the system name with unique hash to benchpark python


pearce8:
Resolved by #450.

PR #450, opened 2024-11-26
`benchpark setup`: easier referencing of dynamically generated systems

scheibelp:
`benchpark setup`, when using a dynamically generated system (from `benchpark system init`), will generate a ramble workspace in part using the name/hash of the system

This can make it awkward to reference, as is demonstrated in our dry run CI checks:

```
./bin/benchpark system init --dest=tioga-system tioga rocm=551 compiler=cce ~gtl
...
./bin/benchpark setup ./saxpy-rocm ./tioga-system workspace/
...
ramble --workspace-dir workspace/saxpy-rocm/Tioga-975af3c/workspace 
^
|
(this has to refer to "Tioga-975af3c")
```

This adds some convenience functionality:

1. New command `benchpark system id <a dir generated by system-init>`: which will print the directory component related to the system (e.g. "Tioga-975af3c" in the above example)
2. `benchpark setup` now generates a symlink that follows the same pattern as for an experiment, using the basename of the system config dir. In terms of the above example, that would be `workspace/saxpy-rocm/tioga-system -> workspace/saxpy-rocm/Tioga-975af3c`

Notes:

* Keeping the hash-based link would help users that want to dynamically generate multiple workspace from a set of (experiment/system) tuples (the symlink would have to be removed for that to work well though)
* Until then (if ever) the symlink based on the simple name means a user can anticipate the name of the Ramble workspace generated by `benchpark setup`
* Either way, keeping the hash-name means that any existing scripts referring to system config dirs will continue to work
  * ... if that is important
  * if not, then this could just switch over to the simple name (i.e. the symlink becomes the actual path, and the hash is omitted)

### DIFF
--- a/.github/workflows/run.yml
+++ b/.github/workflows/run.yml
@@ -177,9 +177,10 @@ jobs:
         run: |
           ./bin/benchpark system init --dest=tioga-system tioga rocm=551 compiler=cce ~gtl
           ./bin/benchpark setup kripke/rocm ./tioga-system workspace/
+          system_id=$(./bin/benchpark system id ./tioga-system)
           . workspace/setup.sh
           ramble \
-            --workspace-dir workspace/kripke/rocm/Tioga-975af3c/workspace \
+            --workspace-dir "workspace/kripke/rocm/$system_id/workspace" \
             --disable-progress-bar \
             --disable-logger \
             workspace setup --dry-run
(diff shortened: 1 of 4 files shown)


pearce8 (approved):

pearce8:
Resolves #443.

PR #451, opened 2024-11-26
System.py for Fugaku

dyokelson:
Test system setup with: 
```
cd benchpark
. setup-env.sh
benchpark system init --dest fugaku-system fugaku 
```
Visually validate files created in fugaku-system directory

Run an existing experiment:
```
benchpark experiment init --dest saxpy-test saxpy +openmp
benchpark setup ./saxpy-test ./fugaku-system workspace/
```
Then follow with the ramble commands that are output. 

Will update qws for testing the new way as well, but best to use one we know works first to validate the system definition, then we can validate the new benchmark styles on fugaku.


jdomke:
- "random" crash with python 3.6 => created PR #461
- why is spack cloned twice? (`benchpark system init ...` does 1x into `~/.benchpark/spack/` and also `benchpark setup ...` does 1x cloning of spack into the workdir if the logs are correct)

jdomke (approved):
managed to replicate until submission (but just are waiting in the queue)

jdomke:
> managed to replicate until submission (but just are waiting in the queue)

runs as well

pearce8 (approved):

PR #462, opened 2024-12-01
features/branson

rfhaque:
## Description

Adding a specification of Branson https://lanl.github.io/benchmarks/01_branson/branson.html.

We should work with @gshipman and @alexrlongne to make progress on incorporating Branson.

- [x] Type of PR: Adding an experiment
- [x] Add a new `application.py` and (maybe) `package.py` under a new directory for this benchmark
- [x] Add an `experiment.py`
- [x] Define an `exec_mode=test` and `exec_mode=perf` experiment
- [x] Change the git location in `package.py` once the source PR https://github.com/lanl/branson/pull/53 is merged
- [x] Add to gitlab on 3 systems


repo/branson/package.py
```diff
@@ -39,16 +44,39 @@ class Branson(CMakePackage):
...
         args.append(f"-DCMAKE_C_COMPILER={spec['mpi'].mpicc}")
         args.append(f"-DCMAKE_CXX_COMPILER={spec['mpi'].mpicxx}")
         args.append(f"-DCMAKE_Fortran_COMPILER={spec['mpi'].mpifc}")
+
+        cflags = " ".join(self.compiler.flags['cflags']) if 'cflags' in self.compiler.flags else ""
+        cxxflags = " ".join(self.compiler.flags['cxxflags']) if 'cxxflags' in self.compiler.flags else ""
+
+        args.append("-DCMAKE_C_FLAGS={} -I{}/src/random123/features".format(cflags, self.stage.source_path))
+        args.append("-DCMAKE_CXX_FLAGS={} -I{}/src/random123/features".format(cxxflags, self.stage.source_path))
+        args.append(f"-DBUILD_TESTING=OFF")
```
rfhaque: @pearce8 This is just a temporary fix for lassen. Need to work with the branson team to fix build issues on that system.

pearce8 (changes requested):
Please add a dry run.

repo/branson/package.py
alexrlongne: I'll look at the RNG package again and see if an update fixes this.

pearce8:
@rfhaque Please update:  which machines does the code work on?  What are the outstanding items to fix?

rfhaque:
> @rfhaque Please update: which machines does the code work on? What are the outstanding items to fix?

The code works on all the LC systems (ruby, lassen, tioga) and on the LANL venado system. Once the cmake changes are upstreamed to the branson repo, we can merge this PR

pearce8 (changes requested):
@rfhaque Do you still want to merge this?  If so, please update.

repo/branson/package.py
```diff
@@ -6,8 +6,10 @@
 from spack.package import *
 from spack.pkg.builtin.boost import Boost
```
scheibelp: Referring to upstream packages changed in #953, but I think since `branson` was in the repo for that, you will get the appropriate update if merging develop into this.

scheibelp (commented):
minor compatibility issue w/ #953 (should be resolved w/ merge from develop)

slabasan:
@scheibelp `NotImplementedError: "branson" cannot run with MPI only without inheriting from MpiOnlyExperiment. Choose from ['rocm', 'cuda', 'openmp']`

scheibelp (dismissed):
I think this was just missing the `MpiOnlyExperiment` logic added in https://github.com/LLNL/benchpark/commit/ce3ffb6ba06bc50cbd8201c191250f56f8708b74#diff-9607e034d65374d4f8618d80af3e2d5033d973bdf09932cb07b5981232425232 - all the dry runs pass now.

I think someone other than me has to approve because I submitted the latest commits to this PR @slabasan, but I am adding an approving review to say the other parts look good to me.

rfhaque:
@scheibelp Does sparta need to provide an `mpi` variant for this to work? mpi is always on in sparta and thus a variant is not needed

scheibelp:
It appears to be necessary based on the logic added by ce3ffb6ba0 but @michaelmckinsey1 am I missing a way for a package that always uses mpi to avoid this?

michaelmckinsey1:
> It appears to be necessary based on the logic added by [ce3ffb6](https://github.com/LLNL/benchpark/commit/ce3ffb6ba06bc50cbd8201c191250f56f8708b74) but @michaelmckinsey1 am I missing a way for a package that always uses mpi to avoid this?

@scheibelp All programming models run with MPI, but to run only with MPI the experiment must inherit from `MpiOnlyExperiment`. This is to differentiate between experiments that cannot be ran only with MPI `experiment+mpi`, such as `babelstream+mpi`. `babelstream+mpi+cuda`, `babelstream+mpi+rocm`, `babelstream+mpi+openmp` are all valid, however `babelstream+mpi` is not. I believe babelstream may be the only current experiment with this behavior.

scheibelp:
I think what @rfhaque is saying is that it would never make sense to `experiment init branson ~mpi`, so the presence of an `mpi` variant is confusing.

@michaelmckinsey1 I think that means that if an `experiment init branson` would fail based on the logic in https://github.com/LLNL/benchpark/blob/develop/lib/benchpark/experiment.py#L279 (it wants mpi to be mentioned explicitly).

I was thinking perhaps https://github.com/LLNL/benchpark/blob/develop/.github/utils/dryruns.py was dispatching the wrong call, but it seems like it was generating the appropriate `experiment init branson` for [e830b9d](https://github.com/LLNL/benchpark/pull/462/commits/e830b9d64460443461bc813d51079d009403b95b), but that it was rejected by the logic in `experiment.py`.

Is it your opinion such experiments should have an `mpi` variant? Or am I missing a way for an experiment to be set up this way?

michaelmckinsey1:
The `mpi` variant is true by default, so that is why `experiment init branson` specs to `branson+mpi` without needing to explicitly write `+mpi`. I think it makes sense to have an explicit `mpi` variant because of the babelstream example. I don't see the test where the logic is failing, but I can take a look if it is.

slabasan (dismissed):
LGTM, but waiting on the package.py to be updated upstream?

michaelmckinsey1:
This PR raised some confusion about the `MpiOnlyExperiment` mechanism. It is described https://github.com/LLNL/benchpark/pull/904 and in the docs for writing an experiment. It should be generally included for every experiment, except cases like `babelstream`, which can not be ran for `babelstream+mpi`, but can be ran for `babelstream+mpi+cuda`. So it is used to indicate whether running `+mpi` without any programming models is valid.

Inheriting from `MpiOnlyExperiment` is used to check spec validity https://github.com/LLNL/benchpark/blob/dd1b469c84f23ad1bd334762a145cb0cb7b416a0/lib/benchpark/experiment.py#L259-L296. https://github.com/LLNL/benchpark/pull/931 may drive some changes to `MpiOnlyExperiment`.

pearce8 (approved):

PR #468, opened 2024-12-03
Aws system: integrate 6a/7a configs

scheibelp:
https://github.com/LLNL/benchpark/tree/develop/legacy/systems/nosite-AWS_PCluster_Hpc6a-zen3-EFA 

and

https://github.com/LLNL/benchpark/tree/develop/legacy/systems/nosite-AWS_PCluster_Hpc7a-zen4-EFA

are quite similar except for `variables.yaml` (different system specs)

The external packages config for each of these had a schema error:

```
  blas:
    externals:
    - spec: blas@3.4.2
      prefix: /usr
      buildable: false
```

should be

```
  blas:
    buildable: false
    externals:
    - spec: blas@3.4.2
      prefix: /usr
```

(this occurs multiple times, fixed in the template files used by the `aws-pcluster` system, but the original static files are not fixed)

### DIFF
--- a/systems/aws-pcluster/system.py
+++ b/systems/aws-pcluster/system.py
@@ -3,6 +3,8 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+import pathlib
+
 from benchpark.system import System
 from benchpark.directives import variant
 
@@ -28,7 +30,7 @@ id_to_resources = {
 }
 
 
-class Aws(System):
+class AwsPcluster(System):
     variant(
         "instance_type",
         values=("c6g.xlarge", "c4.xlarge", "hpc7a.48xlarge", "hpc6a.48xlarge"),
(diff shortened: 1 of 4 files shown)


systems/aws-pcluster/system.py
```diff
@@ -17,21 +19,76 @@
...
+        self.scheduler = "slurm"
         # TODO: for some reason I have to index to get value, even if multi=False
         attrs = id_to_resources.get(self.spec.variants["instance_type"][0])
         for k, v in attrs.items():
             setattr(self, k, v)
+
+    def system_specific_variables(self):
+        return {
+            "extra_cmd_opts": '--mpi=pmix --export=ALL,FI_EFA_USE_DEVICE_RDMA=1,FI_PROVIDER="efa",OMPI_MCA_mtl_base_verbose=100',
+        }
```
scheibelp, line 52: Right now, this produces experiment scripts like:

```
#SBATCH --mpi=pmix --export=ALL,FI_EFA_USE_DEVICE_RDMA=1,FI_PROVIDER="efa",OMPI_MCA_mtl_base_verbose=100
#SBATCH -n 8
#SBATCH -N 1
...
srun --mpi=pmix --export=ALL,FI_EFA_USE_DEVICE_RDMA=1,FI_PROVIDER="efa",OMPI_MCA_mtl_base_verbose=100 -n 8 -N 1 saxpy -n 512 >> ...
```

(i.e. these are generated as sbatch opts and command opts) - this is decided by the allocation modifier (specifically for the SLURM scheduler). I think it's worth checking if these should be batch opts

* If so, this can be rewritten to generate a separate batch opt for `--mpi` and `--export`
* If not, I can change the allocation modifier

I don't think this decision need stop this PR, but the generated experiment scripts will have weird sbatch directives until that is decided.

pearce8 (approved):

PR #469, opened 2024-12-03
includes: name mangle ramble and include from spack normally

becker33:
## Description

This PR allows us to import from Spack in the normal way instead of our crazy workaround.

The key is that we name mangle all of the ramble imports from `spack.*` to `ramble_spack.*` and `llnl.*` to `ramble_llnl.*`.

Fixes issue(s): FIXME:Add list of relevant issues.

## Type of Change

- { } Documentation update
- { } Build/CI update

## Checklist:

If adding/modifying a system:
- { } Create a new directory for the system and a new `system.py` file

      for this benchmark
- { } Configure an experiment
      section

- { } Update docs

### DIFF
--- a/lib/benchpark/cmd/setup.py
+++ b/lib/benchpark/cmd/setup.py
@@ -20,10 +20,28 @@ from benchpark.debug import debug_print
 from benchpark.runtime import RuntimeResources
 import benchpark.system
 
-bootstrapper = RuntimeResources(benchpark.paths.benchpark_home)  # noqa
-bootstrapper.bootstrap()  # noqa
 
-import llnl.util.link_tree  # noqa
+# Note: it would be nice to vendor spack.llnl.util.link_tree, but that
+# involves pulling in most of llnl/util/ and spack/util/
+def symlink_tree(src, dst, include_fn=None):
+    """Like ``cp -R`` but instead of files, create symlinks"""
+    src = os.path.abspath(src)
+    dst = os.path.abspath(dst)
+    # By default, we include all filenames
+    include_fn = include_fn or (lambda f: True)
+    for x in [src, dst]:
+        if not os.path.isdir(x):
+            raise ValueError(f"Not a directory: {x}")
+    for src_subdir, directories, files in os.walk(src):
+        relative_src_dir = pathlib.Path(os.path.relpath(src_subdir, src))
+        dst_dir = pathlib.Path(dst) / relative_src_dir
+        dst_dir.mkdir(parents=True, exist_ok=True)
+        for x in files:
+            if not include_fn(x):
+                continue
+            dst_symlink = dst_dir / x
+            src_file = os.path.join(src_subdir, x)
+            os.symlink(src_file, dst_symlink)
 
 
 def setup_parser(root_parser):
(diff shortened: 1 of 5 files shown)


lib/benchpark/cmd/setup.py
```diff
@@ -186,23 +168,25 @@ def command(args):
...
-    symlink_tree(
-        source_dir / "legacy" / "systems" / "common",
-        ramble_spack_experiment_configs_dir,
-        include_fn,
+        if fname.endswith(".yaml"):
+            return False
+        return True
+
+    configs_tree = llnl.util.link_tree.LinkTree(configs_src_dir)
+    experiment_tree = llnl.util.link_tree.LinkTree(experiments_src_dir)
```
scheibelp: ref error: `experiments_src_dir` -> `experiment_src_dir`

lib/benchpark/runtime.py
```diff
@@ -82,21 +83,52 @@ def __init__(self, dest):
...
+
         if not self.ramble_location.exists():
             self._install_ramble()
+
         ramble_lib_path = self.ramble_location / "lib" / "ramble"
+
+        # If ramble is present but not yet name mangled, do that
+        ramble_spack_path = ramble_lib_path / "ramble_spack"
+        if not ramble_spack_path.exists():
+            self._mangle_ramble()
```
scheibelp, line 106: I think ideally we would conceptually distinguish the case where we clone ramble as a library vs. when we clone it to run it.

* e.g. `_install_ramble` can be called from `bootstrap()` or `.ramble()` (via `ramble_first_time_setup`)
* I'd recommend making the "library-ness" a property of the `RuntimeResources` object, and move this consideration to `_install_ramble()`
    * And maybe rename `_install_ramble` to reflect that it can update existing installations (i.e. so that this works for users that do a `git pull` of Benchpark with an already-established Ramble lib prefix

lib/benchpark/runtime.py
```diff
@@ -82,21 +83,52 @@ def __init__(self, dest):
...
         if internals not in sys.path:
             sys.path.insert(1, internals)
 
-        # Spack does not go in sys.path, but we will manually access modules from it
-        # The reason for this oddity is that spack modules will compete with the internal
-        # spack modules from ramble
-        if not self.spack_location.exists():
-            self._install_spack()
+    def _mangle_ramble(self):
+        """Name mangle ramble spack.* symbols to allow separate spack imports."""
```
scheibelp, line 116: Can you add a bit of detail: mention we are replacing all instances of `spack` with `ramble_spack`, and all instances of `llnl` with [...].

Might be worth broadly outlining why this is expected to work: I don't think this name mangling would make all parts of Spack work, but I believe they make all the parts of Spack that Ramble uses work.

lib/benchpark/runtime.py
```diff
@@ -82,21 +83,52 @@ def __init__(self, dest):
...
+        import llnl.util.filesystem as fs
+
+        files = [
+            f
+            for f in fs.find(self.ramble_location, "*")
+            if os.stat(f).st_mode & stat.S_IWUSR and not os.path.isdir(f)
```
scheibelp, line 122: If we don't own these files, we should probably fail, vs. silently doing nothing.

lib/benchpark/runtime.py
```diff
@@ -82,21 +83,52 @@ def __init__(self, dest):
...
+        ]
+        file_filter = fs.FileFilter(*files)
+        # Don't replace if it's already replaced or if it's a field in an existing module
```
scheibelp, line 125: I think trying to make these regexes idempotent will make them more confusing than they need to be: the existence of the `ramble_spack` directory can function as a sentinel (if you are worried about `CTRL-C`, I'd recommend a marker file); if we depend on that, I think we can stop considering `_`, which will simplify the regex.

Either way, I think this will work as long as:

```
# spack always does this
import spack.util.spack_yaml
# and never does this (which is allowed)
from spack.util import spack_yaml
```

But IMO it would be more straightforward to look for `[\s]spack` (i.e. spaces followed by `spack`)

I think including some examples of things you intend *not* to match would be useful (e.g. `...spack_yaml`

scheibelp (changes requested):
Overall this looks reasonable but

* There's a couple spots that could use documentation
* The mangling of Ramble-as-a-lib vs. Ramble-as-a-driver is handled implicitly
* I'd prefer some tweaks to the regexes (or at least some explanation in a docstring)

lib/benchpark/runtime.py
becker33: I'll add a comment -- the main purpose here isn't to catch files we don't own, it's to catch files that aren't writable (like some things within the `.git` repo).

scheibelp: I assume nothing under `.git` should be modified: I recommend skipping it entirely (and explicitly).

PR #494, opened 2024-12-12
Add an 'all' option to generate a ramble experiment for each defined workload

### LINKED ISSUE #491: Experiment with multiple workloads
We should be able to specify an experiment with multiple workloads, as long as they use the same resources.  This was possible in ramble.yaml, see legacy/experiments/amg2023 for an example (not reflected in the amg2023/experiment.py, where the 2 workloads now become 2 separate jobs).

The current use case is the OSU microbenchmarks which has several dozen tests we want to run in the same allocation back to back.

rfhaque:
This PR adds an implicit workload=all option for the benchmarks.
- workload variant must be defined as a multi-valued variant (set `multi=True` in the workload variant definition)
- Using `workload=all` will generate an experiment for each of the defined workloads keeping all other experiment parameters unchanged. It is the responsibility of the user to ensure that this is a valid scenario
- `all` is a reserved option for the workload variant. An explicitly defined `all` option would be treated the same way as the reserved keyword.

`bin/benchpark experiment init --dest ./osu osu-micro-benchmarks workload=all`

Potentially closes #491.


lib/benchpark/spec.py
```diff
@@ -295,6 +296,20 @@ def _concretize(self):
...
+                    self.variants["workload"] = variant_map["workload"].default
+            if "all" in self.variants["workload"]:
+                if variant_map["workload"].multi:
+                    del self.variants["workload"]
+                    self.variants["workload"] = variant_map["workload"].values
+                else:
+                    raise BenchparkError(f"The workload variant in must be multi-valued to use the 'all' option in package {self.name}")
+        else:
+            raise BenchparkError(f"Package {self.name} must define a workload variant")
+
```
rfhaque: @scheibelp @pearce8 Please review

experiments/osu-micro-benchmarks/experiment.py
```diff
@@ -0,0 +1,116 @@
...
+            "osu_get_acc_latency",
+            "osu_get_bw",
+            "osu_get_latency",
+            "osu_put_bibw",
+            "osu_put_bw",
+            "osu_put_latency",
+            "osu_hello",
+            "osu_init",
+        ),
+        multi=True,
```
rfhaque, line 88: @nhanford Setting the workload to be a multi-valued variant allows for subsets, setting `workload=all` generates the experiment for all workloads

pearce8 (approved):

pearce8 (approved):

PR #500, opened 2024-12-14
[ARCHIVED] Feature/single node check

### LINKED ISSUE #492: Single node experiment
- enforce that a single node experiment is implemented
- sub-select and run a single node experiment
- enforce (in the allocation modifier?) that the experiment doesn't request more experiments

rfhaque:
## Description

~~This PR adds a check for the single node experiment~~
~~- The single_node variant [default=True] ONLY asserts that the given experiment must run on 1 node~~
~~- This check is enforced in the `determine_allocation` method of the allocation modifier. If single node is enabled and the experiment requires/requests more than 1 node, the ramble setup throws an error~~
~~- strong/weak scaling modes are decoupled from the single node mode (it's possible to scale on a single node)~~
~~- An experiment trying to run on more than one node must explicitly specify `~single_node`~~

Take 2:
- We add an integer variant `max_node_limit` to the system specs (`lib/benchpark/system.py`) which is the max number of nodes allowed for an experiment **on that system instantiation**. Default value 1, value of 0 means no limits
- In the experiment setup (`lib/benchpark/experiment.py`), we add a filter clause `'{n_nodes} > 0 and {n_nodes} <= {max_node_limit}'`. 
- The exclude clause filters all experiments that require more than `max_node_limit` nodes during `ramble setup`. Currently, it's not an error if zero experiments satisfy the `max_node_limit`

```
./bin/benchpark system init --dest=tioga llnl-elcapitan rocm=5.5.1 compiler=cce max_node_limit=1
./bin/benchpark experiment init --dest ./saxpy saxpy +openmp~cuda~rocm caliper=mpi,time
./bin/benchpark setup ./saxpy ./tioga workspace/
```

(Potentially) Fixes #492

## Type of Change

- { } Documentation update
- { } Build/CI update

## Checklist:

If adding/modifying a system:
- { } Create a new directory for the system and a new `system.py` file

      for this benchmark
- { } Configure an experiment
      section

- { } Update docs


modifiers/allocation/modifier.py
```diff
@@ -291,6 +293,12 @@ def determine_allocation(self, v):
                     )
             v.n_nodes = max(cores_node_request or 0, gpus_node_request or 0)
 
+        if v.run_single_node_expr and v.n_nodes > 1:
+            raise ValueError(
+                f"Experiment must run on 1 node. Requested {v.n_nodes} nodes"
+            )
+
+
```
rfhaque: @scheibelp Please review

pearce8:
@rfhaque what is the status of this PR?  How does it currently work?  What docs would you need to add?

When you want to run strong or weak scaling, do you need to explicitly disable the single-node test because only one variant is possible at a time?  If that is the case, we need to document it well.

rfhaque:
ARCHIVED

PR #501, opened 2024-12-16
add stage to skip if branch not a PR, default job is interruptible

slabasan:
- Skip CI if branch is not a pull request
- Make all jobs interruptible by default, so new commits can cancel workflow


slabasan:
@pearce8 Ready for merge

pearce8 (approved):

PR #515, opened 2025-01-02
El Capitan system: always include GCC compiler

scheibelp:
And add preference for `%cce` when setting `compiler=cce`

Some low level dependencies won't build with `%cce`, so this adds a GCC compiler definition regardless of what the user requests. To avoid using GCC in other cases, this adds a preference for `%cce`.

The changes here do not *guarantee* using `%cce`: doing so would likely require tighter coupling between the system and experiment class (which is in the works).

### DIFF
--- a/systems/llnl-elcapitan/system.py
+++ b/systems/llnl-elcapitan/system.py
@@ -106,7 +106,29 @@ class LlnlElcapitan(System):
         elif compiler == "gcc":
             selections.append(externals / "libsci" / "00-gcc-packages.yaml")
 
+        cmp_preference_path = self.next_adhoc_cfg()
+        with open(cmp_preference_path, "w") as f:
+            f.write(self.compiler_weighting_cfg())
+        selections.append(cmp_preference_path)
+
         return selections
+
+    def compiler_weighting_cfg(self):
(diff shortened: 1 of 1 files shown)


rfhaque:
@scheibelp Is is possible to add versions for cce (e.g. cce@16.0.0, cce@18.0.0 etc), cray-mpich and other dependencies in `llnl-elcapitan/system.py`

scheibelp:
> Is is possible to add versions for cce (e.g. cce@16.0.0, cce@18.0.0 etc), cray-mpich and other dependencies in

Yes, in some sense handling this is a separate concern, and I could do so if you like (preferably in another PR).

In more detail: as a direct change, you take the current code in `llnl-elcapitan/system.py`

```
        if compiler == "cce":
            compiler_cfg_path = self.next_adhoc_cfg()
            with open(compiler_cfg_path, "w") as f:
                f.write(self.rocm_cce_compiler_cfg(rocm, "16.0.0"))
```

and change "16.0.0" to "18.0.0". To support both, I think it would make sense to add a conditional variant like

```
    variant(cce_version, values=("16", "18"), when="compiler=cce")
```

And then similarly, you would only need to change the one line

pearce8 (approved):

PR #519, opened 2025-01-06
Adds support for Corona to the llnl-cluster system

ilumsden:
## Description

This PR modifies the `llnl-cluster` system to also support Corona. This required a few Corona-specific tweaks to the `LlnlCluster` class, mainly because, unlike the other clusters associated with this system, Corona has AMD GPUs. Most changes relate to adding functions for injecting ROCm into the packages and compilers.

Dependencies: None

Fixes issue(s): None

## Type of Change

- {X} Modifying an existing system, benchmark, or experiment
- { } Documentation update
- { } Build/CI update

## Checklist:

If adding/modifying a system:
- {X} Create a new directory for the system and a new `system.py` file

      for this benchmark
- { } Configure an experiment
      section

- { } Update docs

### DIFF
--- a/systems/llnl-cluster/externals/base/00-packages.yaml
+++ b/systems/llnl-cluster/externals/base/00-packages.yaml
@@ -41,7 +41,7 @@ packages:
   hwloc:
     buildable: false
     externals:
-    - spec: hwloc@2.9.1
+    - spec: hwloc@2.11.1
       prefix: /usr
   fftw:
     buildable: false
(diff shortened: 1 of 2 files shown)


pearce8:
@ilumsden thanks for this!  There may be 2 solutions:
1) corona could be one of the llnl-cluster
2) it might need to be its own system.py

I think this largely depends on how we are implementing "provides" and "requires".  Since corona has GPUs/rocm, it would need to "provide" rocm.  @becker33, it would be good to have you weigh in here.

pearce8:
This PR is way out of date, we have since started specifying what runtime the system provides.  Since corona provides rocm, it can no longer be one of the llnl-clusters.  If we need a system description for corona, we will need to start over.

PR #548, opened 2025-01-15
venado system specs

rfhaque:
## Description

This PR adds the system config specs for the LANL Venado system.

1. saxpy errors out since CUDA 12.5 does not support clang 18 as the host compiler. Adding "-allow-unsupported-compiler" does not work (or the flag isn't begin added correctly by spack) 
2. caliper fails when apply a patch for the aarch64 architecture in spack 

Fixes issue(s): FIXME:Add list of relevant issues.

## Type of Change

- {x} Adding a system, benchmark, or experiment
- { } Documentation update
- { } Build/CI update

## Checklist:

If adding/modifying a system:
- {X} Create a new directory for the system and a new `system.py` file
- { X} Add a new dry run unit test in `.github/workflows`
- {X} System appears in System Specifications table in docs catalogue section

      for this benchmark
- { } Configure an experiment
      section

- { } Update docs


pearce8:
@rfhaque where is the PR checklist template?  Those are helpful for reviewing.  In particular, for adding a new system, we also need:
{ } Add a human-readable system description to systems/all_system_definitions
{ } Verify info added in all_system_definitions generates a new system column in Catalogue/System Specifications (https://software.llnl.gov/benchpark/system-list.html)  
{ } Add a dry run of an existing benchmark using the new system definition

(@slabasan, if those are not listed in the PR template right now, we probably want to add them)

repo/caliper/for_aarch64.patch
pearce8, line 1: @daboehme Can you please see the patch for Caliper on Arm?  We are working with a Grace Hopper machine.  It would be great if you can fix this in Caliper and put up a PR to benchpark to remove the patch (if this PR is already merged).

repo/caliper/sampler-service-missing-libunwind-include-dir.patch
pearce8, line 1: Also this patch @daboehme

repo/caliper/package.py
pearce8, line 1: @rfhaque What is different about this spack package from upstream?  We need to work with @dbo to review/upstream it.

repo/caliper/for_aarch64.patch
daboehme: This should be fixed in Caliper v2.12.1, which is the current default in upstream spack

repo/caliper/sampler-service-missing-libunwind-include-dir.patch
daboehme: This one was only required in v2.9.0, which is quite outdated at this point

systems/lanl-venado/system.py
```diff
@@ -0,0 +1,214 @@
...
+        "sys_cores_per_node": 144,
+        "sys_gpus_per_node": 4,
+    },
+    "grace-grace": {
+        "sys_cores_per_node": 144,
+    },
+}
+
+
+class LanlVenado(System):
```
pearce8, line 22: @rfhaque Do they call them "clusters" or "partitions"?  Do you also need to set "partition" in the slurm script to correspond to which partition of the machine you are wanting to use?

@alexrlongne might know.

systems/lanl-venado/system.py
```diff
@@ -0,0 +1,214 @@
...
+        sw_description = pathlib.Path(output_dir) / "software.yaml"
+
+        with open(sw_description, "w") as f:
+            f.write(self.sw_description())
+
+    def system_specific_variables(self):
+        return {
+            "cuda_arch": "90",
+            "default_cuda_version": self.spec.variants["cuda"][0],
+            "extra_batch_opts": '"-A llnl_ai_g -pgpu"',
```
pearce8, line 85: @rfhaque what does this extra_batch_opts do?

repo/caliper/for_aarch64.patch
pearce8: @daboehme thanks!

@rfhaque are you using the tip of develop of Caliper which should be v2.12.1?

repo/caliper/sampler-service-missing-libunwind-include-dir.patch
pearce8: @rfhaque can you please remove the patches and use the newest Caliper instead?  Is there a reason you are not doing so?

rfhaque: @pearce8 This is just a temporary fix, till we pin the latest spack version.

systems/lanl-venado/system.py
rfhaque: @pearce8 While technically these are hardware partitions, venado slurm defines its own "partitions" (standard, debug, gpu etc) corresponding to some set of resources/qos

systems/lanl-venado/system.py
rfhaque: @pearce8 The extra_batch_opts requests the required partition and account. We may need to parameterize this

pearce8 (approved):

PR #587, opened 2025-02-03
Benchpark info command, new experiment and system directives

### LINKED ISSUE #569: Benchpark info command, new experiment and system directives
New system.py directives:
```
system_site("lanl")
hardware_key(from all_system_definitions/system_hardware/hardware_description.yaml)
maintainer("...")
tags
```

New experiment.py directives:
```
url = ...
spack_name = "amg2023"
ramble_name = "amg"
maintainer("...")
tags
```

Parent classes (CudaExperiment, etc.) directives:
```
programming_model("cuda")
scaling("strong")
tags

```

``Benchpark info`` command needs the following subcommands:
- [x] benchpark info system 
- [x] Benchpark info system variants
- [x] benchpark info system hardware
- [x] benchpark info system maintainer
- [x] benchpark info experiment_name
- [x] benchpark info --spack experiment_name
- [x] benchpark info --ramble experiment_name
- [ ] benchpark info maintainer
- [ ] list from #486 

This is a learning opportunity on the Benchpark project, and @scheibelp and @becker33 are available for iteration on the design and questions.

@becker33 Please add items or correct the list here.

michaelmckinsey1:
## Description

resolves [benchpark/569](https://github.com/LLNL/benchpark/issues/569) by implementing the `benchpark info` command, and adding new directives.

## Adding/modifying a system

- [x] Modify `systems/llnl-cluster/system.py` (and presumably other systems): to refactor `id_to_resources`, adding the `hardware_key`, unless this becomes a separate directive.
- [x] Modify `systems/*/system.py` for `id_to_resources`, `system_site`, `maintainer`
- [x] Modify `experiments/*/experiment.py`

## Adding/modifying core functionality

- [x] Add `lib/benchpark/cmd/info.py` & modify `lib/main.py`: enables `benchpark info ...`
   - [x] Add `benchpark info system`
   - [x] Add `benchpark info experiment_name`
- [x] Modify `lib/benchpark/directives.py`: to add the `system_site` and `maintainer` directives.

## Docs

- [x] Modify `docs/basic-usage.rst`

## Uses
```
$ benchpark info experiment -h
usage: main.py info experiment [-h] [--spack] [--ramble] [--url] [--variants] [--maintainer] spec [spec ...]

positional arguments:
  spec          Experiment spec

options:
  -h, --help    show this help message and exit
  --spack       Information from Spack package
  --ramble      Information from Ramble package
  --url         URL for experiment
  --variants    Available experiment variants
  --maintainer  Maintainer
```

```
$ benchpark info system -h
usage: main.py info system [-h] [--variants] [--hardware] [--system-site] [--maintainer] spec [spec ...]

positional arguments:
  spec           System spec

options:
  -h, --help     show this help message and exit
  --variants     Available system variants
  --hardware     Hardware descriptions per cluster
  --system-site  System site
  --maintainer   Maintainer
```

### `benchpark info system`
image
...
image

### `benchpark info experiment`
image

### `benchpark info system --query`
Query the id_to_resources dictionary for key/value pairs.
image
image

### DIFF
--- a/experiments/amg2023/experiment.py
+++ b/experiments/amg2023/experiment.py
@@ -42,10 +42,6 @@ class Amg2023(
 
     url = "https://github.com/LLNL/AMG2023"
 
-    spack_name = "amg2023"
-
-    ramble_name = "amg2023"
-
     # requires("system+papi", when(caliper=topdown*))
 
     # TODO: Support list of 3-tuples
--- a/experiments/amg2023/experiment.py
+++ b/experiments/amg2023/experiment.py
@@ -40,7 +40,6 @@ class Amg2023(
 
     maintainers("pearce8")
 
-
     # requires("system+papi", when(caliper=topdown*))
 
     # TODO: Support list of 3-tuples
--- a/experiments/amg2023/experiment.py
+++ b/experiments/amg2023/experiment.py
@@ -4,7 +4,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 from benchpark.error import BenchparkError
-from benchpark.directives import variant
+from benchpark.directives import variant, maintainer
 from benchpark.experiment import Experiment
 from benchpark.openmp import OpenMPExperiment
 from benchpark.cuda import CudaExperiment
@@ -37,6 +37,14 @@ class Amg2023(
         default="develop",
         description="app version",
     )
+
+    maintainer("")
+
+    url = "https://github.com/LLNL/AMG2023"
+
+    spack_name = "amg2023"
+
+    ramble_name = "amg2023"
 
     # requires("system+papi", when(caliper=topdown*))
 
--- a/experiments/amg2023/experiment.py
+++ b/experiments/amg2023/experiment.py
@@ -38,6 +38,8 @@ class Amg2023(
         description="app version",
     )
 
+    maintainers("pearce8")
+
     url = "https://github.com/LLNL/AMG2023"
 
     # requires("system+papi", when(caliper=topdown*))
--- a/experiments/amg2023/experiment.py
+++ b/experiments/amg2023/experiment.py
@@ -40,7 +40,6 @@ class Amg2023(
 
     maintainers("pearce8")
 
-    url = "https://github.com/LLNL/AMG2023"
 
     # requires("system+papi", when(caliper=topdown*))
 
--- a/experiments/amg2023/experiment.py
+++ b/experiments/amg2023/experiment.py
@@ -4,7 +4,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 from benchpark.error import BenchparkError
-from benchpark.directives import variant, maintainer
+from benchpark.directives import variant, maintainers
 from benchpark.experiment import Experiment
 from benchpark.openmp import OpenMPExperiment
 from benchpark.cuda import CudaExperiment
@@ -37,8 +37,6 @@ class Amg2023(
         default="develop",
         description="app version",
     )
-
-    maintainer("")
 
     url = "https://github.com/LLNL/AMG2023"
 
--- a/experiments/stream/experiment.py
+++ b/experiments/stream/experiment.py
@@ -3,7 +3,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-from benchpark.directives import variant, maintainer
+from benchpark.directives import variant, maintainers
 from benchpark.experiment import Experiment
 from benchpark.caliper import Caliper
 
@@ -23,8 +23,6 @@ class Stream(
         default="5.10",
         description="app version",
     )
-
-    maintainer("")
 
     url = "https://www.cs.virginia.edu/stream/ref.html"
 
--- a/experiments/stream/experiment.py
+++ b/experiments/stream/experiment.py
@@ -24,6 +24,7 @@ class Stream(
         description="app version",
     )
 
+    maintainers("daboehme","rfhaque")
     url = "https://www.cs.virginia.edu/stream/ref.html"
 
     def compute_applications_section(self):
--- a/experiments/stream/experiment.py
+++ b/experiments/stream/experiment.py
@@ -25,7 +25,7 @@ class Stream(
     )
 
     maintainers("daboehme","rfhaque")
-    url = "https://www.cs.virginia.edu/stream/ref.html"
+
 
     def compute_applications_section(self):
 
--- a/experiments/stream/experiment.py
+++ b/experiments/stream/experiment.py
@@ -3,7 +3,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-from benchpark.directives import variant
+from benchpark.directives import variant, maintainer
 from benchpark.experiment import Experiment
 from benchpark.caliper import Caliper
 
@@ -23,6 +23,14 @@ class Stream(
         default="5.10",
         description="app version",
     )
+
+    maintainer("")
+
+    url = "https://www.cs.virginia.edu/stream/ref.html"
+
+    spack_name = "stream"
+
+    ramble_name = "stream"
 
     def compute_applications_section(self):
 
--- a/experiments/stream/experiment.py
+++ b/experiments/stream/experiment.py
@@ -24,8 +24,7 @@ class Stream(
         description="app version",
     )
 
-    maintainers("daboehme","rfhaque")
-
+    maintainers("daboehme", "rfhaque")
 
     def compute_applications_section(self):
 
--- a/experiments/stream/experiment.py
+++ b/experiments/stream/experiment.py
@@ -28,10 +28,6 @@ class Stream(
 
     url = "https://www.cs.virginia.edu/stream/ref.html"
 
-    spack_name = "stream"
-
-    ramble_name = "stream"
-
     def compute_applications_section(self):
 
         array_size = {"s": 650000000}
--- a/lib/benchpark/cmd/info.py
+++ b/lib/benchpark/cmd/info.py
@@ -1,3 +1,4 @@
+import os
 import subprocess
 import textwrap
 
@@ -65,6 +66,31 @@ def info_system(args):
                 else:
                     resource_key = "@*r" + resource_key + "@."
                     color.cprint(f"{indent()*2}{resource_key}: {resource_value}")
+
+    def _handle_query(query):
+        key, value = query.split("=", 1)
+        exclude = {"all_hardware_descriptions", "repo.yaml", "generic-x86"}
+        all_system_specs = []
+        for d in set(os.listdir(benchpark.paths.benchpark_root / "systems")) - exclude:
+            all_system_specs.append(benchpark.spec.SystemSpec(d))
+
+        matching_systems = []
+        for spec in all_system_specs:
+            resources = spec.system_class.id_to_resources
+            for cluster, resource_dict in resources.items():
+                if str(resource_dict.get(key)) == value:
+                    matching_systems.append((spec.name, cluster))
+
+        if matching_systems:
+            gen_header(query)
+            for system_name, cluster in matching_systems:
+                color.cprint(f"{indent()}@*g{system_name}@.: {cluster}")
+        else:
+            print("No matching systems found.")
+
+    if args.query:
+        _handle_query(args.query)
+        return
 
     system_spec = benchpark.spec.SystemSpec(args.name)
     system_class = system_spec.system_class
@@ -153,7 +179,10 @@ def setup_parser(root_parser):
     )
     system_parser.add_argument("--system-site", action="store_true", help="System site")
     system_parser.add_argument("--maintainers", action="store_true", help="Maintainers")
-    system_parser.add_argument("name", help="System name")
+    system_parser.add_argument(
+        "--query", type=str, help="Query systems with key-value pairs (e.g., key=value)"
+    )
+    system_parser.add_argument("name", help="System name", nargs="?")
 
     experiment_parser = info_subparser.add_parser("experiment")
     experiment_parser.add_argument(
--- a/lib/benchpark/cmd/info.py
+++ b/lib/benchpark/cmd/info.py
@@ -125,10 +125,10 @@ def info_experiment(args):
     experiment_class = conc.experiment
 
     if args.spack:
-        subprocess.run(["spack", "info", experiment_class.spack_name])
+        subprocess.run(["spack", "info", experiment_class.spack_name if experiment_class.spack_name else experiment_class.name])
         return
     elif args.ramble:
-        subprocess.run(["ramble", "info", experiment_class.ramble_name])
+        subprocess.run(["ramble", "info", experiment_class.ramble_name if experiment_class.ramble_name else experiment_class.name])
         return
     else:
         actions = {
--- a/lib/benchpark/cmd/info.py
+++ b/lib/benchpark/cmd/info.py
@@ -1,5 +1,3 @@
-import glob
-import sys
 import textwrap
 
 import yaml
@@ -62,14 +60,6 @@ def info_system(args):
         info_system_variants(system_spec)
         info_system_hardware(system_spec)
 
-    # conc_system_spec = system_spec.concretize()
-    # system = conc_system_spec.system
-    # system.initialize()
-
-    # print("\n\n\n\n")
-    # print([i for i in system.__dict__.keys()])
-    # print(system.sw_description())
-
 
 def setup_parser(root_parser):
     info_subparser = root_parser.add_subparsers(dest="info_subcommand")
@@ -92,7 +82,6 @@ def command(args):
         "system": info_system,
     }
     if args.info_subcommand in actions:
-        print(args)
         actions[args.info_subcommand](args)
     else:
         raise ValueError(f"Unknown subcommand for 'info': {args.info_subcommand}")
--- a/lib/benchpark/cmd/info.py
+++ b/lib/benchpark/cmd/info.py
@@ -9,56 +9,56 @@ def gen_header(name):
     print("=" * 20 + f"\n=== {name}\n" + "=" * 20)
 
 
-def info_system_system_site(system_spec):
-    gen_header("System Site")
-    print(system_spec.system_class.system_site)
-
-
-def info_system_maintainer(system_spec):
-    gen_header("Maintainer")
-    print(system_spec.system_class.maintainer)
-
-
-def info_system_variants(system_spec):
-    gen_header("Variants")
-    # Dict with one value
-    all_variants = list(system_spec.system_class.variants.values())[0]
-    for var in all_variants.values():
-        for attr in var.__dict__.keys():
-            print(attr + ":", var.__dict__[attr])
-        print()
-        # print(var.name+":", var.values)
-
-
-def info_system_hardware(system_spec):
-    gen_header("Hardware")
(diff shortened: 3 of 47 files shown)


systems/llnl-cluster/system.py
```diff
@@ -5,24 +5,35 @@
...
+            "sys_cores_per_node": 96,
+            "hardware_key": str(benchpark.paths.hardware_descriptions)
+            + "/Penguin-icelake-OmniPath/hardware_description.yaml",
+        },
+        "dane": {
+            "sys_cores_per_node": 112,
+            "hardware_key": str(benchpark.paths.hardware_descriptions)
+            + "/DELL-sapphirerapids-OmniPath/hardware_description.yaml",
+        },
+    }
```
michaelmckinsey1: I refactored this object into the class such that it could be accessed in `info.py` from the class object without importing `systems/llnl-cluster/system.py`. I added the new key, `hardware_key` to `id_to_resources`, because this _sometimes_ maps to directly to the machine.

1. Should `hardware_key` be a directive instead of existing in `id_to_resources`? It was listed as a directive in the issue, but I thought this makes more sense. 
2. Should `benchpark info system llnl-cluster --hardware` include other keys in `id_to_resources` (e.g. `sys_cores_per_node`) or only include the `hardware_description.yaml` information?

systems/llnl-sierra/system.py
```diff
@@ -42,8 +70,9 @@ def initialize(self):
         super().initialize()
 
         self.scheduler = "lsf"
-        self.sys_cores_per_node = "44"
-        self.sys_gpus_per_node = "4"
```
michaelmckinsey1, line 46: I noticed `llnl-sierra` and `riken-fugaku` do not assign these attributes using `id_to_resources` like the other systems. Is there a reason this can't be changed?

pearce8: It can be changed if it helps - does it, and if so, how?  We started with the way sierra is specified, and the struct appeared later when we started using the same system spec for multiple systems.  Maybe this is an opportunity for redesign - it would be nice to have a standard, easy to maintain, and not error prone mechanism.  Ideas?

pearce8:
@scheibelp This is work in progress at the moment, but as Michael is becoming familiar with the codebase, I'd like you to help make sure we are using existing Spack functionality where possible, and keep implementation similar/compatible where it is not.

systems/llnl-sierra/system.py
michaelmckinsey1: `llnl-sierra` and `riken-fugaku` were probably left like this because `sys_cores_per_node` and `sys_gpus_per_node` are the same for `lassen` and `sierra`, and `riken-fugaku` has only 1 system. So there isn't any technical difference, but the change I made makes them have the same structure as all of the other systems.

lib/benchpark/directives.py
```diff
@@ -104,6 +104,34 @@ def _make_when_spec(
...
+
+    def _execute_system_site(pkg):
+        pkg.system_site = name
+
+    return _execute_system_site
+
+
+@benchpark_directive("maintainer")
+def maintainer(name: str):
+    """Define the maintainer of the system.
```
scheibelp: This is used for experiments and systems

experiments/stream/experiment.py
```diff
@@ -24,6 +24,14 @@ class Stream(
         description="app version",
     )
 
+    maintainer("")
```
scheibelp: Does something go wrong if this directive doesn't appear at all? I think `maintainer()` should be left off most experiments/systems in this PR (and added later)

lib/benchpark/cmd/info.py
```diff
@@ -0,0 +1,175 @@
...
+        for attr, value in var.__dict__.items():
+            color.cprint(indent() + "@*g" + attr + "@.: " + str(value))
+        print()
+
+
+def info_maintainer(spec_class):
+    """SystemSpec.system_class or ExperimentSpec.experiment_class"""
+    gen_header("Maintainer")
+    mstr = indent()
+    if spec_class.maintainer == "":
```
scheibelp: Related to my other comment about being able to omit `maintainer`, we could use `getattr(spec_class, "maintainer", None)`

lib/benchpark/directives.py
```diff
@@ -104,6 +104,34 @@ def _make_when_spec(
...
+    """Define the site where the system is located.
+
+    Arguments:
+        name: Name of the site
+    """
+
+    def _execute_system_site(pkg):
+        pkg.system_site = name
+
+    return _execute_system_site
```
scheibelp: I think this would make more sense as an attribute vs. a directive.

Reasons you might want to make something a directive:

* It functions like an attribute but you want to merge across subclasses vs. override
* It has potential modifying information

Neither of these things are true for maintainers either (although I assume that was modeled as a directive because spack does it that way).

lib/benchpark/cmd/info.py
```diff
@@ -0,0 +1,175 @@
...
+    experiment_parser.add_argument(
+        "--url", action="store_true", help="URL for experiment"
+    )
+    experiment_parser.add_argument(
+        "--variants", action="store_true", help="Available experiment variants"
+    )
+    experiment_parser.add_argument(
+        "--maintainer", action="store_true", help="Maintainer"
+    )
+    experiment_parser.add_argument("spec", nargs="+", help="Experiment spec")
```
scheibelp: I think this should just take a name, vs. an arbitrary spec (you can construct a spec to retrieve the system object, but I recommend describing it here as a system name, and only accepting one additional argument vs. `+`)

experiments/amg2023/experiment.py
```diff
@@ -38,6 +38,14 @@ class Amg2023(
         description="app version",
     )
 
+    maintainer("")
+
+    url = "https://github.com/LLNL/AMG2023"
+
+    spack_name = "amg2023"
+
+    ramble_name = "amg2023"
```
scheibelp: I recommend adding something to `Experiment` to to to explain the meaning of these variables:

```
@property
def spack_name(self):
    # The name of the spack package that is used to build this benchmark

@property
def ramble_name:
    # The ramble application associated with this benchmark
```

moreover, while I think it will be important to be able to customize these, I think it would be good to extrapolate both from the experiment name by default (from looking through all the experiments, this seems to be true for all of them).

That would create a risk that for a new experiment, a user would forget to update when these values should differ (although fwiw, I think a benchpark audit could catch issues like that).

lib/benchpark/cmd/info.py
```diff
@@ -0,0 +1,175 @@
...
+def info_experiment(args):
+    def _info_url(experiment_class):
+        gen_header("URL")
+        print(indent() + experiment_class.url)
+
+    experiment_spec = benchpark.spec.ExperimentSpec(" ".join(args.spec))
+    experiment_class = experiment_spec.experiment_class
+
+    if args.spack:
+        subprocess.run(["spack", "info", experiment_class.spack_name])
```
scheibelp: (no change needed) this seems like a good way to delegate

scheibelp (commented):
Overall approach looks good (including where it delegates to Spack/Ramble). I have a few implementation suggestions

lib/benchpark/directives.py
michaelmckinsey1: `system_site` made an attribute in [001f1d3](https://github.com/LLNL/benchpark/pull/587/commits/001f1d32c653227b22cf72294ad6d7c2889bf3be).

lib/benchpark/cmd/info.py
michaelmckinsey1: [2f1bbbc](https://github.com/LLNL/benchpark/pull/587/commits/2f1bbbca2f0910aa637d971f4776ca6214e1aad5)

experiments/amg2023/experiment.py
michaelmckinsey1: [c95a8e0](https://github.com/LLNL/benchpark/pull/587/commits/c95a8e0571f67d84a9d062b8869977cc5973fa24). Correct, there are no cases where they differ right now (i.e. benchpark name = spack name = ramble name). If there is a case in the future, we can set it with `spack_name` or ramble_name` and it will be set correctly. The default name for both now is the experiment name in benchpark.

experiments/stream/experiment.py
michaelmckinsey1: [df9fd7a](https://github.com/LLNL/benchpark/pull/587/commits/df9fd7a087a715f90d7dacb9ac68f777e257c9e5) fixes this. I added myself as the `raja-perf` maintainer as an example and left the other experiments unset.

lib/benchpark/cmd/info.py
michaelmckinsey1: [df9fd7a](https://github.com/LLNL/benchpark/pull/587/commits/df9fd7a087a715f90d7dacb9ac68f777e257c9e5) This is done now for the other attributes, but for maintainers, since it can be multiple now I did a string join. If `maintainers()` is not set it is just an empty string.

lib/benchpark/directives.py
michaelmckinsey1: Updated

lib/benchpark/cmd/info.py
```diff
@@ -0,0 +1,195 @@
...
+        return
+    elif args.ramble:
+        subprocess.run(["ramble", "info", experiment_class.ramble_name])
+        return
+    else:
+        actions = {
+            "maintainers": (info_maintainers, [experiment_class]),
+            "ramble_name": (_info_ramble_name, [experiment_class]),
+            "spack_name": (_info_spack_name, [experiment_class]),
+            "url": (_info_url, [experiment_class]),
```
pearce8: We should not be keeping a URL in experiment.py since package.py already keeps a git URL (and in some cases the homepage).  We do need to be able to query the git URL, and the version("develop", branch="caliper"), particularly so we can determine the versions we are using, and where we are using specific branches.

systems/llnl-sierra/system.py
```diff
@@ -5,11 +5,37 @@
...
+            "sys_cores_per_node": 44,
+            "sys_gpus_per_node": 4,
+            "hardware_key": str(hardware_descriptions)
+            + "/IBM-power9-V100-Infiniband/hardware_description.yaml",
+        },
+    }
+
+    system_site = "llnl"
+
+    variant(
```
pearce8: Lassen and Sierra are identical, I am not convinced these should be considered variants.

systems/riken-fugaku/system.py
```diff
@@ -5,11 +5,29 @@
...
+            "sys_cores_per_node": 48,
+            "sys_mem_per_node": 32,
+            "hardware_key": str(hardware_descriptions)
+            + "/Fujitsu-A64FX-TofuD/hardware_description.yaml",
+        },
+    }
+
+    system_site = "riken"
+
+    variant(
```
pearce8: why do we have a variant with a single possible value?

systems/riken-fugaku/system.py
```diff
@@ -5,11 +5,29 @@
...
+    id_to_resources = {
+        "fugaku": {
+            "sys_cores_per_node": 48,
+            "sys_mem_per_node": 32,
+            "hardware_key": str(hardware_descriptions)
+            + "/Fujitsu-A64FX-TofuD/hardware_description.yaml",
+        },
+    }
+
+    system_site = "riken"
```
pearce8: 1) system_site should be part of id_to_resources since we could have multiple sites using the same system.py (for example, Lumi and Frontier).  

2) Is the site now stored in system.py *and* hardware_description?  If yes, we should decide where the info should be, and not duplicate because we won't be able to maintain the duplication.

3) I would like to be able to query benchpark for all systems where system_site = llnl.

lib/benchpark/cmd/info.py
michaelmckinsey1: I can remove the urls.  [c407991](https://github.com/LLNL/benchpark/pull/587/commits/c4079919f21426a10be90e7d581b1c18eff5bed1)

I think the following satisfies that query - `benchpark info experiment --spack experiment_name`? e.g.:
image

systems/llnl-sierra/system.py
michaelmckinsey1: This is fixed in [c8e9fcf](https://github.com/LLNL/benchpark/pull/587/commits/c8e9fcfe4a1aa7525418ba797baeed6cb4656366). I still think for the `id_to_resources` and for setting attributes, `llnl-sierra` and `riken-fugaku` should be the same as the other experiments. This can be done without the clusters as a variant.

systems/riken-fugaku/system.py
michaelmckinsey1: Removed in [c8e9fcf](https://github.com/LLNL/benchpark/pull/587/commits/c8e9fcfe4a1aa7525418ba797baeed6cb4656366)

lib/benchpark/cmd/info.py
pearce8: Does ``benchpark info experiment --spack experiment_name`` look at the package.py in benchpark/repo first?

pearce8: I want to make sure we capture the info from ``benchpark/repo/laghos/package.py``::

    homepage = "https://github.com/wdhawkins/laghos"
    git = "https://github.com/wdhawkins/Laghos.git"
    version("develop", branch="caliper")

systems/riken-fugaku/system.py
michaelmckinsey1: 1. Done in [8111d0c](https://github.com/LLNL/benchpark/pull/587/commits/8111d0c53a7b80d80b71742445a050eef9c7b4db), but doesn't the structure of the system directories preclude multiple sites sharing the same system.py? The site location is prepended e.g. `riken-fugaku`, `llnl-cluster`, etc.

2. `system_site` does not exist as a key in any of the yamls, however the system site information does exist under the `top500-system-instances` key:
```
  top500-system-instances:
    - Frontier (ORNL)  <https://www.top500.org/system/180047>
    - Lumi     (CSC)   <https://www.top500.org/system/180048>
    - Tioga    (LLNL)  <https://www.top500.org/system/180052>
```
and it is embedded in some of the key names like `system_definition->systems_tested->csc-lumi`.

3. So far, for this `info` command, we are able to query for either one system by system name or one experiment by experiment name. To additionally query for `system_site`, I am thinking of something like `benchpark info system --system_site llnl`, which would result in:
```
Systems:
    llnl-cluster
    llnl-elcapitan
    llnl-sierra
```
or do you mean
```
Systems:
    ruby
    dane
    lassen
    ...
```
Also, I need to understand the scope of this. Would we query for other attributes like `sys_cores_per_node`, and query over variants? Because at that point, I think it should be a separate command like `benchpark query` or `benchpark find`.

lib/benchpark/cmd/info.py
michaelmckinsey1: image

Currently it does, but only if you have ran `benchpark setup`, i.e. using the spack that is linked to `benchpark/repos/`, which is created on the setup. ~~I assume that this is not an acceptable constraint though, as it makes sense to use `benchpark info` before setting up any experiments.~~ Actually, `benchpark info experiment --spack` won't work at all until `benchpark setup` for the same reason.

systems/riken-fugaku/system.py
pearce8: 2. Maybe it is time to revisit this format::

```
  top500-system-instances:
     Frontier:
          benchpark_system:
          top500:  <https://www.top500.org/system/180047>
     Lumi:
          benchpark_system: csc-lumi
          top500: <https://www.top500.org/system/180048>
     Tioga: 
          benchpark_system: llnl-elcapitan
          top500: <https://www.top500.org/system/180052>
```

Note that I have removed the system_site from the yaml here, because if I point at a system.py it should be stored there (and we don't want to repeat), and if there is not a system.py in benchpark then the top500 link has enough info.  Hopefully that makes sense?

So, if I am interested in running on Frontier, it would be good to be able to find the yaml file where it is listed, and a system.py where it is listed (the latter may mean we need to add system_name to system.py).

If I found that Frontier is in HPE_..... file, but there is not a system.py for it, I would want to know if there are other systems in the HPE_.... file that do have a system.py so I can take a look at them.  Thoughts on how to implement that?

pearce8: 3. I think this is good::

```
Systems:
    llnl-cluster
    llnl-elcapitan
    llnl-sierra
```

pearce8: 4.  I think we can leave ``sys_cores_per_node`` for later, no need to worry about that now.

It would be good to be able to query for specific hardware - what systems do we have with MI250X?  But, if that is complicated, please create a follow-on issue and we can take a stab at it later.

pearce8: 1. csc-lumi is just a name, there is no reason it could not be used on Frontier if the software stack is indeed the same.

pearce8: Over the weekend, I merged several systems August specified last week.  Please grab the latest develop and update those systems as well.

michaelmckinsey1: 2. [454a4ab](https://github.com/LLNL/benchpark/pull/587/commits/454a4ab88f2ea0129e8a1d9243c46797e047397b) and [bcb531f](https://github.com/LLNL/benchpark/pull/587/commits/bcb531f505f4efe2c6c5baa47594478e8ddb2608) modify the yaml structures to remove redundant information and match the above format.

michaelmckinsey1:
> 4. I think we can leave `sys_cores_per_node` for later, no need to worry about that now.
> 
> 
> It would be good to be able to query for specific hardware - what systems do we have with MI250X? But, if that is complicated, please create a follow-on issue and we can take a stab at it later.

[e64f64a](https://github.com/LLNL/benchpark/pull/587/commits/e64f64a65889dc71a2793bad8de8b5c948e210fa) Adds querying (see examples at the bottom of PR description), however this is limited to the information in the `id_to_resources`. I created https://github.com/LLNL/benchpark/issues/641 to track the extension of this feature.

systems/riken-fugaku/system.py
michaelmckinsey1: > 2. Maybe it is time to revisit this format::
> 
> 
> ```
>   top500-system-instances:
>      Frontier:
>           benchpark_system:
>           top500:  <https://www.top500.org/system/180047>
>      Lumi:
>           benchpark_system: csc-lumi
>           top500: <https://www.top500.org/system/180048>
>      Tioga: 
>           benchpark_system: llnl-elcapitan
>           top500: <https://www.top500.org/system/180052>
> ```
> 
> Note that I have removed the system_site from the yaml here, because if I point at a system.py it should be stored there (and we don't want to repeat), and if there is not a system.py in benchpark then the top500 link has enough info. Hopefully that makes sense?
> 
> So, if I am interested in running on Frontier, it would be good to be able to find the yaml file where it is listed, and a system.py where it is listed (the latter may mean we need to add system_name to system.py).
> 
> If I found that Frontier is in HPE_..... file, but there is not a system.py for it, I would want to know if there are other systems in the HPE_.... file that do have a system.py so I can take a look at them. Thoughts on how to implement that?

Whole list of HPE: `Tuolumne`, `rzAdams`, `Frontier`.

Others: `Summit` & `Grace`.

After the refactor of the yamls, we can now check if `benchpark_system` is defined under the `top500-system-instances` and that would tell us if the `system.py` is not defined (or not linked)

pearce8 (approved):

PR #591, opened 2025-02-05
[DO NOT MERGE] Kripke comm dynamic

gracenansamba:
Added "https://github.com/gracenansamba/Kripke.git" path to /repo/kripke/package.py and adjusted other changes comparing to my old branch kripke_comm_cali

The current error is kripke@develop required in spack concretization.

May need to make a variant but was not successful.

added these lines for the new caliper metrics.
https://github.com/LLNL/benchpark/blob/Kripke_Comm_dynamic/modifiers/caliper/modifier.py#L45C9-L45C10: 
https://github.com/LLNL/benchpark/blob/Kripke_Comm_dynamic/modifiers/caliper-mpi/modifier.py#L27


dyokelson:
@gracenansamba ok so in the experiment.py for kripke there is the variant code I show below, this is setting you to the develop version when you run experiment init, so you need to pass in `version=comm_cali` during the `experiment init` command to grab that version in from the package.py.
```
    variant(
        "version",
        default="develop",
        description="app version",
    )
```

You should be able to do this as is, or you can update this code to look like Caliper's where you can list explicit values (you would just add comm_cali as a value):
```
    variant(
        "caliper",
        default="none",
        values=(
            "none",
            "time",
            "mpi",
            "cuda",
            "topdown-counters-all",
            "topdown-counters-toplevel",
            "topdown-all",
            "topdown-toplevel",
        ),
        multi=True,
        description="caliper mode",
    )
```

PR #593, opened 2025-02-05
Fix versioning logic for llnl_elcapitan

rfhaque:
## Description
This PR fixes the versioning scheme for the llnl_elcapitan system config. This change allows seamlessly creating elcap system instances for different rocm versions.
- [x] Complete the checklist for a relevant section(s) below
- [x] Delete sections below that are not relevant to this PR

### DIFF
--- a/systems/llnl-elcapitan/system.py
+++ b/systems/llnl-elcapitan/system.py
@@ -4,6 +4,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import pathlib
+from packaging.version import Version
 
 from benchpark.directives import variant
 from benchpark.system import System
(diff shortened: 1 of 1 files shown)


systems/llnl-elcapitan/system.py
```diff
@@ -69,6 +70,29 @@ class LlnlElcapitan(System):
...
+            self.short_cce_version = (
+                f"{self.cce_version.major}.{self.cce_version.minor}"
+            )
+        if self.rocm_version >= Version("6.0.0"):
+            self.pmi_version = Version("6.1.15.6")
+            self.llvm_version = Version("18.0.1")
+        else:
+            self.pmi_version = Version("6.1.12")
+            self.llvm_version = Version("16.0.0")
+        # TODO: Replace this with lookups into the working set
```
scheibelp, line 94: I'm not sure what is meant by "working set" here but the logic in this section looks really good to me: we could later offer variants to control each of these versions individually, but for now this essentially handles the recommended combinations of compiler/rocm/utilities.

scheibelp (approved):

systems/llnl-elcapitan/system.py
rfhaque: @scheibelp The "working set" is a - yet to be implemented - generalized set of all version dependencies for a given system. That would completely get rid of hard-coded versions from `system.py`. Still working on the details. I'll go ahead and merge this PR.

Issue #592, opened 2025-02-05
Benchpark version clones spack


pearce8:
**Describe the bug**
```benchpark version``` clones spack

**To Reproduce**
1. Go to ['https://software.llnl.gov/benchpark/getting-started.html'](https://software.llnl.gov/benchpark/getting-started.html)
2. Follow the directions we give users
3. Notice the spack cloning when you type ```benchpark version```

**Expected behavior**
Are we subsequently using the cloned spack?  If not, should we be having instructions here - or move the version check to after a workspace is created?

**Supercomputer (please complete the following information):**
 - system: any

michaelmckinsey1:
In `main.py`, `import benchpark.cmd.audit` calls `bootstrapper.bootstrap()`, which is what installs ramble and spack if they do not already exist in the path (i.e. will happen for fresh install).

One way I can think of to address this is to check before the import and exit if using `benchpark -V` or `benchpark --version`, like here https://github.com/LLNL/benchpark/pull/595

PR #595, opened 2025-02-06
Print --version and --help Without Bootstrapping

michaelmckinsey1:
## Description

Addresses https://github.com/LLNL/benchpark/issues/592

- [x] check for `-V` or `--version`, print and exit
- [x] check for `-h` or `--help`, print and exit

## Adding/modifying core functionality

- [x] Modify `lib/main.py` file

### DIFF
--- a/lib/main.py
+++ b/lib/main.py
@@ -14,6 +14,11 @@ import subprocess
 import sys
 import yaml
 
+__version__ = "0.1.0"
+if "-V" in sys.argv or "--version" in sys.argv:
+    print(__version__)
+    exit()
+
 import benchpark.cmd.audit
 import benchpark.cmd.system
 import benchpark.cmd.experiment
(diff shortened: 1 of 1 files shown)


michaelmckinsey1:
Will do something similar for `--help` if this is the agreed-upon solution

lib/main.py
```diff
@@ -14,6 +14,31 @@
...
+
+Subcommands:
+  {list,tags,system,experiment,setup,unit-test,audit}
+    list                List available experiments, systems, and modifiers
+    tags                Tags in Benchpark experiments
+    system              Initialize a system config
+    experiment          Interact with experiments
+    setup               Set up an experiment and prepare it to build/run
+    unit-test           Run benchpark unit tests
+    audit               Look for problems in System/Experiment repos"""
```
michaelmckinsey1: The downside is this would need to be updated manually when a new command is added

pearce8 (changes requested):
Lint fails

pearce8 (approved):

PR #616, opened 2025-02-13
Run Pre-built Binary

### LINKED ISSUE #609: How to run a pre-built binary
This is how you can do this in Ramble (per discussion with Doug Feb 11, 2025)

1) Don't use any package manager at all
2) Things that were built using eessi on the system: [tutorial](https://ramble.readthedocs.io/en/latest/tutorials/EESSI_package_manager.html)
3) Bash script that does all of the module loads, change eessi to environment-modules (we are trying to get away from modules, so this is probably not the place to start)

Could IMD, OMB, stream
- https://github.com/GoogleCloudPlatform/ramble/tree/develop/var/ramble/repos/builtin/applications/intel-mpi-benchmarks
- https://github.com/GoogleCloudPlatform/ramble/tree/develop/var/ramble/repos/builtin/applications/osu-micro-benchmarks
- https://github.com/GoogleCloudPlatform/ramble/tree/develop/var/ramble/repos/builtin/applications/streamc

michaelmckinsey1:
## Description

resolves https://github.com/LLNL/benchpark/issues/609
This PR enables running an experiment using an existing binary, i.e. not requiring building with spack. The `environment-modules` package manager is used instead, for ramble to load necessary modules for the binary. Includes a refactor of the `compute_spack_section()` function, into a generalized `compute_package_section()`, such that other package managers can share the same functionalities.

- Test benchmark binaries:
   - [x] osu-micro-benchmarks
   - [x] stream
- [x] Do not clone spack if the package manager is not spack
- [x] Add CI
- [x] Add docs

- [x] Modify `lib/benchpark/experiment.py` to enable `environment-modules` package manager and appending the binary location to `PATH`.
- [x] Modify `lib/benchpark/cmd/setup.py` to not setup spack if package manager is not spack.
- [x] Modify `experiments/*.py` to refactor `compute_package_section()` function.  

## Example usage
Expected binary `osu_latency` in provided external path `/usr/workspace/wsa/mckinsey/osu-micro-benchmarks/mpi/pt2pt`:
```
benchpark experiment init --dest=osumb osu-micro-benchmarks package_manager="environment-modules" workload="osu_latency" append_path="/usr/workspace/wsa/mckinsey/bin/osu-micro-benchmarks/mpi/pt2pt"
benchpark setup ./osumb/ ./ruby-system/ osumb-ruby/
```
results in new clause in experiment `ramble.yaml`:
```
env_vars:
    append:
    - paths:
        PATH: /usr/workspace/wsa/mckinsey/osu-micro-benchmarks/mpi/pt2pt
    set: {}
```

`execute_experiment` will now include:
```
export PATH="${PATH}:/usr/workspace/wsa/mckinsey/bin/osu-micro-benchmarks/mpi/pt2pt";
. /usr/WS1/mckinsey/benchpark/osumb-ruby/osumb/LlnlCluster-9751dbe/workspace/software/environment-modules/osu-micro-benchmarks/module_loads
srun -N 2 osu_latency ...
```


michaelmckinsey1:
Ignore this comment. Current version requires no changes to ramble.

~This currently requires a change to ramble that I don't know how to avoid (to prepend the path in `execute_experiment`).~
```
+++ b/var/ramble/repos/builtin/package_managers/environment-modules/package_manager.py
@@ -95,6 +95,14 @@ class EnvironmentModules(PackageManagerBase):
         with open(module_file_path, "w+") as f:
             f.write(loads_content)
 
+    register_builtin("add_to_path")
+
+    def add_to_path(self):
+        paths = []
+        for p in self.app_inst.variables["binary_path"]:
+            paths.append("PATH="+p+":"+"$PATH")
+        return paths
+
     register_builtin("module_load", required=True)
 
     def module_load(self):
```

pearce8:
@scheibelp @becker33 Please assess the design.  Also please suggest how we can prepend the path in execute_experiment - or help formulate a question for Doug.

lib/benchpark/experiment.py
```diff
@@ -47,6 +52,19 @@ def get_helper_name_prefix(self):
...
+    def append_environment_variable(self, name, value):
+        self.env_vars["append"][name] = value
+
+    def compute_config_variables(self):
+        pass
+
+    def compute_config_variables_wrapper(self):
+        self.compute_config_variables()
+        return self.variables, self.env_vars
+
```
michaelmckinsey1: Environment variable implementation from https://github.com/LLNL/benchpark/pull/618/

pearce8: @michaelmckinsey1 please explain what you mean here, and in any other comments/commit messages.  It is impossible to review otherwise.

michaelmckinsey1: Comments added. From a discussion, the functionality Riyaz wrote in #618 can be used here to append to `PATH` in the jobfiles that Ramble generates. The `append_environment_variable` function generates the expected output in the `ramble.yaml` (https://ramble.readthedocs.io/en/latest/workspace_config.html#environment-variable-control)

michaelmckinsey1:
> @scheibelp @becker33 Please assess the design. ~Also please suggest how we can prepend the path in execute_experiment - or help formulate a question for Doug.~

The path part has been figured out. We are using an existing ramble interface to append to the path.

experiments/quicksilver/experiment.py
```diff
@@ -59,9 +59,10 @@ def compute_spack_section(self):
...
+        if pkg_manager == "spack":
+            # empty package_specs value implies external package
+            self.add_spack_spec(system_specs["mpi"])
 
-        self.add_spack_spec(
-            self.name, [f"quicksilver@{app_version} +mpi", system_specs["compiler"]]
-        )
+            self.add_spack_spec(
+                self.name, [f"quicksilver@{app_version} +mpi", system_specs["compiler"]]
+            )
```
scheibelp: It looks like all of these individual subclasses of `Experiment` could be left alone, and you could skip `compute_spack_section` in `compute_package_section_wrapper` if `package_manager` is not spack. Am I missing something? It looks like all these functions are turned effectively into noops when the package manager is not spack, so it seems like it would make sense to skip the function altogether.

pearce8 (changes requested):
1. Please add a test of a binary in dry runs.  Maybe build/run saxpy as is now, then run that binary?
2. Please add a docs page showing your example.  Maybe call the page "Running a binary using Benchpark" and put it last under the "Running Benchpark" section.

experiments/quicksilver/experiment.py
michaelmckinsey1: Good point. Upon checking, the `environment-modules` manager in ramble actually does need this information, to create the file where modules are loaded. I removed the if-statements in the subclasses [313ff73](https://github.com/LLNL/benchpark/pull/616/commits/313ff7364f7b3041aacd8398b380c7ea28b25d4f). Having the `package_specs` be empty for non-spack managers was a mistake.

michaelmckinsey1:
> 1. Please add a test of a binary in dry runs.  Maybe build/run saxpy as is now, then run that binary?
> 
>     2. Please add a docs page showing your example.  Maybe call the page "Running a binary using Benchpark" and put it last under the "Running Benchpark" section.

1. Done in [efa22ab](https://github.com/LLNL/benchpark/pull/616/commits/efa22abcb38587a8bea11ec3089658e34b971f14)
2. [13e7d0c](https://github.com/LLNL/benchpark/pull/616/commits/13e7d0c2f4014f15cab19ccae29b1e7d2da06d6f)

pearce8 (approved):

Issue #551, opened 2025-01-17
Add documentation on experiment dependencies, revisit existing experiments to clean up


pearce8:
## How to specify dependencies for an experiment that should be provided by the system?

1.  The experiments should be using compiler and libraries specified by the system.
2. How should we specify that the experiment will need a blas?  If the system doesn't provide one, Spack will build it.
3. If a system provides multiple blas, it will need a variant for blas with a default.
4. If we want to use intel for blas, and rocm for lapack, how will this be specified?  (@rfhaque please find an example like this).  Can this be fixed by specifying in Spack that lapack maps to rocm, and blas maps to intel?
5. What handles does Ramble require to be in the experiment.py?  Try removing:

-      system_specs["lapack"] = "lapack"
-      system_specs["blas"] = "blas"
-      system_specs["mpi"] = "default-mpi"
-      system_specs["compiler"] = "default-compiler"

We do need to know how to specify:

1. Mandatory for spack to use (e.g., compiler, mpi)
2. Preferable for spack to use (math libraries)
3. Spack can build the rest

## What dependencies need to be built for an experiment?
1. Benchpark experiment init, Benchpark create a workspace
2. `spack -e path/to/environment/inside/the/workspace find`
3. `spack -e path/to/environment/inside/the/workspace python` for programmatic access (import spack.environment and call spack.environment.acitve_environment() to get the object)

## Which libraries are provided by a system?
1. Benchpark system init
2. Look at the packages.yaml
3. `spack -e path/to/environment/inside/the/workspace config get packages` spits out the yaml (not a good way to look at this programmatically...)

@becker33 please help plan this and explain better so @michaelmckinsey1 can take this on.

pearce8:
Superseded by #655.

PR #617, opened 2025-02-13
Standardize blas lapack handles

rfhaque:
This PR standardizes the names of the blas and lapack handles for all application `experiment.py` files tp 'default-blas` and 'default-lapack`. It also adds these handles to each `system.py` if needed.


pearce8:
@becker33 I need you to weigh in on how we should be setting this up.  This PR touches on a portion of the discussion we had in https://github.com/LLNL/benchpark/issues/551.

pearce8:
Superseded by #655.

PR #603, opened 2025-02-11
Avoid using CALI_CONFIG for RAJAPerf

michaelmckinsey1:
## Description

Corresponds to change in [LLNL/RAJAPerf#501 (See explanation)](https://github.com/LLNL/RAJAPerf/pull/501)

- [x] Modifies `modifiers/caliper/modifier.py` the Caliper modifier to disable `CALI_CONFIG` being generated for the `raja-perf` application.
- [x] Modifies `repo/raja-perf/application.py` to pass variables directly to the Caliper config manager in `raja-perf`.
- [x] Modifies `repo/raja-perf/package.py` to add corresponding branch of RAJAPerf (Can be removed after merge in RAJAPerf develop)

PR #618, opened 2025-02-14
[ARCHIVED] New caliper modifier implementation

rfhaque:
Work in progress. This PR changes the caliper modifier implementation (Solves #603)
In the existing caliper modifier, we define caliper modes to collect specific metrics (e.g. cuda, mpi, time etc)
The current benchpark implementation sets these modes at experiment init time. Each of these modes creates an environment variable in the experiment bash script to collect the desired metrics. However, this approach does not facilitate benchmark-specific changes as the caliper modifier is processed during the Ramble setup phase.

In the new impementation, we use the caliper modes specified on experiment init to directly set the required variables in the experiment ramble.yaml. The generated ramble.yaml is functionally equivalent to the one generated by the existing caliper modifier. However, this approach allows for changes to the caliper config at benchmark init time.

- [x] Replace with: A short description of the change, including motivation and context.

### DIFF
--- a/experiments/raja-perf/experiment.py
+++ b/experiments/raja-perf/experiment.py
@@ -35,7 +35,7 @@ class RajaPerf(
 
     def compute_applications_section(self):
         if not self.spec.satisfies("caliper=none"):
-          Caliper.Helper.unset_environment_variable(self, "CALI_CONFIG")
+            Caliper.Helper.unset_environment_variable(self, "CALI_CONFIG")
 
         n_resources = {"n_ranks": 1}
 
--- a/experiments/raja-perf/experiment.py
+++ b/experiments/raja-perf/experiment.py
@@ -29,10 +29,13 @@ class RajaPerf(
     variant(
         "version",
         default="develop",
+        values=("develop", "fixcaliconfig"),
         description="app version",
     )
 
     def compute_applications_section(self):
+        if not self.spec.satisfies("caliper=none"):
+          Caliper.Helper.unset_environment_variable(self, "CALI_CONFIG")
 
         n_resources = {"n_ranks": 1}
 
--- a/lib/benchpark/caliper.py
+++ b/lib/benchpark/caliper.py
@@ -27,18 +27,28 @@ class Caliper:
     )
 
     class Helper(ExperimentHelper):
+        caliper_modes = {
+          "time": "time.exclusive", # Platform-independent collection of time (default mode)
+          "mpi": "profile.mpi", # Profile MPI functions
+          "cuda": "profile.cuda", # Profile CUDA API functions
+          "topdown-counters-all": "topdown-counters.all", # Raw counter values for Intel top-down analysis (all levels)
+          "topdown-counters-toplevel": "topdown-counters.toplevel", # Raw counter values for Intel top-down analysis (top level)
+          "topdown-all": "topdown.all", # Top-down analysis for Intel CPUs (all levels)
+          "topdown-toplevel": "topdown.toplevel", # Top-down analysis for Intel CPUs (top level)
+        }
+
         def compute_modifiers_section(self):
             modifier_list = []
             if not self.spec.satisfies("caliper=none"):
-                for var in list(self.spec.variants["caliper"]):
-                    if var != "time":
-                        caliper_modifier_modes = {}
-                        caliper_modifier_modes["name"] = "caliper"
-                        caliper_modifier_modes["mode"] = var
-                        modifier_list.append(caliper_modifier_modes)
-                # Add time as the last mode
-                modifier_list.append({"name": "caliper", "mode": "time"})
+                modifier_list.append({"name": "caliper"})
             return modifier_list
+
+        def compute_config_variables(self):
+            if not self.spec.satisfies("caliper=none"):
+                modes = [Caliper.Helper.caliper_modes[m] for m in list(self.spec.variants["caliper"])]
+                self.add_experiment_variable("_cali_datafile", "{experiment_run_dir}/{experiment_name}.cali")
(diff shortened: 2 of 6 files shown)


experiments/raja-perf/experiment.py
```diff
@@ -29,10 +29,13 @@ class RajaPerf(
     variant(
         "version",
         default="develop",
+        values=("develop", "fixcaliconfig"),
         description="app version",
     )
 
     def compute_applications_section(self):
+        if not self.spec.satisfies("caliper=none"):
+            Caliper.Helper.unset_environment_variable(self, "CALI_CONFIG")
```
michaelmckinsey1, line 38: The changes here and in `repo/raja-perf/application.py` are no longer strictly necessary due to the updates to https://github.com/LLNL/RAJAPerf/pull/501 (It now works with benchpark develop). However the changes here would make it such that we avoid the warnings that are shown in that PR, by not providing the output file to the spot config. So I think this is still useful.

It would still be nice to omit setting CALI_CONFIG, to begin with. As the way this works currently is potentially confusing to debug, if you miss the `unset CALI_CONFIG` part. For context, this is what is generated in `execute_experiment`:
```
export CALI_CONFIG="spot(output=/usr/WS1/mckinsey/benchpark/exp/rajaperf-fix-lassen/rajaperf-fix/LlnlSierra-0b88187/workspace/experiments/raja-perf/suite/raja-perf_suite_strong_scaling_cuda_caliper_time_1/raja-perf_suite_strong_scaling_cuda_caliper_time_1.cali,time.exclusive)";
export CALI_CONFIG_MODES=time.exclusive;
unset CALI_CONFIG;
```

rfhaque: > The changes here and in `repo/raja-perf/application.py` are no longer strictly necessary due to the updates to [LLNL/RAJAPerf#501](https://github.com/LLNL/RAJAPerf/pull/501) (It now works with benchpark develop). However the changes here would make it such that we avoid the warnings that are shown in that PR, by not providing the output file to the spot config. So I think this is still useful.
> 
> It would still be nice to omit setting CALI_CONFIG, to begin with. As the way this works currently is potentially confusing to debug, if you miss the `unset CALI_CONFIG` part. For context, this is what is generated in `execute_experiment`:
> 
> ```
> export CALI_CONFIG="spot(output=/usr/WS1/mckinsey/benchpark/exp/rajaperf-fix-lassen/rajaperf-fix/LlnlSierra-0b88187/workspace/experiments/raja-perf/suite/raja-perf_suite_strong_scaling_cuda_caliper_time_1/raja-perf_suite_strong_scaling_cuda_caliper_time_1.cali,time.exclusive)";
> export CALI_CONFIG_MODES=time.exclusive;
> unset CALI_CONFIG;
> ```

The `unset CALI_CONFIG` was just a quick hack for testing purpose. We can unset `CALI_CONFIG` during experiment setup itself, so that it does not show up in the final executable script.

lib/benchpark/caliper.py
```diff
@@ -27,19 +27,36 @@ class Caliper:
...
+                modes = [
+                    Caliper.Helper.caliper_modes[m]
+                    for m in list(self.spec.variants["caliper"])
+                ]
+                self.add_experiment_variable(
+                    "_cali_datafile", "{experiment_run_dir}/{experiment_name}.cali"
+                )
+                self.set_environment_variable("CALI_CONFIG_MODES", " ".join(modes))
+                self.set_environment_variable(
+                    "CALI_CONFIG", f"spot(output={{_cali_datafile}},{','.join(modes)})"
```
scheibelp, line 57: (question) I only see `_cali_datafile` read here: is there a reason not to just set `CALI_CONFIG:` directly? e.g.

`set...(CALI_CONFIG, "spot(output={{{experiment_run_dir}/{experiment_name}.cali}},{','.join(modes)})"` or something close to that (I figure if it can expand `_cali_datafile` then it could expand `experiment_run_dir`.

lib/benchpark/experiment.py
```diff
@@ -168,12 +191,25 @@ def compute_applications_section(self):
...
         self.zips = {}
         self.matrix = []
         self.excludes = []
 
+        for cls in self.helpers:
+            variables, env_vars = cls.compute_config_variables_wrapper()
+            self.variables |= variables
+            self.env_vars["set"] |= env_vars["set"]
+            self.env_vars["append"] |= env_vars["append"]
+            self.env_vars["prepend"] |= env_vars["prepend"]
```
scheibelp, line 210: (question) Do you know if the prepends should be ordered based on helpers?

* If helper foo inherits helper bar
* And `MyExperiment` inherits both
* Then foo will appear before bar in the mro (and therefore `.helpers`)
* So we'll prepend everything from foo, followed by bar

If there aren't dependencies between helpers (WRT prepends) then this won't be an issue.

repo/raja-perf/application.py
```diff
@@ -18,7 +18,13 @@ class RajaPerf(ExecutableApplication):
...
             'c++','raja','cuda','hip','openmp','sycl']
 
-    executable('run', 'raja-perf.exe', use_mpi=True)
+    executable(
+        "run",
+        "raja-perf.exe"
+        + " --add-to-spot-config $CALI_CONFIG_MODES"
+        + " --outdir {experiment_run_dir}",
+        use_mpi=True,
+    )
```
scheibelp, line 27: (question) are other experiments expected to work without changes?

scheibelp (commented):
Overall this looks good to me. I have no requested changes, but a few questions

pearce8 (changes requested):
@rfhaque Are you still wanting this PR merged?  If so, please address Peter's comments.

rfhaque:
Archived for future reference

PR #391, opened 2024-10-07
Fix legacy salmon and add experiment class for salmon

dyokelson:
## Description

Removing dependence on the Fujitsu compiler.

This PR works on ruby with legacy salmon, but needs testing on Fugaku. I had to remove `-stdin` from the `application.py`, and I'm concerned this is required by Fugaku but not other systems. Here are the steps to reproduce with this branch:
```
cd benchpark
. setup-env.sh
benchpark setup salmon/openmp RCCS-Fugaku-Fujitsu-A64FX-TofuD workspace
```

Then follow the steps in output

Stephanie TODO:
* (DONE) Create `experiment.py` and verify generated `ramble.yaml` with hand written one in this PR
* (DONE) Replace dry run with `dynamic cts ruby`
* (DONE) Add dry run for fugaku (new way)

With the new way, these changes are included in this PR:
* Update spack version
* Remove babelstream package in benchpark in favor of using more recently updated package in spack
* Adds `-fallow-argument-mismatch`  for gcc10 and later
* Creates benchpark salmon-tddft package and related patches with openmp variant

Generated `ramble.yaml`. It does not generate a section for `rt` workload.
```
ramble:
  applications:
    salmon-tddft:
      workloads:
        gs:
          experiments:
            salmon-tddft_gs_single_node_openmp_{n_nodes}:
              exclude: {}
              matrix: []
              variables:
                arch: OpenMP
                exercise: exercise_01_C2H2_gs
                experiment_setup: ''
                inp_file: C2H2_gs.inp
                n_nodes:
                - '1'
                n_ranks: '{processes_per_node} * {n_nodes}'
                omp_num_threads:
                - '12'
                processes_per_node:
                - '4'
              variants:
                package_manager: spack
              zips: {}
  config:
    deprecated: true
    spack_flags:
      concretize: -U -f
      install: --add --keep-stage
  include:
  - ./configs
  modifiers:
  - name: allocation
  - name: exit-code
  software:
    environments:
      salmon-tddft:
        packages:
        - salmon-tddft
    packages:
      salmon-tddft:
        compiler: default-compiler
        pkg_spec: salmon-tddft@master+openmp
```

Dependencies: None

Fixes issue(s): Part of #460 

## Type of Change

- { } Adding a system, benchmark, or experiment
- {x} Modifying an existing system, benchmark, or experiment
- { } Documentation update
- {x} Build/CI update
- { } Benchpark core functionality

## Checklist:

If adding/modifying a benchpark:
[...]

PR #619, opened 2025-02-14
Experiment/salmon

slabasan:
Depends on #703 to pass lint CI checks.

~~Blocked by:~~
- [x] ~~#623 -- debug errors when updating spack version~~ Remove `depends_on()` and no longer need a newer spack
- [x] ~~Babelstream package in benchpark vs updated in spack~~ Fixed in develop

## Description

- [x] Add salmon experiment
- [x] Replaces #391, moves branch to upstream so CI tests run

- [x] If modifying the source code of a benchmark: create, self-assign, and link here a follow up issue with a link to the PR in the benchmark repo.
- [x] Add/modify a dry run unit test in `.github/workflows/run.yml`


checkout-versions.yaml
```diff
@@ -5,4 +5,4 @@
 
 versions:
   ramble: f5970aa     # develop on 1/08/2025 (newer than 0.5.0 release)
-  spack: 0b3b49b      # develop-2024-06-28 (newer than 0.22.0 release)
+  spack: c6d4037      # v0.23.0 (2024-11-13)
```
pearce8: I pulled the spack update version into its own PR https://github.com/LLNL/benchpark/pull/623; lets test it with other benchmarks and merge that PR first.

pearce8:
@slabasan What was the reason to pull in a new version of Spack here?  I think it is causing some of the issues with your tests, so I pulled out the version update into its own PR #623 to try to figure that part out.

checkout-versions.yaml
slabasan: Yes you beat me to it. Needed a new spack version because it was looking for `Package 'c' not found.` when trying to install salmon.

repo/salmon-tddft/package.py
```diff
@@ -0,0 +1,110 @@
...
+
+    homepage = "https://salmon-tddft.jp"
+    url = "https://salmon-tddft.jp/download/SALMON-v.2.0.0.tar.gz"
+
+    license("BSD-3-Clause-Open-MPI")
+
+    version("2.0.0", sha256="c3bb80bc5d338cba21cd8f345acbf2f2d81ef75af069a0a0ddbdc0acf358456c")
+    version("1.2.1", sha256="a5045149e49abe9dd9edefe00cd1508a1323081bc3d034632176b728effdbaeb")
+
+    depends_on("c", type="build")  # generated
```
slabasan: Remove these lines, then don't need a newer spack

experiments/salmon-tddft/experiment.py
```diff
@@ -0,0 +1,55 @@
...
+    def compute_spack_section(self):
+        # get package version
+        app_version = self.spec.variants["version"][0]
+
+        system_specs = {}
+        system_specs["compiler"] = "default-compiler"
+
+        self.add_spack_spec(
+            self.name, [f"salmon-tddft@{app_version}", system_specs["compiler"]]
+        )
```
michaelmckinsey1:
```suggestion
        self.add_package_spec(
            self.name, [f"salmon-tddft@{app_version}"]
        )
```

michaelmckinsey1 (changes requested):

experiments/salmon-tddft/experiment.py
slabasan: Can you help me understand why we need to remove more of this code?

slabasan:
The lint CI check caught a mis-spelling of optimization in `repo/salmon_tddft/fjmpi.patch`. But is also thinking that `nd` is a mis-spelled `2nd`. This is a false positive error. Should lint CI be checking these files?

Update: I've updated our github actions to ignore applying codespell to patch files. These files seem out of scope for benchpark.

experiments/salmon-tddft/experiment.py
michaelmckinsey1: https://github.com/LLNL/benchpark/pull/655 Removed experiment handles. 655 was merged after this PR was opened so you would not have noticed changes to the other experiments until the recent rebase.

The only part we need now is the package spec

```
self.add_package_spec(
   self.name, [f"salmon-tddft@{app_version}"]
)

michaelmckinsey1 (dismissed):
Looks good once lint passing

slabasan:
This one is (finally) ready!

dyokelson (commented):
Do we want all the tests in the run.yml or just in the new format? Just not sure if lines 265-285 are still needed or it would be best to just keep the later versions.

We should have the Riken folks test this out on Fugaku as well.

pearce8 (approved):

PR #447, opened 2024-11-25
[Do Not Merge] Quicksilver experiment

august-knox:
## Description

Sharing quicksilver compiler experiment.
Setup script for dane included. To setup everything up, run setupDane.sh, then combine resulting scripts into one with createLargeDane.py. The resulting shell script will be named largeDane
Dependencies: FIXME:Add a list of any dependencies.

Fixes issue(s): FIXME:Add list of relevant issues.

## Type of Change

- { } Adding a system, benchmark, or experiment
- { } Modifying an existing system, benchmark, or experiment
- { } Documentation update
- { } Build/CI update
- { } Benchpark core functionality

## Checklist:

If adding/modifying a system:
- { } Create a new directory for the system and a new `system.py` file
- { } Add a new dry run unit test in `.github/workflows`
- { } System appears in System Specifications table in docs catalogue section

If adding/modifying a benchpark:
- { } Add a new `application.py` and (maybe) `package.py` under a new directory
      for this benchmark
- { } Configure an experiment
- { } Benchmark appears in Benchmarks and Experiments table in docs catalogue
      section

If adding/modifying a experiment:
- { } Extend `experiment.py` under existing directory for specific benchmark
- { } Define a single node and multi-node experiments

If adding/modifying core functionality:
- { } Update docs
- { } Update `.github/workflows` and `.gitlab/ci` unit tests (if needed)

PR #644, opened 2025-02-25
documentation for comparing binaries

### LINKED ISSUE #627: Workflow of comparing 2 versions of an experiment
We need docs on how to compare 2 experiments like @august-knox did in #447, including how to use spack-diff (or is it altdiff?)

august-knox:
## Description

Progress for documentation comparing two binaries. 
Currently using quicksilver for the example, but I've noticed that quicksilver in develop doesn't have a caliper variant. Open to changing the example or implementing my non-upstreamed version in a separate PR.

Another question: should this include how to compare two cali outputs in thicket?

- [x] Update docs


docs/generate-multiple-binaries.rst
```diff
@@ -0,0 +1,108 @@
...
+
+``benchpark system init --dest=ruby-intel  llnl-cluster cluster=ruby compiler=intel``
+
+Creating experiment ramble.yaml:
+---------------------------
+
+Create the experiment description
+Parameters could include: version, scaling, etc.
+If all experiment-based variables are the same, create one ramble.yaml. We are also enabling the caliper modifier, allowing us to track performance through annotations in the source code.
+
```
dyokelson: I would rework this paragraph a bit, since we're comparing the two compilers, we want to keep all experiment variables the same, so that we know we're only comparing compilers.

docs/generate-multiple-binaries.rst
```diff
@@ -0,0 +1,108 @@
...
+``benchpark experiment init --dest=quicksilver  quicksilver caliper=mpi +weak +openmp ~single_node``
```
dyokelson: I would explain each of the parameters in this command just like: 
In this command we are intializing a quicksilver experiment in the quicksilver directory, we are doing weak scaling with openMP, and measuring MPI metrics with Caliper. 

Or something like that

docs/generate-multiple-binaries.rst
```diff
@@ -0,0 +1,108 @@
...
+
+
+Note: to create multiple duplicate experiments, change the experiment name ({experiment1}, {experiment2}). 
+Functionality to combine runs into a single job script is in progress, but not currently available.
```
dyokelson: Can you explain this more? Why would they need to do this?

docs/generate-multiple-binaries.rst
```diff
@@ -0,0 +1,108 @@
...
+
+
+
+ 
+Verifying build details, differences between builds
+---------------------------------------------------
+
+``spack find -L quicksilver``
+
+Returns all unique builds of a benchmark, along with their associated hashes. We can use these hashes to independently double-check the details of each build.
```
dyokelson: Can you show example output here?

docs/generate-multiple-binaries.rst
```diff
@@ -0,0 +1,108 @@
...
+
+Benchpark offers two ways to double check that each binary has built according to the specifications:
+
+In this case, we can check the quicksilver spec, along with its dependencies by running spack spec for each binary
+
+``spack spec quicksilver/{hash}``
```
dyokelson: show example output here too

docs/generate-multiple-binaries.rst
```diff
@@ -0,0 +1,108 @@
...
+
+This can also be done in a single command by the altdiff command built into benchmark, highlighting all differences in red.
+
+``spack-python  lib/scripts/altdiff.py quicksilver/{hash1}  quicksilver/{hash2}``
```
dyokelson: example output here, then people know what to expect and you can explain what they are looking at a little

docs/generate-multiple-binaries.rst
```diff
@@ -0,0 +1,108 @@
...
+
+Running Experiments
+-------------------
+
+To run each binary on different nodes, run the following commands::
+
+  ramble -P -D workspace/quicksilver/ruby-gcc/workspace on
+  ramble -P -D workspace/quicksilver/ruby-intel/workspace on
+
+However, we can manually combine each ``execute_experiment`` file into a single script, allowing us to run both binaries on the same node
```
dyokelson: If you're able to generate a small version of your example for this that would be good, like that just shows some portion of your script...maybe think about that a bit more, how you might be able to show it without it being too much detail.

docs/generate-multiple-binaries.rst
```diff
@@ -0,0 +1,108 @@
...
+
+
+
```
dyokelson: I think we need a section here that is just comparing FOMs or runtimes (something they can do without caliper if they know what they want to look at for their benchmark)

docs/generate-multiple-binaries.rst
```diff
@@ -0,0 +1,108 @@
...
+Collecting Data with Caliper
+----------------------------
+
+When the caliper modifier is enabled, benchpark generates a .cali file within the experiments directory
+This .cali file can be read in by thicket, and used to graph performance data
+
+Need to include: example graph
```
dyokelson: In this section I think you can start with, "Enabling the Caliper modifier gives us a much more detailed picture about any performance differences, beyond looking at runtimes we can generate a profile to see which functions are taking the longest..." something along those lines.

Some good example output here without needing to get into a thicket description would be to just print the cali-query output? Or you can say "reading the caliper files into thicket we can visualize the call tree" and include those graphs. I think you can get away with saying, "for more details on how to use Caliper and Thicket refer to their docs" and link to their docs pages. But it's nice to advertise what is possible a little bit.

dyokelson (changes requested):

dyokelson:
@michaelmckinsey1 let's sync up on the overlap between this PR and yours and just move forward with one combined version for external facing diffs

examples/multiple_binaries/setupDane.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+
+compilers=("gcc12" "gccSpack")
+optParams=("O2")
+scaling=("weak")
+. setup-env.sh
+rm -rf daneTCE
+rm -rf quicksilvergcc*
+benchpark system init --dest=daneTCE rubyExp cluster=dane compiler=gcc
+benchpark system init --dest=daneSpack rubyExp cluster=dane compiler=gccSpack
```
michaelmckinsey1: ~what is `gccSpack` compiler variant? Some branch?~

examples/multiple_binaries/setupDane.sh
```diff
@@ -0,0 +1,39 @@
...
+benchpark experiment init --dest=quicksilvergcc12O2weak1 quicksilver experiment=weak caliper=mpi
```
michaelmckinsey1: caliper variant for quicksilver not in develop

examples/multiple_binaries/setupDane.sh
michaelmckinsey1: This would depend on https://github.com/LLNL/benchpark/pull/447, so removing

examples/multiple_binaries/setupDane.sh
michaelmckinsey1: Depends on https://github.com/LLNL/benchpark/pull/645

examples/multiple_binaries/setupDane.sh
```diff
@@ -0,0 +1,39 @@
...
+benchpark setup quicksilvergcc12O2weak1 daneTCE workspace
+cp config.yaml workspace/spack/etc/spack/defaults/config.yaml
```
michaelmckinsey1: `config.yaml`?

michaelmckinsey1: Probably https://github.com/LLNL/benchpark/pull/447/files#diff-d8d0422389f03d783e32e627250fe29834bd09c6361640d1ff00661dd6820034

examples/multiple_binaries/combine_execute.py
michaelmckinsey1, line 1: Does this really need to exist or can we just `ramble on` after buidling in `setupDane.sh`?

dyokelson (approved):
@pearce8 Stephanie and Michael and I walked through this together today, it's ready to merge.

pearce8 (approved):

PR #608, opened 2025-02-11
[ARCHIVED] Adding basic mpibind modifier

august-knox:
## Description

Adding a basic modifier for mpibind. Modes include: on, off, v, vv, greedy:0. Extra modes are available through mpibind, and can be added if necessary. 
So far, mpibind modifier only imported into saxpy and kripke experiment classes.

To build saxpy experiment with mpibind=off:
`benchpark experiment init --dest=saxpy saxpy +openmp mpibind=off` 

Extra (potential) tasks:

- [x] send mpibind output to its own  json file
- [ ] add extra usage modes? (gpu? smt?) (round-robin, one at a time)
- [ ] add some more manual support for setting mpibind configs?

PR #646, opened 2025-02-27
[ARCHIVED] Allocation implementation for mpibind modifier

august-knox:
## Description

Alternative implementation for mpibind(#608), moving all mpi-command editing to the allocation modifier, where this already takes place by default. Functionality remains the same. 
Example command:
`benchpark experiment init --dest=saxpy saxpy mpibind=v +openmp`

So far, modes include: off, on, v, vv, greedy:0

- [ ] Add/modify a dry run unit test in `.github/workflows/run.yml`
- [ ] Remove mention of mpibind modifier since it is part of allocation modifier in this implementation
- [ ] Test
- [ ] Add documentation: how to test
- [ ] @michaelmckinsey1 how to propagate info into Caliper files when Caliper modifier is on


experiments/amg2023/experiment.py
```diff
@@ -24,6 +25,7 @@ class Amg2023(
     WeakScaling,
     ThroughputScaling,
     Caliper,
+    Mpibind,
```
scheibelp, line 28: Does this class exist?

modifiers/allocation/modifier.py
```diff
@@ -230,6 +232,30 @@ class Allocation(BasicModifier):
...
+
+    mode(
+        name="vv",
+        description="Run mpibind in very verbose mode",
+    )
+    mode(
+        name="greedy:0",
+        description="Run mpibind in very greedy mode",
+    )
+    depends_on = ["mpibind"]
```
scheibelp, line 258: I don't think these should be passed as alternate modes for the allocation modifier: I'd prefer this was passed as something like an `mpibind` variable (with all these possible values)

modifiers/allocation/modifier.py
```diff
@@ -461,3 +491,43 @@ def determine_scheduler_instructions(self, v):
...
+                "vv": "verbose:2",
+                "on": "on",
+                "off": "off",
+                "greedy:0": "greedy:0",
+            }
+            mpi_end = flags.get(self._usage_mode)
+            return f"{base_string} {mpi_string}{mpi_end}"
+        else:
+            mpi_end = self._usage_mode
+            return f"{base_string} {mpi_string}{mpi_end}"
```
scheibelp, line 516: I think this is just adding an extra option onto the mpi command invocation.

I'd recommend modifying `_init_batch_and_cmd_opts` to include this extra option, vs doing string manipulation on the command.

modifiers/allocation/modifier.py
```diff
@@ -461,3 +491,43 @@ def determine_scheduler_instructions(self, v):
...
+
+    executable_modifier("mpibind")
+
+    def mpibind(self, executable_name, executable, app_inst=None):
+        pre_exec = []
+        post_exec = []
+        output_file = "{experiment_run_dir}/{experiment_name}.out"
```
scheibelp, line 523: (question) is `.out` the output captured by Ramble from running the experiment? And then `parse_mpibind_output.py` is searching for output from `mpibind` that is generated as part of running that experiment? I don't know how mpibind works: are these values determined by `mpibind` itself (i.e. not known before running the experiment)?

modifiers/allocation/parse_mpibind_output.py
```diff
@@ -0,0 +1,48 @@
...
+                cpus = match.group(4)  # CPUs value
+
+                task_map["task " + task] = {
+                    "cpus": cpus,
+                    "gpus": gpus,
+                    "nths": nths,
+                }
+
+    output_file = "mpibind_log_file.json"
+    with open(output_file, "w") as json_file:
```
scheibelp, line 39: This looks relative: are you sure it runs with the basedir you want? You might want to specify an abspath (e.g. as another argument to `parse_mpibind_output`.

scheibelp (commented):
I have a few suggestions and a question (let me know if they are unclear or you think another approach is better).

pearce8 (changes requested):
@rfhaque Are you still wanting to merge this? If yes, please update and mark ready for review.  If not, please close this PR.

pearce8:
@rfhaque @nhanford We need to pick this back up and decide what to do.

rfhaque:
Archived. To be followed up with a new PR linked to the issue #962

PR #656, opened 2025-03-03
System with multiple compilers

pearce8:
## Description
When some of the dependencies don't build with the default compiler, we use this syntax to provide an alternative compiler::

```
packages:
  all:
    require:
    - one_of: ["%cce", "%gcc"]  # this line makes spack use cce to build where possible, gcc otherwise
```

- [x] Depends on #658 
- [ ] What is this called in Spack, and can we link to where this is explained? 
- [ ] How to detect that Spack is building a secondary compiler?
- [ ] This should be explained in [how to add a system](docs/add-a-system-config.rst)
- [ ] Add a test to verify that only the special packages use the secondary compiler to be added to CI
- [ ] ``altdiff`` option to show which compiler was used to build which dependence


pearce8:
@scheibelp Please take this on.

pearce8:
Superseded with https://github.com/llnl/benchpark/pull/1183.

PR #667, opened 2025-03-12
[WIP] Re-use Bootstrapped Spack and Ramble

michaelmckinsey1:
## Description

- [x] Currently in `develop`, every time we set up an experiment in a new workspace (i.e. `benchpark setup` is called with a new workspace) we clone spack and ramble into the workspace directory. Because all of our benchmarks are using the same version of spack/ramble, this should be unecessary. This change only clones spack and ramble one time, to be used across workspaces.
- [x] (Done in #669) Add check along with if the directory already exists, to check if the spack/ramble version in root is different from `checkout-versions.yaml`, so we can update that spack/ramble.
- [x] Depends on #669 
- [x] Add git fetch origin before checkout

- [x] Update `.github/workflows/docs.yml` to support work-around.
- [x] Update `cmd/audit.py`, `cmd/setup.py`, `experiment.py`, `repo.py`, `spec.py`, and `system.py` to add benchpark root as source path.
- [x] Update `lib/benchpark/runtime.py` to change the way bootstrapping works.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -28,7 +28,7 @@ jobs:
         bin/benchpark system init --dest=generic-x86 generic-x86
         bin/benchpark setup saxpy generic-x86 /tmp/workspace
         # this is gross and we should better setup ramble for people or make it a Spack package
-        pip install -r /tmp/workspace/ramble/requirements.txt
+        pip install -r ~/.benchpark/ramble/requirements.txt
 
     - name: Build with sphinx
       run: |
--- a/.github/workflows/run.yml
+++ b/.github/workflows/run.yml
@@ -102,12 +102,12 @@ jobs:
           ./bin/benchpark experiment init --dest=saxpy/binary saxpy+openmp package_manager="environment-modules" append_path="$SAXPY_PATH"
           ./bin/benchpark setup saxpy/openmp generic-x86 workspace-binary/
 
-          ./workspace-binary/ramble/bin/ramble \
+          ramble \
             --disable-progress-bar \
             --workspace-dir \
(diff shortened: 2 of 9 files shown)


pearce8:
@michaelmckinsey1 are we still moving forward with this PR?  If yes, please resolve conflicts and request a re-review from Peter.

pearce8 (changes requested):
@michaelmckinsey1 are we still moving forward with this PR? If yes, please resolve conflicts and request a re-review from Peter.

michaelmckinsey1:
This is being closed in favor of https://github.com/LLNL/benchpark/pull/784. Reasons why we are not proceeding with this PR at this point:
1. _Using the benchpark bootstrap in home for building packages will be space constrained. Cloning spack in benchpark workspaces allows the user to choose where packages are built. Additionally users can simply remove the workspace to remove those packages and start clean with a new workspace._ 
   - To address this we would either need: (1) a mechanism to choose where the bootstrap exists, and have benchpark know that location even after the current environment is closed (probably write this path to home dir). (2) a mechanism to choose where spack installs packages. 
2. _Users might maintain two benchpark workspaces, using a different version of spack for each one._
   - Currently, benchpark will attempt to update the bootstrap ramble/spack if the `checkout-versions.yaml` is updated (upon any `benchpark experiment init`, `benchpark system init`, etc). How benchpark works now, the user could create a separate clone of benchpark for a newer version, and keep their old clone, and benchpark would switch the bootstrap versions accordingly. 
   - This change would require a spack that is capable of installing packages from multiple spack versions. Each new checkout of spack would need to coordinate additional cleanup steps to clear cached state about packages (that could become wrong when changing spack versions). We do not have anything within the benchpark/ramble workspace that associates it with the spack checkout, so that information would need to be manually tracked.

PR #676, opened 2025-03-17
spack v1.0 support: `%foo +bar` -> `+bar %foo`

haampie:
Spack v1.0 will parse `pkg %gcc +foo` as "pkg with direct dependency gcc
with variant foo enabled" instead of "pkg with variant foo enabled and compiler
gcc". Reorder as `pkg +foo %gcc` to be compatible with Spack v0.x and
v1.0.


pearce8:
@scheibelp As you review, please also make sure this works for system.py as we have it now.

- [ ] this for sure seems to impact Fugaku, and needs to be updated in riken-fugaku/system.py
- [ ] how does one find this pattern?

pearce8 (dismissed):
@scheibelp This has been here too long, probably want to reapply commits to new develop so we can merge it.

haampie:
Run

```
spack style --spec-strings <files...>
```

to regenerate

scheibelp (approved):

pearce8 (approved):

PR #680, opened 2025-03-18
AD: Require a Clang-based compiler in a package

pearce8:
## Enzyme is a Clang-based source code transformation tool.  

- [ ] experiments/ad/experiment.py says nothing about clang
- [ ] repo/ad/package.py says nothing about clang
- [ ] repo/enzyme/package.py says ``depends_on("llvm@18:")``.  How should we correctly specify in Enzyme's package.py that only a Clang-based compiler (so LLVM, intel oneapi, etc.) should be used here? 
- [ ] Add/modify a dry run unit test in `.github/workflows/run.yml` - forcing a use of a clang compiler on a system where the default is not clang (Do we have a system like that?)
- [ ] We need a test that should fail: test on a system where a clang compiler is *not* specified.  How to make sure Spack doesn't build Clang? 
- [ ] Is there a Spack docs page we should point at for this?
- [ ] Update Benchpark docs on how to specify a compiler/library in an experiment that experiment requires


pearce8:
Superseded by #782.

PR #658, opened 2025-03-04
Refactor Systems to Initialize Using Dictionaries

michaelmckinsey1:
## Refactor the way systems are initialized to be similar to how experiments are initialized, using dictionaries instead of multi-line strings and yaml files.

- [x] Addresses https://github.com/LLNL/benchpark/issues/654
- [x] Write a python utility to read in [file1](https://github.com/LLNL/benchpark/blob/develop/systems/aws-pcluster/compilers/gcc/00-gcc-7-compilers.yaml) and [file2](https://github.com/LLNL/benchpark/blob/develop/systems/aws-pcluster/externals/base/00-packages.yaml)  into a dictionary (does file1 get deprecated when we start using compilers as packages @becker33 ??)
   - [x] This is `yaml.safe_load(file)`. 
- [x] Adding/updating packages docs
   - [x] Write docs on how to generate a package list on your system, and convert to a dictionary for defining a new system
   - [x] Write docs on how to refresh a package list on your already defined system 
`benchpark system external llnl-cluster cluster=ruby`
image

- [x] How to handle multiple versions of available libraries?
   - [x] list-like
   ```"cmake": {
                    "externals": [
                        {"spec": "cmake@3.20.2", "prefix": "/usr"},
                        {"spec": "cmake@3.23.1", "prefix": "/usr/tce"},
                        {"spec": "cmake@3.24.2", "prefix": "/usr/tce"},
                    ],
- [x] Write python utility to write out a dictionary to yaml (does the experiment example write a temporary file in a workspace?)
   - [x] Not necessary as `yaml.dump` will do this. 
- [x] What is needed in bin to use it this way?  Look at the experiment as an example.
   - [x] Nothing additional is needed
- [x] Update all `system.py`'s which have yaml files stored
- [x] ~For packages stored as a string right now, move the string to a file first?~ Packages stored as strings converted to dictionaries in-place.
- [x] Create script to validate generated systems are the same as previous -- for the system using yaml files, this may be a yaml to dictionary comparison, which can also be useful for knowing when the package dict on the system needs to be updated.
     ![image](https://github.com/user-attachments/assets/d163a990-8739-46d3-86af-c4b57a3bbd9a)


## Adding/modifying a system (docs: [Adding a System](https://software.llnl.gov/benchpark/add-a-system-config.html))

Done (Migrate information in yamls `systems/x/*.yaml` to `system.py`)
- [x] llnl-cluster
- [x] aws-pcluster
- [x] csc-lumi
- [x] cscs-daint
[...]

PR #682, opened 2025-03-19
Fix System Definitions

michaelmckinsey1:
## Description

- [x] There was an issue with `lib/scripts/compareyaml.py`, which was reporting false-negatives, so bugs with the system configurations were not caught in #658. This PR fixes the script, which I've used to validate that the `system.py` definitions for the modified systems were broken, and applies the fixes.

### DIFF
--- a/lib/scripts/compareyaml.py
+++ b/lib/scripts/compareyaml.py
@@ -34,27 +34,27 @@ def compare_yaml(file1, file2):
     def normalize_types(data):
         """fix to compare strings to int '56' == 56 for yaml purposes"""
         if isinstance(data, dict):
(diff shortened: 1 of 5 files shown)


michaelmckinsey1:
@rfhaque @scheibelp This addresses the bug we were looking at today. Most notably, the lassen configuration was missing a bunch of package definitions due to a mistake in #658.

pearce8 (approved):

alecbcs:
I *think* I figured out what's going on here and submitted a fix in https://github.com/LLNL/benchpark/pull/684. Once that PR merges this should work in the merge queue.

PR #703, opened 2025-03-25
ci: add `all` check 

slabasan:
## Description

Add `all` check to allow for required checks that can be passed or skipped but not failed.


alecbcs:
I would add `*.patch` to the existing set of skips in the configuration file here instad of modifying the CI scripts. https://github.com/LLNL/benchpark/blob/57bf7aa74447bade1aad624ae3bffecd525b31cf/pyproject.toml#L60.

slabasan:
Fixed, thanks @alecbcs. That's a better place to change it.

pyproject.toml
```diff
@@ -57,5 +57,5 @@ per-file-ignores = """
 """
 
 [tool.codespell]
-skip = './docs/_build,./docs/_static,./taxonomy.yaml'
+skip = './docs/_build,./docs/_static,./taxonomy.yaml,./repo/*.patch'
```
alecbcs:
```suggestion
skip = './docs/_build,./docs/_static,./taxonomy.yaml,./repo/**/*.patch'
```

Does the previous single `*.patch` work locally for you for patches in packages? I'd think you'd have to do a double star to get it to recurse into the packages, but I could be wrong / they might be using a different glob library.

pearce8 (dismissed):

PR #648, opened 2025-02-27
Script to Compare Packages in Benchpark Against Spack Tag

michaelmckinsey1:
## Description

- [x] Add a new script to report whether packages in `benchpark/repo` are the same as `var/spack/repos/builtin/packages`.
- [x] Addresses https://github.com/LLNL/benchpark/issues/596
- [x] Complete the checklist for a relevant section(s) below
- [x] Delete sections below that are not relevant to this PR

Todo:
- [x] superceded by #695 -> ~We need a doc page for utilities in lib/scripts, perhaps call it "Benchpark Utils".~
- [x] Try running this script as CI for package x if `repo/x/package.py` is modified

## Adding/modifying core functionality, CI, or documentation:

- [x] Add `benchpark/lib/scripts/spackdiff.py`
- [x] Update `bin/benchpark-python`, `pyproject.toml`, and `lib/benchpark/cmd/unit_test.py` to make interactive python optional during execution of `bin/benchpark-python`. It should be disabled while using this script.
- [x] Rename `altdiff.py` to `diffSpecs.py` to match new script name.

## Examples

Note: terminal output is colorized
image

We can check against spack tags like `v0.23.1`, or commits like `0b3b49b`, which benchpark develop uses.
`python spackdiff.py --spack-tag="0b3b49b"`
```
Comparing benchpark packages to packages in spack 0b3b49b
Previous HEAD position was 99405e6a4d nlcglib: add v1.1.0 (#44580)
HEAD is now at 0b3b49b4e0 installer.py: handle external roots the same (#44917)
ad
    ad/package.py does not exist in spack
amg2023
     7 different lines
babelstream
     28 different lines
branson
     1 different lines
caliper
     95 different lines
cray-mpich
     67 different lines
cublas
    cublas/package.py does not exist in spack
cusolver
    cusolver/package.py does not exist in spack
enzyme
     33 different lines
essl
     34 different lines
genesis
     87 different lines
gromacs
     59 different lines
hipsycl
     58 different lines
hpcg
     55 different lines
hpl
     13 different lines
hypre
     208 different lines
kokkos
     207 different lines
kripke
     28 different lines
laghos
     22 different lines
lammps
     438 different lines
lapack-xl
    lapack-xl/package.py does not exist in spack
md-test
    md-test/package.py does not exist in spack
mfem
     604 different lines
osu-micro-benchmarks
     42 different lines
phloem
    phloem/package.py does not exist in spack
quicksilver
     9 different lines
qws
[...]

PR #707, opened 2025-03-31
adds --update-head-ok to git fetch to force update

slabasan:
## Description

- [x] Adds `--update-head-ok` to `git fetch` to force update. By default, git fetch refuses to update the head which corresponds to the current branch. This flag disables the check.

@michaelmckinsey1 I think this should fix CI on `develop` at `e069d61`. Could you take a look?


michaelmckinsey1 (approved):
This was caused by a change in #648. LGTM

pearce8 (approved):

PR #714, opened 2025-04-02
Scaling function implementation

### LINKED ISSUE #706: Experiments not working with default scaling
Ref: https://github.com/LLNL/benchpark/pull/701#discussion_r2013047812

Several experiments (hpl, kripke, remhos, etc.) check for multiple scaling options (single_node, strong, weak, throughput) and error if there are more than 1 specified. Since `+single_node` is silently added to every config, certain experiments require `~single_node` to be explicitly specified in order to perform other scaling experiments, e.g. `hpl+strong~single_node` works, `hpl+strong` doesn't. 

Disabling `single_node` to be on by default also breaks certain experiments, that expect `+single_node` to set certain experiment variables in the `experiment.py`.

1. Should the default behavior instead be to [error](https://github.com/LLNL/benchpark/pull/701#discussion_r2013054860) if the scaling type is not specified?
2. Do we then expect the following spec syntax for every experiment: `[benchmark]+[programming_model]+[scaling]`, e.g. `amg2023+openmp+strong`, otherwise `experiment.py` errors?

Proposed changes:
- Remove `single_node`
- Make scaling modes mutually exclusive

rfhaque:
See #723 

## Description
This PR provides a new implementation for strong, weak and throughput scaling for the experiments.
In this implementation an experiment that supports strong scaling, weak scaling or throughput scaling must implement, respectively, the methods `compute_strong_scaling_expr_config`, `compute_weak_scaling_expr_config` and `compute_throughput_scaling_expr_config`. Additionally, each experiment must provide a default experiment (not necessarily a single node experiment) by implementing the `compute_default_expr_config` method. Rather than scaling each variable individually, the new scaling implementation generates a per-dimension scaling vector that gets multiplied to each variable to be scaled e.g, the `ramble.yaml` for 4 instances of strong scaling on a variable `{'px' : 2, 'py' : 2, 'pz' : 1}` would look like
```
variables:
  sf0: [1, 1, 2, 2]
  sf1: [1, 1, 1, 2]
  sf2: [1, 2, 2, 2]
  px : '{sf0} * 2'
  py : '{sf1} * 2'
  pz : '{sf2} * 1'
```
For weak scaling, the same scaling vector is multiplied to multiple variables. For any experiment, only one of strong, weak or throughput scaling can be specified at one time. If no scaling is specified, the default experiment configuration is generated.

- [ ] Update docs


experiments/kripke/experiment.py
```diff
@@ -39,102 +39,72 @@ class Kripke(
...
-                f"Only one type of scaling per experiment is allowed for application package {self.name}"
-            )
-
-        input_variables = {
+    def initialize_expr_variables(self):
+        expr_vars = {
+            # Number of processes in each dimension
+            "num_procs": {"npx": 2, "npy": 2, "npz": 1},
+            # Number of zones in each dimension, per process
+            "problem_sizes": {"nzx": 64, "nzy": 64, "nzz": 32},
```
michaelmckinsey1, line 53: this would be `x`, `y`, `z` (`--zones <x>,<y>,<z>`), instead of `nzx`, `nzy`, `nzz`?

experiments/amg2023/experiment.py
```diff
@@ -54,84 +54,57 @@ class Amg2023(
...
     # )
 
-    def compute_applications_section(self):
-        # TODO: Replace with conflicts clause
-        scaling_modes = {
-            "strong": self.spec.satisfies("+strong"),
-            "weak": self.spec.satisfies("+weak"),
-            "throughput": self.spec.satisfies("+throughput"),
-            "single_node": self.spec.satisfies("+single_node"),
+    def initialize_expr_variables(self):
```
michaelmckinsey1, line 62: amg: local problem size - number of zones is number of zones per process
strong scaling: 2 procs 64 x 64 x 64 -> on 4 process is 64 x 64 x 32 -> on 8 is 64 x 32 x 32
weak scaling: always 64 x 64 x 64

experiments/kripke/experiment.py
```diff
@@ -39,102 +39,72 @@ class Kripke(
...
-        }
-
-        scaling_mode_enabled = [key for key, value in scaling_modes.items() if value]
-        if len(scaling_mode_enabled) != 1:
-            raise BenchparkError(
-                f"Only one type of scaling per experiment is allowed for application package {self.name}"
-            )
-
-        input_variables = {
+    def initialize_expr_variables(self):
```
michaelmckinsey1, line 48: kripke: global size
strong scaling: always 64 x 64 x 64
weak scaling: 2 procs 64 x 64 x 64 -> on 4 process is 64 x 64 x 128 -> on 8 is 64 x 128 x 128

experiments/amg2023/experiment.py
```diff
@@ -12,9 +12,14 @@
...
 from benchpark.scaling import ThroughputScaling
+from benchpark.domaindecomposition import PerProcessDomains
 from benchpark.caliper import Caliper
 
-
+@PerProcessDomains(
+strong_scaling_policy="conserveglobalproblemsize",
+weak_scaling_policy="default",
+throughput_scaling_policy="default",
+)
```
rfhaque, line 22: @pearce8 @scheibelp Implementation of per-process scaling policies `lib/benchpark/domaindecomposition.py::PerProcessDomains`

experiments/kripke/experiment.py
```diff
@@ -12,9 +12,15 @@
...
 from benchpark.scaling import ThroughputScaling
+from benchpark.domaindecomposition import GlobalDomains
 from benchpark.caliper import Caliper
 
 
+@GlobalDomains(
+strong_scaling_policy="default",
+weak_scaling_policy="default",
+throughput_scaling_policy="default",
+)
```
rfhaque, line 23: @pearce8 @scheibelp Implementation of global scaling policies lib/benchpark/domaindecomposition.py::GlobalDomains

lib/benchpark/domaindecomposition.py
```diff
@@ -0,0 +1,746 @@
...
+
+                    for i, k in enumerate(num_procs_keys):
+                        result["num_procs"][k].append(next_num_procs_list[i])
+                    for i, k in enumerate(prob_sizes_keys):
+                        result["problem_sizes"][k].append(next_prob_sizes_list[i])
+
+                return result
+        elif self.strong_scaling_policy == "custom":
+            if 'custom_strong_scaling_policy_impl' not in cls.__dict__:
+                raise TypeError(f"Class '{cls.__name__}' must define 'custom_strong_scaling_policy_impl(input_vars)' method when strong_scaling_policy = 'custom'")
```
scheibelp, line 97: I'm not seeing the benefit of making these decorators vs. inherited classes. Also I still see the experiments modified in this PR inheriting `StrongScaling` (if that is intended, then I don't know why this logic doesn't go there).

lib/benchpark/domaindecomposition.py
```diff
@@ -0,0 +1,746 @@
...
+
+                if not isinstance(num_procs, dict) and not isinstance(problem_sizes, dict):
+                    raise BenchparkError(f"Input vars must be a dictionary of type str->int")
+                else:
+                    num_dims = len(num_procs)
+
+                if len(problem_sizes) != num_dims:
+                    raise BenchparkError(
+                        f"problem_sizes dimensions {len(problem_sizes)} do not match num_procs dimensions {num_dims}"
+                    )
```
scheibelp, line 437: I'd recommend converting `input_vars` from a dict to a class that can auto-check this for `num_procs` and `problem_sizes` attributes.

experiments/amg2023/experiment.py
```diff
@@ -54,84 +59,55 @@ class Amg2023(
...
+    def compute_default_expr_config(self):
+        expr_input_vars = self.initialize_expr_variables()
+        variables = expr_input_vars["num_procs"] | expr_input_vars["problem_sizes"]
+        self.add_expr_variables(expr_input_vars, variables)
+
+    def add_expr_variables(self, expr_input_vars, variables):
+        for k, v in variables.items():
+            self.add_experiment_variable(k, v, True)
+
+        n_resources = " * ".join(f"{{{k}}}" for k in expr_input_vars["num_procs"])
```
scheibelp, line 100: This appears to have gone from passing along pre-computed numbers based on resolving variables into values, to passing along variable references with string-ified arithmetic to Ramble: what is the benefit of this? I would think even if the dimensions can change, that the resource requests could still be pre-computed.

This may cause issues with the allocation modifier (i.e. the one specifically distributed with Benchpark); IIRC some logic was added to Ramble in the last few months that would make the allocation modifier unnecessary, and it would probably also not have this issue.

lib/benchpark/domaindecomposition.py
```diff
@@ -0,0 +1,746 @@
...
+                    "num_procs" : {k: [v] for k, v in num_procs.items()},
+                }
+
+                for itr in range(num_exprs):
+                    idx = (start_dim + itr) % num_dims
+                    curr_num_procs_list = [result["num_procs"][k][-1] for k in num_procs_keys]
+                    next_num_procs_list = curr_num_procs_list[:]
+                    next_num_procs_list[idx] *= scaling_factor
+                    for i, k in enumerate(num_procs_keys):
+                        result["num_procs"][k].append(next_num_procs_list[i])
```
scheibelp, line 300: This seems a bit more involved than it needs to be:

* You pull the "latest" vector
* You scale one of its dimensions to produce a new vector
* You append that updated vector

I think this doesn't have to reinitialize the current vector by retrieving it (i.e. retrieving what was just appended).

Furthermore, this logic is repeated in at least one other spot (`apply_strong_scaling_policy`) and IMO it would be worth consolidating into a function like `produce_series_of_scaled_vectors_from_init_vector`.

lib/benchpark/domaindecomposition.py
rfhaque: @scheibelp The actual scaling logic depends on whether the input problem sizes are specified per-process or as a global problem size. We would need to extend StrongScaling (and WeakScaling) to differentiate between the two. Decorators allow injecting the required scaling logic directly without subclassing the scaling classes. That said, it would indeed be better to consolidate all scaling functionality into one construct.

experiments/amg2023/experiment.py
rfhaque: The idea was to compute the scaling vectors per dimension once and then use multiplication expressions to generate the other variables. However, this code is kind of obsolete in that we do not always have multiplicative scaling, e.g. strong scaling in the case of per-process input sizes like amg2023 require the problem size to the halved for every doubling of processor count to maintain the global problem size constant. TL;DR this function needs to be changed

pearce8:
Decided not to go with the decorators for this design.

PR #717, opened 2025-04-02
Starting provides, requires directives

### LINKED ISSUE #725: Use provides and requires directives
- [ ] Use ``provides`` directive in ``benchpark list`` and associated docs tables
- [ ] Use ``requires`` directives in setting up CI tests 

Depends on #717.

pearce8:
## Preparing ``provides``, ``requires`` directives to enable post-solve check

Example: If the experiment requires cuda, check if the system we are planning to run it on provides cuda.

- [X] Starting ``provides``, ``requires`` directives
- [X] Adding ``requires`` directives to ``OpenMPExperiment``, ``CudaExperiment``, ``ROCmExperiment`` 
- [x] Create ``OpenMPSystem``, ``CudaSystem``, ``ROCmSystem``
- [x] Convert all individual systems to derive from the above classes


pearce8:
@scheibelp I would like to enforce that all CudaSystems:

- [ ] expose Cuda version as a variant
- [ ] specify cuda_arch 
- [ ] specify a default_cuda_version

What is the best way to do this?

scheibelp:
> If the experiment requires cuda, check if the system we are planning to run it on provides cuda.

I think when systems and experiments can interact directly, this will be easier. If all non-cuda systems set cuda as `buildable: false`, then this would at least stop benchpark/spack from trying to build an experiment that isn't likely to work.

> I would like to enforce that all CudaSystems:
> expose Cuda version as a variant

If there was a `CudaSystem` superclass, it could define a `NotImplemented` method to add cuda configs:  any system that inherits this class would have to define that method; likewise for cuda_arch and default version.

pearce8:
@scheibelp That makes sense, thank you.  A couple more questions:

- [ ] We do have system_specific_variables which lets us iterate over all of them - but doesn't enforce the specific ones are set.  
- [ ] We also have id_to_resources structs, which is sometimes where rocm_arch is set (e.g., elcapitan system.py has a tioga variant where tioga has a different rocm_arch than elcapitan)
- [ ] How can I enforce that the classes inheriting from cudasystem specify a cuda variant like this::

```
    variant(
        "cuda",
        default="11-8-0",
        values=("11-8-0", "10-1-243"),
        description="CUDA version",
    )
```

scheibelp:
I was originally thinking of doing this with superclasses, but I think a component-based approach would work better:


```
class System:
    @property
    def components(self):
        return []

    def verify(self):
        for c in self.components:
            c.verify(self)

    def system_specific_variables(self):
        vars = {}
        for c in self.components:
            for x in c.system_attrs():
                vars[x] = getattr(self, x)
        return vars

class Cuda:
    def verify(self, system):
        assert "cuda" in system.variants
        assert hasattr(system, "cuda_arch")

    def system_attrs(self):
        return ["cuda_arch"]

class LlnlSierra(System):
    def __init__(self):
        self.components = [Cuda()]
```

This would address

> We do have system_specific_variables which lets us iterate over all of them - but doesn't enforce the specific ones are set.

and

>  How can I enforce that the classes inheriting from cudasystem specify a cuda variant like this:

and as far as

> We also have id_to_resources structs, which is sometimes where rocm_arch is set (e.g., elcapitan system.py has a tioga variant where tioga has a different rocm_arch than elcapitan)

LlnlElcapitan could be refactored to use an analogous `Rocm` component.

pearce8:
@scheibelp you mentioned a few mechanisms I'd like more details on, please.

- [ ] How to check in the initializer that a specific variant is implemented?
- [ ] In spack there is a notion of an =-style variant where only one value can be assigned (not multiple).

PR #728, opened 2025-04-09
update runners

### LINKED ISSUE #533: Ubuntu version
I noticed the following note in #507 CI actions:

```
[changes](https://github.com/LLNL/benchpark/actions/runs/12397297704/job/34607338185)
ubuntu-latest pipelines will use ubuntu-24.04 soon. For more details, see https://github.com/actions/runner-images/issues/10636
```

Please take a look at the issue they are referencing.  Maybe the problem we were hitting was from an earlier version of ubuntu being used?  It also details what this version restricts us to testing, so worth a check for anything we might care about.

I guess the main question is whether we can go back to using ubuntu-latest once the pipelines are using 24.04 everywhere.

scheibelp:
~This is a work in progress to~ update the Ubuntu image used for the test runs in CI (pinned in https://github.com/LLNL/benchpark/pull/521).

Closes #533.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,7 @@ concurrency:
   cancel-in-progress: true
 jobs:
   changes:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     permissions:
       pull-requests: read
     outputs:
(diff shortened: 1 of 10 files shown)


scheibelp:
Without any changes, just updating to the latest Ubuntu image seems to work. All the logs have since been deleted for PRs that were affected by this (e.g. https://github.com/LLNL/benchpark/pull/515), but I do know it was the dry-run check that was failing, and that works now, so I think something was resolved in the meantime (and this can be merged).

pearce8 (approved):

PR #735, opened 2025-04-18
Implementation of new scaling strategy

rfhaque:
## Description
This PR implements a new scaling strategy for experiments. Any experiment switching to the new implementation must import the required components from`benchpark.new_scaling`

- [x] Update docs
- [x] Update `.github/workflows` and `.gitlab/ci` unit tests (if needed)


pearce8:
Let's separate the required variables which are here implemented in a Variable dict into a separate PR.  Please see how Doug implements this in Ramble.  Please work with Peter because the allocation modifier is filling out some of these variables.  I like the method in the base experiment that throws the notImplemented error and makes it clear to the user which variables to set.

lib/benchpark/experiment.py
```diff
@@ -127,16 +137,21 @@ class Experiment(ExperimentSystemBase, SingleNode):
         description="Append to environment PATH during experiment execution",
     )
 
+    @requires_method("setup_default_experiment")
```
rfhaque: @scheibelp This decorator checks if the class implements the specified method

lib/benchpark/experiment.py
```diff
@@ -127,16 +137,21 @@ class Experiment(ExperimentSystemBase, SingleNode):
...
         self._ramble_name = None
+        self._expr_vars = VariableDict()
+        self._required_vars = []
 
+        visited_helper_impls = set()
         for cls in self.__class__.mro()[1:]:
             if cls is not Experiment and cls is not object:
-                if hasattr(cls, "Helper"):
+                if hasattr(cls, "Helper") and cls.Helper not in visited_helper_impls:
+                    visited_helper_impls.add(cls.Helper)
```
rfhaque: @scheibelp Changed this to ensure the same Helper is not called twice if extended by a child class

lib/benchpark/experiment.py
```diff
@@ -248,18 +279,47 @@ def compute_applications_section_wrapper(self):
...
             self.env_vars["set"] |= env_vars["set"]
             self.env_vars["append"][0] |= env_vars["append"][0]
 
+        self.initialize_experiment_variables()
+
+        visited_required_variables_setters = set()
+        for cls in self.__class__.mro():
+            if hasattr(cls, "register_required_variables") and cls.register_required_variables not in visited_required_variables_setters:
+                visited_required_variables_setters.add(cls.register_required_variables)
+                cls.register_required_variables(self)
```
rfhaque: @scheibelp This code checks for the implementation of the `register_required_variables` method in the current class or any of its super classes in the mro hierarchy. It also makes sure the same method is not called twice under inheritance.

lib/benchpark/variables.py
```diff
@@ -0,0 +1,162 @@
...
+        self._dims = list(self._var.keys())
+        self._ndims = len(self._var)
+        self._named = named
+
+    def __getitem__(self, key):
+        return self._var[key]
+    
+    def __setitem__(self, key, value):
+        if key in self._var:
+            self._var[key].append(value)
```
scheibelp: Seems unexpected that `Variable["x"] = "y"` would append `y` to the list. I think this would make the lists not match in length. I assume callers may iterate through all the dimensions, but a method that updates them all together would likely be safer (and be able to continue with the same care shown in the initializer).

lib/benchpark/variables.py
```diff
@@ -0,0 +1,162 @@
...
+    def prod(self):
+        return self.reduce(lambda x, y: x * y)
+
+    def reduce(self, func):
+        return [reduce(func, col) for col in zip(*self._var.values())]
+
+    @property
+    def min_dim(self):
+        last_values = [v[-1] for v in self._var.values()]
+        return last_values.index(min(last_values))
```
scheibelp, line 143: I'm not clear on what this is supposed to be. The minimum value across all dimensions? If so, I don't think you'd want to start with `v[-1]` (or at least, I don't see why that would be the minimum value for a given dimension).

lib/benchpark/variables.py
```diff
@@ -0,0 +1,162 @@
...
+    @property
+    def is_named(self):
+        return self._named
+
+    @property
+    def prod(self):
+        return self.reduce(lambda x, y: x * y)
+
+    def reduce(self, func):
+        return [reduce(func, col) for col in zip(*self._var.values())]
```
scheibelp, line 132: IMO `math.prod` would be more straightforward

lib/benchpark/variables.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+class VariableDict:
+    def __init__(self):
+        self._vars = {}
+
+    def __getattr__(self, name):
+        if name in self._vars:
+            return self._vars[name]
+        raise AttributeError(f"'DictWrapper' object has no attribute '{name}'")
```
scheibelp: (minor) recommend `__class__.__name__` vs. `DictWrapper`

lib/benchpark/new_scaling.py
```diff
@@ -0,0 +1,398 @@
...
+                Raises an error if scaling down the per-process size does not conserve the global problem size.
+                """
+
+                num_procs = self.expr_vars.num_procs
+                problem_sizes = self.expr_vars.problem_sizes
+
+                if problem_sizes.ndims != num_procs.ndims:
+                    raise BenchparkError(
+                        f"problem_sizes dimensions {problem_sizes.dims} do not match num_procs dimensions {num_procs.dims}"
+                    )
```
scheibelp: I think a docstring would be useful to explain the type of experiment this is supposed to help: I assume not all experiments need process dimensions that map directly to problem dimensions.

lib/benchpark/new_scaling.py
```diff
@@ -0,0 +1,398 @@
...
+    return type(
+        "ExperimentScaling",
+        (BaseScaling,),
+        {
+            "Helper": Helper,
+        },
+    )
+
+
+def UsesPerProcessDomains(*modes):
```
scheibelp: I think a user would only use one of `UsesPerProcessDomains` or `UsesGlobalDomains` (never both): is there a reason these can't be super-classes? Or maybe put another way, did you initially try this as super classes and ran into a modeling issue?

lib/benchpark/experiment.py
scheibelp: [1 - for reference] This seems to have some overlap with https://github.com/LLNL/benchpark/pull/717 - my recommendation there (https://github.com/LLNL/benchpark/pull/717#issuecomment-2807032536) was to add components that can be aggregated in the initializer: I thought that would be more straightforward than MRO inspection; while I think this would work, it feels like if we are creating a shadowing problem that we need to traverse the MRO to solve, it would be cleaner to avoid it in the first place.

lib/benchpark/experiment.py
scheibelp: For methods like this that all `Experiment` subclasses should implement, why not have a method that raises `NotImplemented` (like you did with `initialize_experiment_variables`), rather than this decorator?

One advantage the decorator has is that it's verified at init time, so you don't have to run any particular methods on the class after instantiating it to know it implements the method (or at least declares it). However, I'd still prefer to handle this with the type of validation described in [1] instead.

lib/benchpark/new_scaling.py
```diff
@@ -0,0 +1,398 @@
...
+                start_dim = num_procs.min_dim
+                ndims = num_procs.ndims
+
+                total_problem_size = [orig_global_prob_size]
+         
+                for itr in range(num_exprs):
+                    dim = (start_dim + itr) % ndims
+
+                    num_procs.scale_dim(dim, lambda v: v * scaling_factor)
+                    problem_sizes.scale_dim(dim, lambda v: v // scaling_factor)
```
scheibelp: For initial per-process problem sizes that are not a power of two, this will eventually not cleanly subdivide. You have an error below to catch this, but could this be avoided by taking the ceiling? If not, should this just require up front that the dimension is divisible by `2^x` where `x = num_exprs/ndims`? The outcome is similar, but the error message would be clearer IMO.

lib/benchpark/new_scaling.py
```diff
@@ -0,0 +1,398 @@
...
+    ScalingType = Scaling(*modes)
+
+    for mode in modes:
+        if mode == ScalingMode.Strong:
+            @requires_experiment_variables("num_procs", "problem_sizes")
+            def strong_scale(self):
+                """
+                Strong scales the problem by increasing the number of processes along each dimension,
+                starting with the minimum process dimension, in a round-robin manner.
+                There are no changes to the per process sizes.
```
scheibelp: I was assuming if you increase the number of processes, that these types of experiments (i.e. ones that apply `UsesGlobalDomains`) were capable of splitting up the work between the processes automatically. In that sense, I think

> There are no changes to the per process sizes.

isn't quite accurate: if the benchmark were not able to change the per-process problem size, I think strong scaling would not be possible (i.e. the total problem size would always increase when the # of processes was increased).

scheibelp (commented):
I have a few questions and suggestions.

experiments/amg2023/experiment.py
```diff
@@ -40,115 +35,99 @@ class Amg2023(
...
+        # then after 4 scaling iterations (3 scalings), the 
+        # final values of the scaled variables will be
+        # "np" : { "px": [2,2,4,4]
+        #          "py": [2,2,2,4]
+        #          "pz": [1,2,2,2] } and, 
+        # "probs" : { "nx": [16,16,32,32]
+        #             "ny": [32,32,32,64]
+        #             "nz": [32,64,64,64] },
+        # Note that scaling starts with the minimum value dimension (pz) of the
+        # first variable (np) and proceeds in a round-robin manner
```
rfhaque: @pearce8 Please review the scaling implementation.

experiments/amg2023/experiment.py
```diff
@@ -3,15 +3,12 @@
...
-from benchpark.error import BenchparkError
 from benchpark.directives import variant, maintainers
 from benchpark.experiment import Experiment
 from benchpark.openmp import OpenMPExperiment
 from benchpark.cuda import CudaExperiment
 from benchpark.rocm import ROCmExperiment
-from benchpark.scaling import StrongScaling
-from benchpark.scaling import WeakScaling
-from benchpark.scaling import ThroughputScaling
+from benchpark.new_scaling import ScalingMode, Scaling
```
rfhaque, line 11: @pearce8 An experiment must import new_scaling to use the new implementation. Other experiments using the current scaling implementation are unaffected. Eventually, all experiments will be converted to the new scaling approach.

lib/benchpark/experiment.py
```diff
@@ -306,12 +313,13 @@ def compute_modifiers_section_wrapper(self):
...
 
-    def add_experiment_variable(self, name, values, use_in_expr_name=False):
-        self.variables[name] = values
-        if use_in_expr_name:
-            self.expr_name.append(f"{{{name}}}")
+    def add_experiment_variable(self, name, values, named=False, matrixed=False):
+        if isinstance(values, dict):
+            self.expr_vars.add_dimensional_variable(name, values, named, matrixed)
+        else:
+            self.expr_vars.add_scalar_variable(name, values, named, matrixed)
```
scheibelp: Two questions

1 - does something go wrong if scalars are treated like 1D vectors?

2 - what does `matrixed` mean?

I think I only ever see it set `True` for `n_threads_per_proc` when it has a vector of values, like:

```
self.add_experiment_variable("n_threads_per_proc", ["2", "4"], True, True)
```

so it seems like something that could be automatically inferred from `values`, like `if isinstance(values, list)`

lib/benchpark/variables.py
scheibelp: After reading through things: this is is for values that are a list of vectors, it pulls the last vector off the list, then figures out which dimension is the lowest value and returns that (because that's where the scaling starts. IMO a comment explaining that would be useful

lib/benchpark/new_scaling.py
```diff
@@ -0,0 +1,203 @@
...
+        for itr in range(num_exprs):
+            dim = (start_dim + itr) % ndims
+            for var_name, scaling_func in scaling_config.items():
+                getattr(self.expr_vars, var_name).scale_dim(
+                    itr, dim, scaling_func, scaling_factor
+                )
+
+    BaseScaling.scale_params = scale_params
+
+    # The register_scaling_policy method defines the scaled variables and their
```
scheibelp: Now called `register_scaling_config` (I think, but in any case this method name doesn't exist anymore)

lib/benchpark/new_scaling.py
```diff
@@ -0,0 +1,203 @@
...
+    # scaling_factor: The factor by which the variable dimension must be scaled
+    # The scaling_function must return the new scaled value for the variable dimension
+    #
+    # scaling starts from the dimension with the minimum value for the first variable
+    # in the list of variables and proceeds through the dimensions in a round-robin
+    # manner for the specified number of scaling iterations
+    # e.g. if the scaling config is defined as:
+    # ScalingMode.Strong: {
+    #     "np": lambda var, itr, dim, sf: var.val(dim) * sf,
+    #     "sizes": lambda var, itr, dim, sf: var.val(dim) * sf,
```
scheibelp: Should this be "probs" rather than "sizes" (or in the below dict, change "probs" to "sizes")

You may want to explain that this implements a global scaling policy, and e.g. per-process scaling you would drop `* sf` on "slizes" (for weak scaling, and `/ sf` for strong)

lib/benchpark/new_scaling.py
```diff
@@ -0,0 +1,203 @@
...
+    #    return ...
+    # The arguments for the scaling_function are:
+    # var: benchpark.Variable instance of the scaled variable
+    # itr: The current iteration in the specified number of scaling iterations
+    # dim: The current dimension that is being scaled
+    # scaling_factor: The factor by which the variable dimension must be scaled
+    # The scaling_function must return the new scaled value for the variable dimension
+    #
+    # scaling starts from the dimension with the minimum value for the first variable
+    # in the list of variables and proceeds through the dimensions in a round-robin
```
scheibelp, line 145: Question about this: if you start with a vector like

```
2
4
2
```

do you want to scale it like

```
2    4    4
4 -> 4 -> 8
2    2    2
```

or 

```
2    4    4
4 -> 4 -> 4
2    2    4
```

experiments/amg2023/experiment.py
```diff
@@ -40,115 +35,99 @@ class Amg2023(
...
+        # Note that scaling starts with the minimum value dimension (pz) of the
+        # first variable (np) and proceeds in a round-robin manner
+
+        self.register_scaling_config(
+            {
+                ScalingMode.Strong: {
+                    "num_procs": lambda var, itr, dim, scaling_factor: var.val(dim)
+                    * scaling_factor,
+                    "problem_sizes": lambda var, itr, dim, scaling_factor: var.val(dim)
+                    // scaling_factor,
```
scheibelp, line 74: A couple things:

1. Passing in the functions offers customizability, but do you have in mind other setups besides per-process or global scaling semantics, that would be easily implemented by passing in other functions? I think it would make it clearer if you had a label for global/per-process and then define the functions automatically based on that label (you could also offer this mechanism if they want complete control)
2. I think this inherits my concern from https://github.com/LLNL/benchpark/pull/735#discussion_r2060908690 - i.e. for certain numbers this will not cleanly subdivide (in this case I don't think there is any catch for it, where before you caught it at the point where it would have become a float (using `//` keeps things as int, but I wanted to check: if you scale a per-process size down like `18, 9, 4, 2, 1` and scale processes up like `1, 2, 4, 8, 16`, would that be desired?)

lib/benchpark/experiment.py
```diff
@@ -306,12 +313,13 @@ def compute_modifiers_section_wrapper(self):
         return modifier_list
 
     def add_experiment_name_prefix(self, prefix):
```
scheibelp: Is this function still needed? I don't see anything calling it anymore.

lib/benchpark/new_scaling.py
```diff
@@ -0,0 +1,203 @@
...
+from enum import Enum
+
+
+class ScalingMode(Enum):
+    Strong = "strong"
+    Weak = "weak"
+    Throughput = "throughput"
+
+
+def Scaling(*modes):
```
scheibelp, line 19: I see this lets you generate a superclass on the fly, e.g. in `amg2023`:

```
class Amg2023(
    Experiment,
    OpenMPExperiment,
    CudaExperiment,
    ROCmExperiment,
    Scaling(ScalingMode.Strong, ScalingMode.Weak, ScalingMode.Throughput),
    Caliper,
):
```

I'm curious, what does that enable over the an inheritance model like:

```
class Amg2023(
    ...
    StrongScaling
    WeakScaling
    ...
):
```

?

lib/benchpark/variables.py
```diff
@@ -0,0 +1,153 @@
...
+    @property
+    def is_named(self):
+        return self._named
+
+    @property
+    def is_matrixed(self):
+        return self._matrixed
+
+    @property
+    def prod(self):
```
scheibelp: Is this used? I was trying to get a sense of what it's intended for but I don't see things calling it.

experiments/amg2023/experiment.py
```diff
@@ -40,115 +35,99 @@ class Amg2023(
...
-
-        scaling_mode_enabled = [key for key, value in scaling_modes.items() if value]
-        if len(scaling_mode_enabled) != 1:
-            raise BenchparkError(
-                f"Only one type of scaling per experiment is allowed for application package {self.name}"
-            )
-
         # Number of processes in each dimension
-        num_procs = {"px": 2, "py": 2, "pz": 2}
+        self.add_experiment_variable("num_procs", {"px": 2, "py": 2, "pz": 2}, True)
```
scheibelp: (question) Is there a desire to have users be able to specify `benchpark experiment init num_procs=2,2,2` on the command line? That doesn't have to happen here, but I'm curious if you find the interface cumbersome for doing that.

scheibelp (commented):
I have a few requests and several questions. I also collapsed earlier comments that I think no longer apply.

lib/benchpark/experiment.py
rfhaque: > Two questions
> 
> 1 - does something go wrong if scalars are treated like 1D vectors?
> 
> 2 - what does `matrixed` mean?
> 
> I think I only ever see it set `True` for `n_threads_per_proc` when it has a vector of values, like:
> 
> ```
> self.add_experiment_variable("n_threads_per_proc", ["2", "4"], True, True)
> ```
> 
> so it seems like something that could be automatically inferred from `values`, like `if isinstance(values, list)`

"scalar" here means scalar in the context of the application (e.g. mass), compared to dimensional variables (e.g. num_procs {px, py, pz})
In benchpark/ramble, a scalar variable can be a 1D vector each corresponding to a separate experiment. All dimensional variables are themselves laid out as their scalar components in the `ramble.yaml` e.g. num_procs is written as:
px : [values for x dim]
py : [values for y dim]
pz : [values for z dim]
Each application variable must either be a single value of a 1D vector of the same length as other variables except if a variable is "matrix"ed which we define using `self.matrix_experiment_variables`

lib/benchpark/new_scaling.py
rfhaque: Will fix that

lib/benchpark/new_scaling.py
rfhaque: It is round-robin
```
2    4    4
4 -> 4 -> 8
2    2    2
```

experiments/amg2023/experiment.py
rfhaque: > A couple things:
> 
>     1. Passing in the functions offers customizability, but do you have in mind other setups besides per-process or global scaling semantics, that would be easily implemented by passing in other functions? I think it would make it clearer if you had a label for global/per-process and then define the functions automatically based on that label (you could also offer this mechanism if they want complete control)
There are some benchmarks (e.g. kripke) where it's possible to scale the problem on the basis of other variables than the ones we currently define scaling for. It's not clear how the scaling would look like, but having a customizable scaling function for each variable ensures that any scaling logic is explicitly defined in the experiment.py itself. Anyone looking at the experiment.py should be able to piece together/modify the scaling strategy per their need.

> 
>     2. I think this inherits my concern from [[WIP] Implementation of new scaling strategy #735 (comment)](https://github.com/LLNL/benchpark/pull/735#discussion_r2060908690) - i.e. for certain numbers this will not cleanly subdivide (in this case I don't think there is any catch for it, where before you caught it at the point where it would have become a float (using `//` keeps things as int, but I wanted to check: if you scale a per-process size down like `18, 9, 4, 2, 1` and scale processes up like `1, 2, 4, 8, 16`, would that be desired?)
Right now I've omitted the check for that, but we can add that

lib/benchpark/new_scaling.py
rfhaque: This gives us the ability to enable only one scaling mode at a time:
With the current setup it's up to the experiment to handle scaling modes correctly, e.g. we can write ` amg2023 +strong +weak` and the `experiment.py` has to make sure this is handled correctly. 
With the dynamic superclass, each experiment gets a single-value variant of the form 
`scaling : [off, only the scaling modes defined for the experiment]`

lib/benchpark/variables.py
rfhaque: Probably not

experiments/amg2023/experiment.py
rfhaque: The interface might become quite cumbersome if an experiment has a lot of variables. It might be easier to define a method `init_experiment_variables` that sets the initial values of experiment variables.

experiments/laghos/experiment.py
```diff
@@ -57,33 +46,41 @@ def compute_applications_section(self):
...
-        scaling_factor = {"scaling_factor": 1}
+        self.add_experiment_variable("scaling_factor", 1, False)
+
+        # Register the scaling variables and their respective scaling functions
+        # required to correctly scale the experiment for the given scaliing policy
+        self.register_scaling_config(
+            {
+                ScalingMode.Strong: {
+                    "scaling_factor": lambda var, itr, dim, scaling_factor: var.val(dim)
+                    * scaling_factor,
```
scheibelp, line 57: [1] (for reference)

It seems a bit confusing to use an internal variable called `scaling_factor` to scale up an experiment variable also called `scaling_factor`, although running `benchpark experiment init --dest ex-laghos laghos scaling=strong` gives a functional-looking `ramble.yaml`.

I think if this were refactored, it could be made to match kripke/amg2023 definition, at which point we could potentially factor out the common scaling configs into a `global`/`per_process` definition.

I think in this case it's per process, so e.g. we could adapt `amg2023` from

```
                    "problem_sizes": lambda var, itr, dim, scaling_factor: var.val(dim)
                    // scaling_factor,
```

into

```
                    "process_problem_size": lambda var, itr, dim, scaling_factor: var.val(dim)
                    // scaling_factor,
```

i.e.

```
                n_ranks: '{sys_cores_per_node} * {scaling_factor}'
                n_resources: '{n_ranks}'
                n_threads_per_proc: 1
                process_problem_size: '{zones} / {n_resources}'
                scaling_factor:
                - 1
                - 2
                - 4
                - 8
                total_problem_size: '{zones}'
                zones: 1024

--->

                # For this to work, I think we have to take the system
                # context out of it (that could be reintroduced when I
                # connect experiments to systems)
                n_ranks: '{n_resources}'
                n_resources: '{px}'
                n_threads_per_proc: 1
                process_problem_size:
                - 256
                - 128
                - 64
                - 32
                # Only one dimension of procs for laghos
                px:
                - 8
                - 16
                - 32
                - 64
                total_problem_size: '{process_problem_size} * {n_ranks}'
                process_zones: 256
```

experiments/amg2023/experiment.py
scheibelp: >> do you have in mind other setups besides per-process or global scaling semantics

> There are some benchmarks (e.g. kripke) where it's possible to scale the problem on the basis of other variables than the ones we currently define scaling for.

That makes sense, although I think everything in this PR is global or per-process scaling, and it would be useful if they could use a common definition (looking at it more I see why that would be hard currently, which I think could be mitigated in part by [1]: https://github.com/LLNL/benchpark/pull/735#discussion_r2151112367)

scheibelp (commented):
I still need to think more about https://github.com/LLNL/benchpark/pull/735#discussion_r2138977123, but I spent some time thinking about `register_scaling_config` and wanted to ask about that in the meantime

lib/benchpark/variables.py
scheibelp: (do here) Can you rename this to a function like `append_key`? I think it's going to confuse people to give this operation append behavior.

lib/benchpark/experiment.py
```diff
@@ -301,13 +308,16 @@ def compute_modifiers_section_wrapper(self):
...
         return modifier_list
 
-    def add_experiment_name_prefix(self, prefix):
-        self.expr_name = [prefix] + self.expr_name
-
-    def add_experiment_variable(self, name, values, use_in_expr_name=False):
-        self.variables[name] = values
-        if use_in_expr_name:
-            self.expr_name.append(f"{{{name}}}")
+    def add_experiment_variable(self, name, values, named=False, matrixed=False):
```
scheibelp, line 311: (do here) I think this should explain the purpose of `named`/`matrixed`: when should they be set to `True`?

lib/benchpark/variables.py
```diff
@@ -0,0 +1,164 @@
...
+
+    def __getattr__(self, name):
+        if name in self._vars:
+            return self._vars[name]
+        raise AttributeError(f"'{__class__.__name__}' object has no attribute '{name}'")
+
+    # values must be a dict of type str->type or str->list(type)
+    def add_dimensional_variable(
+        self, name, values, named=False, zipped=True, matrixed=False
+    ):
```
scheibelp, line 22: (do here) Is `named` ever not `True` for this? If it's always true, the parameter should be removed.

scheibelp (commented):
I have a few small requests for this PR (I've prefixed them with "do here"). There's several larger comments that still apply, but I don't need to block the PR: they can be taken as suggestions for a follow-up PR - that includes:

- https://github.com/LLNL/benchpark/pull/735#discussion_r2141196746
    - Potential issues with "misalignment" of resources/problem sizes when they don't cleanly map to powers of 2
    - Right now we just have global and per-process scaling, but the mechanism for setting these up IMO could be more straightforward
        - I think documentation/comments around where you do `self.register_scaling_config(` like "this is global scaling" would help
- https://github.com/LLNL/benchpark/pull/735#discussion_r2151112367
    - This gets into how the implementation could be refactored to have a common concept of global/per-process scaling

lib/benchpark/experiment.py
rfhaque: @scheibelp 
- Setting `named` to `True` adds the (unconsumed) variable name to the experiment name in `ramble.yaml`
- Setting `matrixed` to `True`, adds the variable to the experiment matrix, that allows generating experiments as a cross product of variables.
I'll add documentation for that

lib/benchpark/variables.py
rfhaque: @scheibelp `named` is set to `True` in `experiment/py` for variables whose name should appear in the experiment name e.g. to get the experiment name
`amg2023_problem1_single_node_rocm_no_scaling_caliper_mpi_{px}_{py}_{pz}_{nx}_{ny}_{nz}_{n_gpus}`, 
in `amg2023/experiment.py` we set 
```
self.add_experiment_variable(
    "n_resources_dict", {"px": 2, "py": 2, "pz": 2}, named=True
)
self.add_experiment_variable(
    "process_problem_size_dict", {"nx": 80, "ny": 80, "nz": 80}, named=True
)
self.add_experiment_variable("n_gpus", "{n_resources}", named=True)
```

lib/benchpark/experiment.py
rfhaque: @scheibelp 
> (do here) Can you rename this to a function like append_key? I think it's going to confuse people to give this operation append behavior.

I will remove this function altogether. It's not being used and was only added for the sake of completeness (complementing the definition of `__getitem__`)

lib/benchpark/variables.py
rfhaque: Added documentation to explain the `add_experiment_variable` method

pearce8 (dismissed):

pearce8 (approved):

PR #584, opened 2025-01-31
[WIP] Add modifier for affinity package

rfhaque:
## Description
[WIP] Adding a modifier that prints thread/core/gpu affinity for an experiment.   
- [x] Potentially closes #480.
- [ ] Update docs
- [ ] Update `.github/workflows` and `.gitlab/ci` unit tests (if needed)

PR #740, opened 2025-04-24
Add Affinity modifier docs

pearce8:
- [x] Docs for #584


michaelmckinsey1 (changes requested):
@pearce8 Are we intending to support `affinity.omp`? It was not included in https://github.com/LLNL/benchpark/pull/584. I removed it from this doc to reflect the current state, but we can add the option for `affinity.omp` in this PR.

pearce8:
@michaelmckinsey1 please add the option for `affinity.omp` in this PR.

michaelmckinsey1 (approved):
We've determined we don't need `affinity.omp`, as it is without MPI

PR #750, opened 2025-05-05
Set and Require `n_resources`, `total_problem_size`, and `process_problem_size` in experiment.py

### LINKED ISSUE #723: Formalizing scaling and how to think about problem sizes
### Example: 

**Kripke** 

1. Benchmark parameters are ``total_problem_size`` (in 3 dimensions) which are then decomposed into ``process_problem_size``s (in 3 dimensions)
2. For strong scaling, the problem size parameters stay the same - while the number of processors varies
3. For weak scaling, we need to double the problem size as we double the number of processors
4. ``process_problem_size`` = ``nx``x``ny``x``nz``/(``px``x``py``x``pz``)
5. ``total_problem_size`` = ``nx``x``ny``x``nz``
6. ``nprocs`` = ``px``x``py``x``pz``

**AMG2023** 

1. Benchmark parameters are ``process_problem_size``s (in 3 dimensions) which are then multiplied by the number of processors to compute ``total_problem_size`` 
2. For strong scaling, we need to half the problem size as we double the number of processors 
3. For weak scaling, the problem size parameters stay the same - while the number of processors varies
4. ``process_problem_size`` = ``nx``x``ny``x``nz``
5. ``total_problem_size`` = ``nx``x``ny``x``nz``x``px``x``py``x``pz``
6. ``nprocs`` = ``px``x``py``x``pz``

### How we can address this problem

#750 
- [x] Require that the user specifies that the benchmark parameters are ``total_problem_size`` or ``process_problem_size``
- [x] The user should have to specify how to use the parameters to compute ``process_problem_size`` and ``total_problem_size``
- [x] Fix all of the experiments accordingly

#761 
- [x] Store `process_problem_size`` and ``total_problem_size`` in the Caliper metadata
- [x] (later) when we know the number of processors, need to be able to compute ``node_problem_size`` and store in Caliper metadata
- [x] (later) canned analysis should show **Strong scaling**: nprocs on x, time on y, line per ``total_problem_size`` (the line should go down)
- [x] (later) canned analysis should show **Weak scaling**: nprocs on x, time on y, line per ``process_problem_size`` (or possibly ``node_problem_size``) (the line should be flat)

### LINKED ISSUE #843: Replicate RAJAPerf Experiments
## Description

Enable replication of RAJAPerf experiments from recent Jan 2025 paper. These experiments were not ran in Benchpark, but were built with cmake & build scripts and ran manually.

- [x] Defining per-process problem size and scaling modes correctly for RAJAPerf #750 
- [ ] Caliper configs we need to be able to use:
   - [x] time
   - [ ] topdown
   - [ ] gpu-side calls (cuda-activity-profile)
   - [ ] ncu 
- [ ] Replicate Datasets:
   - [x] V100
   - [ ] MI250x
   - [x] MI300a #829 
   - [ ] A100
   - [ ] GH200

pearce8:
- [X] Require setting `n_resources`, `total_problem_size`, `process_problem_size` to each `experiment.py`.
- [X] A parent class to enforce variable names
- [x] Laghos - problem size defined in mesh, unknown in experiment.py -> hardcode in experiment.py
- [x] Remhos - epm being set in experiment now 
- [x] What needs to be done to propagate to Caliper for canned analysis?
   - [x] requires changes in `lib/benchpark/caliper.py` to expect these experiment variables. 
- [x] Add docs
- [x] Closes #723 
- [x] Tracking in #776
   - [x] What to do with applications which only have one size?  What makes sense for the allocation modifier?
   - [x] How to change the allocation modifier to expect these 3 variables? 

## Experiments
- [x] Change `experiments/*/experiment.py` to set `n_resources`, `total_problem_size`, and `process_problem_size`.

## Applications
- [x] Update `repo/remhos/application.py` to set `epm` from experiment, so it can be used in problem size.
- [x] Update `repo/raja-perf/application.py` 

## Other
- [x] Update `docs/add-an-experiment.rst` for section on `comput_applications_section` and setting required variables.
- [x] Fix `lib/benchpark/test/caliper.py` unit test for req variables metadata.


experiments/kripke/experiment.py
```diff
@@ -80,6 +80,7 @@ def compute_applications_section(self):
             for nk, nv in problem_sizes.items():
                 self.add_experiment_variable(nk, nv, True)
         elif self.spec.satisfies("+throughput"):
+            problem_sizes = {"nzx": 32, "nzy": 32, "nzz": 32} # TEMP: Lower problem size for lassen oom error
```
michaelmckinsey1: Needs to be undone before merge

michaelmckinsey1:
raja-perf_mpi_strong_time_exc

The strong scaling implementation in RAJAPerf shows how to currently achieve strong scaling for benchmarks that use per-process problem size.

Do we want to do the same for amg until a long-term solution is made?

experiments/raja-perf/experiment.py
```diff
@@ -35,29 +36,83 @@ class RajaPerf(
...
+                {tuple(problem_sizes.keys()): list(problem_sizes.values())},
+                1 / int(self.spec.variants["scaling-factor"][0]),
+                int(self.spec.variants["scaling-iterations"][0]),
+            )
+            problem_sizes = scaled_problem_sizes["size"]
+            for nk, nv in scaled_problem_sizes.items():
+                self.add_experiment_variable(nk, nv, True)
 
-        if self.spec.satisfies("+cuda") or self.spec.satisfies("+rocm"):
+        if self.spec.satisfies("+cuda"):
```
pearce8, line 92: Please give RAJAPerf variables specific names, such as RAJAPerf_variant, RAJAPerf_tuning.  Especially since "variant" is a keyword in benchpark/spack.

lib/benchpark/experiment.py
```diff
@@ -133,6 +133,7 @@ def __init__(self, spec):
         self.helpers = []
         self._spack_name = None
         self._ramble_name = None
+        self.req_vars = ["n_resources", "process_problem_size", "total_problem_size"]
```
pearce8: @michaelmckinsey1 Please add a device_type variable, which can be set to cpu or gpu.  You can then set it to cpu in  openmp.py, and gpu in rocm.py and cuda.py.  It is possible that we need a mpionly experiment for CPU experiments that don't use openmp......

experiments/raja-perf/experiment.py
michaelmckinsey1: [0e6c92d](https://github.com/LLNL/benchpark/pull/750/commits/0e6c92d1cacb5340e1f63299c09ecd7b537926a9)

lib/benchpark/experiment.py
michaelmckinsey1: [23c80bc](https://github.com/LLNL/benchpark/pull/750/commits/23c80bc6151a0efebeffba890bfdba7bad0d69cf) sets `device_type` to "cpu" by default which will be overwritten by openmp, cuda, and rocm (openmp is same value)

pearce8 (commented):
@michaelmckinsey1 Please fix lint.

pearce8 (commented):
@michaelmckinsey1 Please fix coverage tests.

lib/benchpark/experiment.py
```diff
@@ -129,10 +129,18 @@ class Experiment(ExperimentSystemBase, SingleNode):
 
     def __init__(self, spec):
         self.spec: "benchpark.spec.ConcreteExperimentSpec" = spec
+        # Device type must be set before super with absence of mpionly experiment type
+        self.device_type = "cpu"
```
michaelmckinsey1, line 133: @pearce8 MPI-only experiments will be set here, because the will still inherit from the `Experiment` class.

pearce8 (commented):
@michaelmckinsey1  please fix:

- [x] pytest
- [x] ~coverage~ 
- [x] amg2023 strong scaling

michaelmckinsey1:
amg2023 scaling after changes to adjust for per-process problem size

strong cpu (gpu does not scale)
amg2023_mpi_strong_time_exc

weak gpu
amg2023_cuda_weak_time_exc

PR #758, opened 2025-05-09
Fix input source in lammps ramble repo

rfhaque:
## Description
This PR fixes the input source repo path in ramble application.py. These changes need to be upstreamed to the main Ramble repo.


pearce8 (changes requested):
Please implement this in package.py instead of (use Laghos data install as an example).

michaelmckinsey1:
Upstream Ramble change is https://github.com/GoogleCloudPlatform/ramble/pull/1022, would still need the change in benchpark until our Ramble is updated again.

rfhaque:
> Please implement this in package.py instead of (use Laghos data install as an example).

@pearce8 Done. Please review. We'll need to upstream the application.py to Ramble. Tested on lassen and tuolumne

pearce8 (approved):

PR #637, opened 2025-02-24
Name mangling remove import hack

becker33:
## Description

- [x] Name mangle all ramble references to spack at bootstrap time instead of hacky imports of Spack. This allows us to import from both Spack and Ramble.
- [x] This will allow us to update to more modern Spack (@slabasan)

## Adding/modifying core functionality, CI, or documentation:

- [x] Should not require docs update
- [x] Should not require changes to CI

PR #762, opened 2025-05-14
Update spack library

scheibelp:
See also: https://github.com/LLNL/benchpark/pull/637

We want to be able to update Spack. There was a divergence between Spack/Ramble that caused a conflict when using both as libraries within benchpark. #637 is looking at how to make Spack "fully" import-able by name mangling it. I am continuing that work, but in the course of that noticed that only one command in benchpark seems to care about importing spack as a library. I'm testing removing that here.

If this works out, then Spack isn't needed as a lib (and it the workspaces will be able to use newer Spack versions w/o having to deal with name mangling).


scheibelp:
Current failures are because venado/capitan systems generate invalid `packages` config, namely `"gtl_libs": ["libmpi_gtl_cuda"],` (used by the benchpark `cray-mpich` package), the schema only supports strings as values (vs. a list in this case). AFAICT this was always true (including for older spack), but not triggered until now.

See also:

* https://github.com/LLNL/benchpark/blob/develop/systems/lanl-venado/system.py
* https://github.com/LLNL/benchpark/blob/develop/repo/cray-mpich/package.py

lib/benchpark/paths.py
```diff
@@ -17,6 +17,7 @@ def _source_location() -> pathlib.Path:
 lib_path = benchpark_root / "lib" / "benchpark"
 test_path = lib_path / "test"
 benchpark_home = pathlib.Path(os.path.expanduser("~/.benchpark"))
+# benchpark_home = benchpark_root / ".benchpark"
```
michaelmckinsey1:
```suggestion
```

pearce8 (dismissed):

pearce8 (approved):

PR #765, opened 2025-05-15
Laghos: patch source to fix cuda build error for caliper

rfhaque:
## Description
This PR fixes an issue with building laghos in the cuda mode with caliper enabled. The patch needs to be upstreamed to the laghos source. 

- [x] If modifying the source code of a benchmark: create, self-assign, and link here a follow up issue with a link to the PR in the benchmark repo.


pearce8 (changes requested):
@rfhaque Please put up a PR to Daryl's branch of laghos instead of adding the patch here.

rfhaque:
> @rfhaque Please put up a PR to Daryl's branch of laghos instead of adding the patch here.

@pearce8 I'll do that. However, we still need to keep one change in laghos/package.py. We can merge this PR and I can create a new one once laghos source changes get merged

michaelmckinsey1 (approved):

PR #780, opened 2025-05-21
Update Affinity Modifier

### LINKED ISSUE #741: Add affinity modifier data to Caliper attributes if running with both modifiers
- [X] Building on #584 and #740 
- [ ] What does Affinity modifier need to output?
- [ ] How to get Caliper to pick up that output?
- [ ] Can we specify the Affinity modifier to only run on one node of the job (even if more nodes are allocated)
- [ ] We always allocate the node exclusively.  When the subsequent experiment uses a subset of resources (e.g., allocated the node exclusively but only using 4 ranks because there are 4 GPUs on the node), how can we run the allocation modifier on the 4 ranks as well?

michaelmckinsey1:
## Description

- [x] Changes to the affinity modifier:
   - [x] Set affinity based on programming model in spec
      - setting affinity now looks like `benchpark experiment init --dest amg2023/cuda amg2023+cuda affinity=on caliper=time`, where cuda affinity will be inferred from `+cuda` in spec    
   - [x] Expose affinity to all experiments by setting in base class, meaning we don't need to inherit from `Affinity` in each experiment file separately. 
   - [x] Inject affinity metadata json into caliper configuration to add to adiak metadata, enabling affinity information in `.cali` files generated with caliper modifier.
   - [x] Fix newline issue from json output by removing indent in script

- [x] Remove affinity from `experiments/saxpy/experiment.py` to abstract to base class

- [x] Update docs
- [x] Update affinity modifier files
- [x] Update 3 dyanmic dry run unit tests with affinity


pearce8 (changes requested):
- [x] fix lint
- [x] add a test that uses the affinity modifier

pearce8 (approved):

PR #770, opened 2025-05-19
Laghos, Remhos Bugfix

rfhaque:
This PR fixes several issues with laghos and remhos builds.

- [x] Update laghos/remhos package.py to use the latest mfem and hypre versions. 
- [x] Fixes the mfem build issue for rocm >= 6 #757
- [x] Closes issues #718, #757, #715
- [x] Subsumes: #719
- [x] Waiting on merge: https://github.com/wdhawkins/Laghos/pull/1
- [x] After https://github.com/wdhawkins/Laghos/pull/1, what needs to be done?  Before merging this PR, the laghos source repo needs to be merged and the URL in laghos package.py should be reverted to the original one
- [x] If modifying the source code of a benchmark: create, self-assign, and link here a follow up issue with a link to the PR in the benchmark repo.
- [x] If package.py upstreamed to Spack is insufficient, add/modify `repo/benchmark_name/package.py` plus: create, self-assign, and link here a follow up issue with a link to the PR in the Spack repo.

Leaving to address in future PRs, documenting in Issues #851 and #852.
- [ ] This PR also uses the latest mfem spack package.py. Once benchpark is updated to the latest spack commit (#762), mfem should use the upstreamed spack version - verify, remove from repo.  
- [ ] Waiting on merge: https://github.com/CEED/Remhos/pull/53. Remhos also tries to include a .cpp file from mfem, that is not installed as a header by the mfem build system (https://github.com/LLNL/benchpark/issues/718). 
- [ ] After https://github.com/CEED/Remhos/pull/53 merge, remove the hack in this PR (explicit copy of the missing include to the intended location). 
- [ ] Waiting on merge: https://github.com/mfem/mfem/pull/4866. 
- [ ] After https://github.com/mfem/mfem/pull/4866 merge, what needs to be done?

Update:
06/24/2025: This PR now uses c++17 standard for the mfem build. It has been tested on tioga/tuolumne, lassen, matrix and dane for both laghos and remhos with the following commit versions:
```
Test date: 06/24/025
benchpark: 01dd9611ee1317a5f9a646236c97946d5a2beb30
ramble: 6b4ecac  # develop on 5/06/2025 (newer than 0.6.0 release)
spack: 7e4b8aa  # develop on 2/22/2025 (pre-breaking changes on 1.0)
```
The test results are as follows:
**Laghos:**
_tioga/tuolumne:_
```
[...]

PR #784, opened 2025-05-22
Clone From Local Bootstrap to Speedup `benchpark setup`

michaelmckinsey1:
## Description

- [x] Instead of cloning from the internet, clone from local bootstrap for benchpark workspaces.
- [x] Enable setting an upstream in `RuntimeResources`, explicitly connecting the bootstrapped spack/ramble and workspace spack/ramble.
   - [x] Upstream set in `benchpark setup ...` (`setup.py`, when benchpark workspace is created, so bootstrap created by previous command no longer required, as it will be created here if it does not exist.
   - [x] Use upstream to bootstrap (if instance does not have an upstream bootstrapping will occur).
   - [x] Upstream uses remote url for clone, "origin" uses local upstream repository to clone.
- [x] Refactor remote urls logically similar to hash versions. 

## Cloning Performance improvement

local performance ~8x
```
develop - benchpark setup ...
real    0m55.378s
user    0m25.981s
sys     0m20.266s

#784 - benchpark setup ...
real    0m7.611s
user    0m4.146s
sys     0m5.958s
```

dane performance ~2x
```
develop - benchpark setup ...
real    4m33.224s
user    0m50.784s
sys     0m23.990s

#784 - benchpark setup ...
real    2m14.887s
user    0m4.101s
sys     0m9.247s
```

- [x] Update `lib/benchpark/runtime.py`

### DIFF
--- a/lib/benchpark/runtime.py
+++ b/lib/benchpark/runtime.py
@@ -64,9 +64,13 @@ class Command:
 
 
 class RuntimeResources:
-    def __init__(self, dest):
+    def __init__(self, dest, upstream=None):
         self.root = benchpark.paths.benchpark_root
         self.dest = pathlib.Path(dest)
+        self.upstream = upstream
+
+        self.ramble_location = self.dest / "ramble"
+        self.spack_location = self.dest / "spack"
 
         checkout_versions_location = self.root / "checkout-versions.yaml"
         with open(checkout_versions_location, "r") as yaml_file:
@@ -74,8 +78,9 @@ class RuntimeResources:
             self.ramble_commit = data["versions"]["ramble"]
             self.spack_commit = data["versions"]["spack"]
 
-        self.ramble_location = self.dest / "ramble"
-        self.spack_location = self.dest / "spack"
+        # If instance does not have an upstream, it is the upstream
+        if not self.upstream:
+            self.bootstrap()
 
     def update_old_bootstrap(self, desired_commit, location):
         # Store first 7 of hash in checkout-versions.yaml
--- a/lib/benchpark/runtime.py
+++ b/lib/benchpark/runtime.py
@@ -83,7 +83,7 @@ class RuntimeResources:
             remote_ramble_url = data["urls"]["ramble"]
             remote_spack_url = data["urls"]["spack"]
 
-        # If instance does not have an upstream, it is the upstream
+        # If this does not have an upstream, then we will be cloning from the URLs indicated in remote-urls.yaml
         if self.upstream is None:
             self.ramble_url = remote_ramble_url
             self.spack_url = remote_spack_url
--- a/lib/benchpark/runtime.py
+++ b/lib/benchpark/runtime.py
@@ -84,7 +84,7 @@ class RuntimeResources:
             self.spack_url = data["urls"]["spack"]
 
         # If instance does not have an upstream, it is the upstream
-        if not self.upstream:
+        if self.upstream is None:
             self.bootstrap()
 
     def _check_and_update_bootstrap(self, desired_commit, location):
@@ -124,7 +124,7 @@ class RuntimeResources:
         git_clone_commit(
             (
                 self.ramble_url
-                if not self.upstream
+                if self.upstream is None
                 else self.upstream.ramble_location  # Clone from local "upstream" repository
             ),
             self.ramble_commit,
@@ -137,7 +137,7 @@ class RuntimeResources:
         git_clone_commit(
             (
                 self.spack_url
-                if not self.upstream
+                if self.upstream is None
                 else self.upstream.spack_location  # Clone from local "upstream" repository
             ),
             self.spack_commit,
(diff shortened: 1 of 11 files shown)


michaelmckinsey1:
This would supersede #667, as far as a solution for speeding up `benchpark setup`

lib/benchpark/runtime.py
```diff
@@ -109,23 +109,33 @@ def bootstrap(self):
...
             self.update_old_bootstrap(self.spack_commit, self.spack_location)
 
-    def _install_ramble(self):
+    def _install_ramble(self, bootstrap=False):
         print(f"Cloning Ramble to {self.ramble_location}")
         git_clone_commit(
-            "https://github.com/GoogleCloudPlatform/ramble.git",
+            (
+                "https://github.com/GoogleCloudPlatform/ramble.git"
+                if bootstrap
```
scheibelp: Could we possibly instead allow instantiating `RuntimeResources` with an "upstream" `RuntimeResources`?

We could pass the library `RuntimeResources` in as the upstream for the workspace `RuntimeResources`. That would more-explicitly connect these (and you could collect `upstream.spack_location`, which would account for a case where the library `RuntimeResources` was instantiated differently).

This could also account for the possibility that a user runs `benchpark setup` before `benchpark system`/`benchpark experiment`

michaelmckinsey1:
Tioga CI failure unrelated, will be fixed by #770

scheibelp (changes requested):
Refactoring request: let me know if it makes sense and is agreeable.

lib/benchpark/runtime.py
michaelmckinsey1: Done

remote-urls.yaml
```diff
@@ -0,0 +1,7 @@
+# Copyright 2023 Lawrence Livermore National Security, LLC and other
+# Benchpark Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: Apache-2.0
+urls:
+  ramble: https://github.com/GoogleCloudPlatform/ramble.git
+  spack: https://github.com/spack/spack.git
```
scheibelp, line 7: (question) this seems like it would be useful but I can't immediately think of how - did you have something in mind for this? (I'm fine with it going in either way)

lib/benchpark/runtime.py
```diff
@@ -64,22 +64,32 @@ def __call__(self, *args):
...
-        self.spack_location = self.dest / "spack"
+        # Read remote urls for ramble and spack
+        with open(benchpark.paths.remote_urls, "r") as yaml_file:
+            data = yaml.safe_load(yaml_file)
+            self.ramble_url = data["urls"]["ramble"]
+            self.spack_url = data["urls"]["spack"]
+
+        # If instance does not have an upstream, it is the upstream
+        if self.upstream is None:
+            self.bootstrap()
```
scheibelp: I prefer `.bootstrap()` not being called automatically as part of the `__init__` (usually doing such things makes it harder to test), but a convenience method that does both of these things would work fine.

Also, the comment is true, but I'm not sure how much it conveys to a reader, I think something like

> If this does not have an upstream, then we will be cloning from the URLs indicated in `remote-urls.yaml`

Also question: could this set `self.ramble_url = upstream.ramble_location` here? At that point it wouldn't affect the logic in `_install_ramble`.

Finally, I think this all works assuming that we've always `.bootstrap`ed an upstream before we ever call it (which is true as long as we have always called `benchpark system/experiment init` before `benchpark setup`).

scheibelp (commented):
A couple of requests/questions

remote-urls.yaml
michaelmckinsey1: I think this is a preferable interface if the user wants to change these to point at a different than hardcoding the urls in `lib/`.

lib/benchpark/runtime.py
michaelmckinsey1: > I prefer .bootstrap() not being called automatically as part of the __init__ (usually doing such things makes it harder to test), but a convenience method that does both of these things would work fine.

Ok done. Do you have a specific case in mind? Is it that the trace would not be clear?

> Also, the comment is true, but I'm not sure how much it conveys to a reader, I think something like: If this does not have an upstream, then we will be cloning from the URLs indicated in remote-urls.yaml

Done

> Also question: could this set self.ramble_url = upstream.ramble_location here? At that point it wouldn't affect the logic in _install_ramble.

Yes this is done and I agree it is cleaner.

> Finally, I think this all works assuming that we've always .bootstraped an upstream before we ever call it (which is true as long as we have always called benchpark system/experiment init before benchpark setup).

This implementation actually does not require `system/experiment init` to be ran prior to `benchpark setup`. Running `benchpark setup` without a bootstrap will create the bootstrap and then clone spack and ramble into the benchpark workspace from that bootstrap. The block of imports in `main.py` are responsible for this so any `benchpark ...` will bootstrap. With this in mind, I think it makes more sense to bootstrap in main, since all of the subsequent imports will be redundant anyway.

The above gave me an idea for further refactoring in https://github.com/LLNL/benchpark/pull/822/commits/575f54f1aa4cc125b60cecbf9e6b47f850ce4d19, but this is tangential and will be a separate PR after this one is merged.

scheibelp (approved):

PR #788, opened 2025-05-27
Introducing Hwloc as a new modifier

### LINKED ISSUE #787: Introduce hwloc new modifier

amroakmal:
The goal of this PR is to introduce a new modifier that leverages Hwloc to capture and understand the underlying infrastructure and topology that the HPC job was executed on.

Resolves #787


.gitlab/tests/test.yml
```diff
@@ -39,7 +39,7 @@ workflow:
         ARCHCONFIG: llnl-cluster
         SCHEDULER_PARAMETERS: -N 1 -t 00:20:00 --exclusive --reservation=ci
         BENCHMARK: [kripke]
-        VARIANT: ['+strong~single_node caliper=mpi,time']
+        VARIANT: ['+strong~single_node caliper=mpi,time hwloc=on']
```
amroakmal: The reason behind adding it to this test is because hwloc depends on Caliper to be enabled, so I thought utilizing is better to so to save time to use this one already instead of having a new test so to have less test execution time and resources.

modifiers/hwloc/parse_hwloc_output.py
```diff
@@ -0,0 +1,114 @@
...
+        }
+
+        for key, value in obj.items():
+            if isinstance(value, (str, int, float)):
+                node_info[key] = value
+
+        if "info" in obj and isinstance(obj["info"], list):
+            for item in obj["info"]:
+                k = item.get("@name")
+                v = item.get("@value")
```
amroakmal, line 38: `xmltodict` lib. by default prefix `@` to xml attributes when converting them from xml to json.

modifiers/hwloc/modifier.py
```diff
@@ -0,0 +1,90 @@
...
+                "lstopo-to-get-underlying-infrastructure",
+                template=[
+                    f"lstopo --of xml --whole-system --whole-io --verbose {hwloc_output_xml_file} 2> /dev/null"
+                ],
+            )
+        )
+
+        caliper_modifier = any(
+            [modifier["name"] == "caliper" for modifier in app_inst.modifiers]
+        )
```
amroakmal, line 53: This exact code snippet is duplicated in the `affinity modifier`
I tried adding a `modifiers/utils.py` and have all helpful common funcs our modifiers need in it. The probelm lies in ramble, as importing the utils.py to the modifiers didn't work as ramble has a different working directory path from ours. There were some hacks to solve this but were not clean nor safe at all
I reached out to the ramble team here: [Ramble Issue](https://github.com/GoogleCloudPlatform/ramble/issues/1051) and will check their response. (I just gave this lower priority compared to other goals we are working on)

UPDATE:
 I didn't check using the absolute path (I can try this later, and just came to my mind while writing this comment)

modifiers/hwloc/parse_hwloc_output.py
```diff
@@ -0,0 +1,114 @@
...
+            lines = xml_file.readlines()
+
+        print("lines length", len(lines))
+        # Filter lines that appear to be XML tags or declarations
+        xml_like_lines = [line for line in lines if re.match(r"\s*<[^>]+>", line)]
+
+        if not xml_like_lines:
+            raise ValueError("No valid XML lines found in the file.")
+
+        xml_content = "".join(xml_like_lines)
```
amroakmal, line 84: This is to guard against some lstop (hwloc) warnings/errors that are sometimes  generated.

modifiers/hwloc/parse_hwloc_output.py
```diff
@@ -0,0 +1,114 @@
...
+
+        # Attempt to parse the cleaned XML string
+        data_dict = xmltodict.parse(xml_content)
+
+        parsed_pairs = parse_hwloc_tree_full_metadata(data_dict["topology"]["object"])
+        flat_dict = {path: metadata for path, metadata in parsed_pairs}
```
amroakmal, line 89: The generated json is flattened, to prevent having complex dictionaries as values in the metadata, this was easier for parsing and filtering the thicket.metadata to get needed information.

.gitlab/tests/test.yml
michaelmckinsey1: Made this its own test with affinity, instead of overlapping with the benchpark analyze test ([18b1159](https://github.com/LLNL/benchpark/pull/788/commits/18b115942c26701d37bf85a3b64111a5e14f2a38))

docs/requirements.txt
```diff
@@ -0,0 +1,3 @@
+sphinx
+sphinx_rtd_theme
+sphinxcontrib-programoutput
```
amroakmal: Remove this file, should not be needed

michaelmckinsey1: These are contained in `.github/workflows/requirements/docs.txt`

michaelmckinsey1:
Add standard columns to metadata from the json parser script:
- `X Cache Size` (L1, L2, LLC)
- `Number of Cores`
- etc

slabasan (approved):

PR #803, opened 2025-06-04
Adds an override for the location of the .benchpark directory

ilumsden:
## Description

This feature adds a simple override for the location of the `.benchpark` directory via a `BENCHPARK_HOME` environment variable.

This is a bit of a niche feature, but it can be useful e.g., when trying to install Benchpark in a container. 

- [ ] Update docs (should I really do this? Or should this be kept an internal feature?)


lib/benchpark/paths.py
```diff
@@ -16,7 +16,12 @@ def _source_location() -> pathlib.Path:
 benchpark_root = _source_location()
 lib_path = benchpark_root / "lib" / "benchpark"
 test_path = lib_path / "test"
-benchpark_home = pathlib.Path(os.path.expanduser("~/.benchpark"))
+
+try:
+    benchpark_home = pathlib.Path(os.environ["BENCHPARK_HOME"])
+except KeyError:
+    benchpark_home = pathlib.Path(os.path.expanduser("~/.benchpark"))
```
michaelmckinsey1: Could we not do
```suggestion
# If location is not set, use user home as default location
benchpark_home = pathlib.Path(os.getenv("BENCHPARK_HOME", os.path.expanduser("~/.benchpark")))
```

ilumsden: Good point. I just changed this

michaelmckinsey1 (dismissed):

pearce8 (dismissed):

PR #806, opened 2025-06-05
Prevent Allocation of OS Reserved Cores

michaelmckinsey1:
## Description

- [x] Currently, we create job files that allocate all of the hardware cores on a node, which _may_ include OS reserved cores. These cores will not be available for the job, so it will oversubscribe _n_ cores (n OS reserved cores). This change subtracts that amount from `sys_cores_per_node` and provides a new variable for clarity (so user is not confused why there are less cores), `sys_cores_os_reserved_per_node`. Information was verified by checking affinity on each system.
- [x] Add system information to caliper metadata (if applicable)
```
            "sys_cores_per_node",
            "scheduler",
            "rocm_arch",
            "cuda_arch",
            "sys_cores_os_reserved_per_node",
            "sys_cores_os_reserved_per_node_list",
            "sys_gpus_per_node",
            "sys_mem_per_node",
            "system_site",
``` 

- [x] Update `sys_cores_per_node` and add `sys_cores_os_reserved` in `systems/llnl-*/system.py`.

### DIFF
--- a/lib/benchpark/system.py
+++ b/lib/benchpark/system.py
@@ -43,6 +43,8 @@ class System(ExperimentSystemBase):
         self.external_resources = None
 
         self.sys_cores_per_node = None
+        self.sys_cores_os_reserved = None
+        self.sys_cores_os_reserved_list = None
         self.sys_gpus_per_node = None
         self.sys_mem_per_node = None
         self.scheduler = None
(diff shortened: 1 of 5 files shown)


michaelmckinsey1:
dane failure fixed by #808

pearce8 (approved):

PR #810, opened 2025-06-05
Standard Hwloc Configs

### LINKED ISSUE #809: Provide standard configs from hwloc for pre-defined analysis
The goal of this issue it to provide the necessary standard and essential configs that will be tied to Caliper so that Canned Analysis can consume it easily. We also provide the necessary configs that Flux report.

The standard hwloc metadata configs that will be supported, so far, are as follows:

```
L1Cache.count
L1Cache.cache_size
L1Cache.cache_linesize
L2Cache.count
L2Cahce.cache_size
L2Cache.cache_linesize
L3Cache.count
L3Cahce.cache_size
L3Cache.cache_linesize
NUMANode.count
NUMANode.local_memory [TBD]
Package.count
Core.count
Machine.count
Machine.OSName
Machine.OSRelease
Machine.OSVersion
Machine.HostName
Machine.Architecture
```

amroakmal:
Resolves #809


michaelmckinsey1:
Can we refactor
```
class HwlocVariantValues(str, Enum):
    NONE = "none"
    ON = "on"
```
into
```
    variant(
        "hwloc",
        default=HwlocVariantValues.NONE.value,
        values=tuple(v.value for v in HwlocVariantValues),
        multi=False,
        description="Get underlying infrastructure topology",
    )
```
so it is the similar to the `Affinity` and `SingleNode` classes? I did not catch this in the last PR, but for consistency I think they should work the same.

amroakmal:
@michaelmckinsey1 : Yes, sure. I thought this is cleaner and safer honestly, and more maintainable, so to not miss any value by a typo. Do you think better to stick old behavior? Or better adopt the Enum to other modifiers so all be on consistent and safer to deal with?

Thanks so much for bringing this up.

amroakmal:
Self reminder: Improve variable names of the HWLOC class in experiment.py

pearce8:
@amroakmal What needs to be updated in `benchpark/docs/modifiers.rst`?  While you are updating that, please rename `underlying topology` with `hardware architecture`.  `hardware infrastructure` is also ambiguous, so please also replace it with `hardware architecture`.  From there, try to specify categories such as `memory hierarchy` or `network topology` - lets use textbook names for everything.

michaelmckinsey1:
Waiting to check gitlab tests

pearce8 (approved):

PR #820, opened 2025-06-09
Kripke in mpi-only mode should use 'Sequential' arch

rfhaque:
## Description
This PR adds the 'Sequential' arch value for kripke's mpi-only mode


michaelmckinsey1 (approved):
Tested that this fixes the issue. There is also a pending gitlab pipeline.

slabasan (approved):

PR #825, opened 2025-06-10
Revert "Make oneapi Compiler Default on llnl-cluster"

slabasan:
Reverts LLNL/benchpark#817


slabasan:
More unit tests were failing with the switch to oneapi, needs more debugging. Reverting for now and will open up a new PR once those are resolved.

PR #827, opened 2025-06-11
Make oneapi Compiler Default on llnl-cluster

michaelmckinsey1:
## Description

- [x] Use the oneapi compiler in favor of using gcc as a default on llnl-cluster.
- [x] Depends on #770 

- [x] Update `systems/llnl-cluster/system.py`


rfhaque (dismissed):
@michaelmckinsey1 This can be taken to main

pearce8 (approved):

PR #835, opened 2025-06-16
Features/gromacs

rfhaque:
## Description
This PR updates the implementation of the gromacs experiment in benchpark.
1. The supported version of gromacs has been updated to 2025.2 (latest available in the upstreamed spack packages). 
2. Support for rocm has been added
    - The preferred way to run gromacs on a rocm backend is to build it with AdaptiveCpp/HipSycl (`-DGMX_GPU=SYCL -DGMX_SYCL=ACPP`). However, this only works for amdclang compilers provided with a rocm installation. The hipsycl version being used has been updated to 24.10.0 (latest available in the upstreamed spack packages). With this groamcs now builds with rocm-6.2+ on tuolumne. The rocm build works with and without GPU-aware mpi (benchpark provides a way to enable/disable GPU-aware MPI)
    -  ~Running gromacs on 2 nodes (8 ranks/8 gpus) of tuolumne hangs at the first step of execution that executes `gmx_mpi grompp`. Plainly running `gmx_mpi grompp ...` in a batch job script with 2 nodes seems to hang. If the command is predicated with the job scheduler e.g. `flux run -n1 -N1 gmx_mpi grompp ... `the code runs correctly. Currently ramble doesn't allows predicating different steps within the same experiment with different node/core counts.~ This has been fixed by adding a variable called `single_rank_mpi_command` to gromacs and overriding it in benchpark's allocation modifier. 
    - We can also build with rocm 6+ on tuolumne without hipsycl, using the plain HIP backend introduced in gromacs 2025 (`-DGMX_GPU=HIP`). However at runtime we get a "Chosen GPU implementation does not support constraints" error.
3. Building on lassen (POWER9+NVIDIA V100), requires gcc 11 and cuda 12.2.2 (this is the recommended gromacs configuration for POWER machines) The default clang compiler and cuda 11 do not work on lassen
4. The code also builds and runs on matrix (x86+NVIDIA H100), using cuda 12 and gcc 12, however, cmake 3.28+ required for gromacs is not available on matrix. Building cmake in spack throws and error that is currently circumvented by manually adding additional config details in the generated `ramble.yaml`

The gromacs `package.py` in the benchpak repo is almost identical to the upstreamed version in spack, with only the rocm changes added. The hipsycl package.py is identical to the one in the spack develop branch. Once benchpark is updated to the latest spack, this package.py can be removed.
~The `application.py` in the benchpak repo has been deleted, we now directly use the version upstreamed in ramble.~
We use the same `application.py` as the upstreamed ramble version adding changes to support the `single_rank_mpi_command` command
All new compiler definitions required by gromacs on the target machines have been added to their respective system configs.


pszi1ard:
Thanks for the update!

> However, this only works for amdclang compilers provided with a rocm installation where llvm version < 18. On tuolumne (x86+AMD MI300), all available rocm 6+ versions are llvm 18. 

We don't see issues with newer ROCm on our machines. What issues are you running into with llvm > 18? Is it a spack restriction or some other issue?

> Also running gromacs on 2 nodes (8 ranks/8 gpus) of tuolumne hangs, running on a single node (4 ranks/4 gpus) works. The rocm build works with and without GPU-aware mpi.

Does it hang both with GPU-aware MPI and without?

rfhaque:
> Thanks for the update!
> 
> > However, this only works for amdclang compilers provided with a rocm installation where llvm version < 18. On tuolumne (x86+AMD MI300), all available rocm 6+ versions are llvm 18.
> 
> We don't see issues with newer ROCm on our machines. What issues are you running into with llvm > 18? Is it a spack restriction or some other issue?

@pszi1ard gromacs now works on tuolumne with rocm-6.2.4. The hipsycl package.py had to be also updated to the latest upstreamed spack version.

> > Also running gromacs on 2 nodes (8 ranks/8 gpus) of tuolumne hangs, running on a single node (4 ranks/4 gpus) works. The rocm build works with and without GPU-aware mpi.
> 
> Does it hang both with GPU-aware MPI and without?

It hangs in both cases. I'm trying to figure out if this might have something to do with the tuolumne scheduler rather than gromacs itself.

pszi1ard:
> > Thanks for the update!
> > > However, this only works for amdclang compilers provided with a rocm installation where llvm version < 18. On tuolumne (x86+AMD MI300), all available rocm 6+ versions are llvm 18.
> > 
> > 
> > We don't see issues with newer ROCm on our machines. What issues are you running into with llvm > 18? Is it a spack restriction or some other issue?
> 
> @pszi1ard gromacs now works on tuolumne with rocm-6.2.4. The hipsycl package.py had to be also updated to the latest upstreamed spack version.

OK, good to know.

> > > Also running gromacs on 2 nodes (8 ranks/8 gpus) of tuolumne hangs, running on a single node (4 ranks/4 gpus) works. The rocm build works with and without GPU-aware mpi.
> > 
> > 
> > Does it hang both with GPU-aware MPI and without?
> 
> It hangs in both cases. I'm trying to figure out if this might have something to do with the tuolumne scheduler rather than gromacs itself.

OK. Let me know if you need feedback on that. Just to confirm, do see the same issue on your MI250x machine?

pszi1ard:
> We can also build with rocm 6+ on tuolumne without hipsycl, using the plain HIP backend introduced in gromacs 2025 (-DGMX_GPU=HIP). However at runtime we get a "Chosen GPU implementation does not support constraints" error.

Note that this is incomplete and only supports basic offload of one force computation, no support for GPU-residet mode nor direct GPU communication.

rfhaque:
> > > Thanks for the update!
> > > > However, this only works for amdclang compilers provided with a rocm installation where llvm version < 18. On tuolumne (x86+AMD MI300), all available rocm 6+ versions are llvm 18.
> > > 
> > > 
> > > We don't see issues with newer ROCm on our machines. What issues are you running into with llvm > 18? Is it a spack restriction or some other issue?
> > 
> > 
> > @pszi1ard gromacs now works on tuolumne with rocm-6.2.4. The hipsycl package.py had to be also updated to the latest upstreamed spack version.
> 
> OK, good to know.
> 
> > > > Also running gromacs on 2 nodes (8 ranks/8 gpus) of tuolumne hangs, running on a single node (4 ranks/4 gpus) works. The rocm build works with and without GPU-aware mpi.
> > > 
> > > 
> > > Does it hang both with GPU-aware MPI and without?
> > 
> > 
> > It hangs in both cases. I'm trying to figure out if this might have something to do with the tuolumne scheduler rather than gromacs itself.
> 
> OK. Let me know if you need feedback on that. Just to confirm, do see the same issue on your MI250x machine?

@pszi1ard The problem seems to be with the first step in the experiment that executes `gmx_mpi grompp`. Plainly running `gmx_mpi grompp ...` in a batch job script with 2 nodes seems to hang. If the command is predicated with the scheduler `flux run -n1 -N1 gmx_mpi grompp ...` the code runs correctly. Is there a way to tell gmx_mpi to only run on a single node if no scheduler is specified? @pearce8 Currently I'm not sure if ramble allows predicating different steps within the same experiment with different node/core counts, but that is what is needed in this case. `gmx_mpi` is the MPI version of gromacs, but some steps like `grompp` need to be run on a single node.

repos/spack_repo/benchpark/packages/gromacs/package.py
```diff
@@ -111,16 +157,36 @@ class Gromacs(CMakePackage, CudaPackage, ROCmPackage):
...
     variant("opencl", default=False, description="Enable OpenCL support")
     variant("sycl", default=False, when="@2021:", description="Enable SYCL support")
-    variant("sycl", default=True, when="@2021: +rocm", description="Enable SYCL support when using ROCm")
+
+    with when("+rocm"):
+        depends_on("hip")
+        depends_on("rocfft")
+        with when("+sycl"):
+            depends_on("hipsycl")
+            depends_on(f"hipsycl@24.10.0:", when=f"@2025:")
```
pszi1ard, line 178: this should 23.10 or newer, we don't require 24.10.

modifiers/allocation/modifier.py
```diff
@@ -322,6 +322,7 @@ def slurm_instructions(self, v):
         sbatch_directives = list(f"#SBATCH {x}" for x in (srun_opts + sbatch_opts))
 
         v.mpi_command = f"srun {' '.join(srun_opts)}"
+        v.single_rank_mpi_command = "srun -n 1 -N 1"
```
rfhaque: @scheibelp I've added a single_rank_mpi_command that is used for mpi executables meant to be run on a single node

repo/gromacs/application.py
```diff
@@ -4,212 +4,443 @@
...
+            "adh_dodec",
+        ],
+    )
+
+    workload_variable(
+        "single_rank_mpi_command",
+        default="",
+        description="Job scheduler command for running on a single mpi rank",
+        workload_group="all_workloads",
+    )
```
rfhaque: @scheibelp Correspondingly gromacs defines a `single_rank_mpi_command` variable that is set to an empty string by default. The allocation modifier above overrides this variable with the appropriate string. This setup ensures that the `application.py` works outside of benchpark and the allocation modifier

repo/gromacs/application.py
```diff
@@ -4,212 +4,443 @@
...
+        default="gmx_mpi",
+        description="Name of the gromacs binary",
+        workload_group="all_workloads",
+    )
+    workload_variable(
+        "grompp",
+        default="{single_rank_mpi_command} {gmx} grompp",
+        description="How to run grompp",
+        workload_group="all_workloads",
+    )
```
rfhaque: @scheibelp The `single_rank_mpi_command` variable is prefixed to the gromacs command

repo/gromacs/package.py
```diff
@@ -420,11 +521,23 @@ def setup_run_environment(self, env):
...
+                    self.spec["nvhpc"].version,
+                    "comm_libs",
+                    "nvshmem",
+                    "lib",
+                ),
+            )
+        if "force" in self.spec.variants["gpu-aware-mpi"].value:
+            env.set("GMX_FORCE_GPU_AWARE_MPI", "1")
+        elif "off" in self.spec.variants["gpu-aware-mpi"].value:
+            env.set("GMX_DISABLE_DIRECT_GPU_COMM", "1")
```
pszi1ard: I know the `gpu-aware-mpi` was added earlier, but I wonder why is this needed as a package variant? This is a purely runtime option, the environment variables are only used to guide runtime choices, i.e. 
* `GMX_DISABLE_DIRECT_GPU_COMM` to enable direct GPU communication (which, when built with MPI uses GPU-aware MPI)
*  `GMX_FORCE_GPU_AWARE_MPI` to tell mdrun that even if build-time GPU-aware capability was not possible to detect, assume that it should work (e.g. needed with Cray MPICH)

pszi1ard:
> is there a way to tell gmx_mpi to only run on a single node if no scheduler is specified? @pearce8 Currently I'm not sure if ramble allows predicating different steps within the same experiment with different node/core counts, but that is what is needed in this case. `gmx_mpi` is the MPI version of gromacs, but some steps like `grompp` need to be run on a single node.

@rfhaque I've no idea how to solve that, if you invoke a non-MPI program (e.g. `gmx grompp`) with an MPI wrapper you will just run N instances of the same program, so this is a general concern and it sounds like what @scheibelp (thanks!) is proposing is aimed to tackle that.

repo/gromacs/package.py
rfhaque: @pszi1ard This benchpark variant (on, off, and force are the allowed values) sets the runtime variables `GMX_DISABLE_DIRECT_GPU_COMM` and `GMX_FORCE_GPU_AWARE_MPI` correctly in spack. gromacs cannot detect GPU-aware MPI on tuolumne, so I have to use the `gpu-aware-mpi=force` variant to enable `GMX_FORCE_GPU_AWARE_MPI`

repos/spack_repo/benchpark/packages/gromacs/package.py
rfhaque: 23.10 requires llvm < 18 and that does not work with the rocm 6+ compilers on tuolumne.

repo/gromacs/package.py
pszi1ard: Why does spack need the environment variables set? These are only needed at runtime, not compile-time.

Tow comments:
* the GROMACS nomenclature is "Direct GPU communication" rather than "GPU-aware MPI" as it affects not only MPI-based parallelization. Could we add a synonymous `direct-gpu-comm` variant?
* before GROMACS v2025, even when support is detected direct GPU comm is not default, so `GMX_ENABEL_DIRECT_GPU_COMM` is needed to enable it; starting v2025 the default is changed and `GMX_DISABLE_DIRECT_GPU_COMM` is used to manually disable. I suggest adding corresponding logic.

rfhaque: > Why does spack need the environment variables set? These are only needed at runtime, not compile-time.

Spack is used to set the variables at runtime. Specifically, we set these in the spack function `setup_run_environment` that is invoked by spack at runtime to set the environment variables.

> Tow comments:
> 
>     * the GROMACS nomenclature is "Direct GPU communication" rather than "GPU-aware MPI" as it affects not only MPI-based parallelization. Could we add a synonymous `direct-gpu-comm` variant?

I'll change the variant name. 

>     * before GROMACS v2025, even when support is detected direct GPU comm is not default, so `GMX_ENABEL_DIRECT_GPU_COMM` is needed to enable it; starting v2025 the default is changed and `GMX_DISABLE_DIRECT_GPU_COMM` is used to manually disable. I suggest adding corresponding logic.

I had the pre-2025 logic (default disable) earlier, but I changed to conform to the new logic (default enable). Do we want to support both?

repos/spack_repo/benchpark/packages/gromacs/package.py
pszi1ard: This sounds like a machine-specific limitation.

We should avoid hardcoding requirement that  conflict with official docs, and limit usage of version combinations. The official requirement are:
* with 2023 / 2024 >=0.9.4 (recommended 23.10)
* 2025 >=23.10.

We could simplify a bit and require 23.10 for 2024 and 2025 noting that this is not fully in line with official docs.



I've used 23.10 with llvm >= 18.

rfhaque: I've been consistently getting that error on tuolumne, but let me try that again.

pszi1ard: @rfhaque realted to hipSYCL / ACPP, there is an ACPP feature we collaborated on with their team called "instant submission". This provides major perf boost and it is available from ACPP >=23.10; it is enabled by default from 2025 but for earlier it needs to be enabled manually, see https://manual.gromacs.org/current/release-notes/2025/major/performance.html#instant-submission-mode-enabled-by-default-when-building-with-adaptivecpp

repo/gromacs/package.py
pszi1ard: > Spack is used to set the variables at runtime. Specifically, we set these in the spack function setup_run_environment that is invoked by spack at runtime to set the environment variables.

Thanks, this still sounds confusing, I guess I need to read up how this works. I thought spack's role was build/deploy, but this sounds like runtime concerns are leaking in which seem counter-intuitive.

> I had the pre-2025 logic (default disable) earlier, but I changed to conform to the new logic (default enable). Do we want to support both?

I think it is useful mainly for "reference" runs e.g. when GPU-aware MPI is suspected to be causing issues.

repos/spack_repo/benchpark/packages/gromacs/package.py
pszi1ard: Correction: I have reproduced the issues with ACPP 23.10 and llvm bundled with ROCm 6.2 and later. I am not sure if this is an issue with the ROCm llvm which is not a clean version and sometimes claims conflicting version vs features it supports. Have you been able to try ACPP with upstream llvm >=18?

We will need to update the support matrix of ACPP vs ROCm versions, I have some WIP tests I'll provide more details when I have them.

pszi1ard:
I ran into some build-time issues because the `hipsycl/package.py` requires `llvm` but `llvm-amdgpu` seems to not satisfy this. This leads to further issues when trying to use rocmcc as a compiler.

Looking into this, this brings up multiple ROCm/rocmcc/llvm-related concerns:
* Ideally our default GROMACS recommendation is to use a single ROCm and its bundled llvm for both ACPP/hipsycl and GROMACS -- how can we make `depends_on(llvm...)` be satisfied by `llvm-amdgpu`? 
* While upstream llvm does work as the ACPP compiler library dependency, when we use `rocmcc` for GROMACS, we’d need to ensure we are not mixing llvm versions i.e. if `rocmcc@6.3` is used `llvm@18.x` should be allowed as the ACPP dependency. Can that be implemented?
* the current `hipsycl/package.py` expresses some llvm version dependencies -- these are pessimistic and represent only the limited subset that's been tested by upstream (e.g. v24.02 is known to work with llvm18 at least with those bundles with ROCM <=6.3); verified this withe the author, so some work (verification mainly) will be needed to relax the restrictions.

rfhaque:
@pearce8 We can merge this as-is and fix any issues in a new PR

systems/llnl-cluster/system.py
```diff
@@ -201,8 +201,9 @@ def compute_packages_section(self):
                 },
                 "cmake": {
                     "externals": [
-                        {"spec": "cmake@3.26.5", "prefix": "/usr"},
-                        {"spec": "cmake@3.23.1", "prefix": "/usr/tce"},
+                        {"spec": "cmake@3.26.5", "prefix": "/usr/tce/packages/cmake/cmake-3.26.3"},
+                        {"spec": "cmake@3.23.1", "prefix": "/usr/tce/packages/cmake/cmake-3.23.1"},
+                        {"spec": "cmake@3.31.7", "prefix": "/collab/usr/global/tools/cmake/toss_4_x86_64_ib/cmake-3.31.7"},
```
pearce8: @rfhaque is this change necessary?

modifiers/allocation/modifier.py
```diff
@@ -349,6 +349,7 @@ def slurm_instructions(self, v):
         sbatch_directives = list(f"#SBATCH {x}" for x in (sbatch_opts))
 
         v.mpi_command = f"{launch_cmd} {' '.join(srun_opts)}"
+        v.single_rank_mpi_command = f"srun -n 1 -N 1 {'--gpus 1' if v.n_gpus else ''}"
```
pearce8, line 356: Why is single_rank_mpi_command needed here?

pearce8:
@michaelmckinsey1 Can we turn off the dry runs on this?

michaelmckinsey1:
> @michaelmckinsey1 Can we turn off the dry runs on this?

Disable the dryruns for gromacs? Why when the failure seems like a True positive. The application.py references an executable that is not defined. I don't know how the changes in here make it fail, though. Maybe it is due to ramble being updated recently making the validation more strict.

pearce8 (approved):

michaelmckinsey1:
@pearce8 @rfhaque This changed the default experiment to `water_gmx50`, which uses a ramble executable `execute-gen`, which is not defined in the `application.py`. So the default configuration of the gromacs experiment shouldn't even build, because ramble will fail.

The fix is to define an executable for `execute-gen`, like `execute-nsteps-gen`.

This was caught by the failed dryruns, but the PR was merged.

PR #842, opened 2025-06-17
Gitlab CI Shared Allocation Implementation for Daily Pipelines

michaelmckinsey1:
## Description

- [x] Currently, allocations are granted and released sequentially, which wastes a lot of time between tests and limits the amount of tests we can run. Since jobs can get stuck in the queue, this change can make overall test execution orders of magnitude faster in some cases when the clusters are busy. For a given cluster, we can request one long-running allocation and perform all of the tests within that allocation. The workflow is to:
   1. Request a single-node allocation on a cluster
   2. For each experiment in sequence, attach job to allocation and run test. If one job fails, this is reported separately from the other jobs and the next job continues.
   3. Release the allocation
- [x] This works natively for flux clusters (Tuolumne, Tioga) but not Slurm, since we need to srun within an srun (to "attach" to long running allocation).
   - [x] Tested to be working on Tuolumne and Tioga
   - [x] Clusters such as Ruby and Lassen can still be tested with non-shared allocations
   - [x] Tested to be working on Dane via Flux

Known downsides/issues to be addressed in the future:
- Moved these bullets to #859

- [x] `.gitlab/tests/flux_clusters.yml` contains a matrix for all native flux clusters, so adding tests for a new cluster only requires modifying a few lines.
- [x] `.gitlab/tests/slurm_clusters.yml` is only running dane as of now, and also using flux because of the aforementioned limitations. This is separate from `flux_clusters.yml` because the method of creating the allocation is still different (need to start flux by allocating a node using slurm and then starting flux on the node).
- [x] `.gitlab/tests/non_shared.yml` are the tests for clusters not using a shared allocation `ruby` and `lassen`
- [x] Replace `.gitlab/utils/run-experiment.yml` with `.gitlab/utils/run-experiment.sh` so we can submit the test as a job.
- [x] Replace `.gitlab/tests/test.yml` with the aforementioned refactored yml files


michaelmckinsey1:
laghos failures will be fixed once #770 merged

slabasan (approved):

Issue #336, opened 2024-08-19
Benchpark init


pearce8:
`benchpark init` needs to perform `system init` and `experiment init`.  Followed by `benchpark setup` command.

Later, may want `benchpark setup` to do `benchpark init` at the same time.

PR #849, opened 2025-06-24
System experiment interface

### LINKED ISSUE #724: Post-solve check of provides, requires directives
Example: If the experiment requires cuda, check if the system we are planning to run it on provides cuda.

Post-solve check should work as follows:

1. CudaExperiment would require cuda, and so all of the experiments that derive from CudaExperiment
2. A system with NVIDIA GPUs would provide cuda by deriving from CudaSystem (doesn't exist yet)
3. Notes from Greg: The check itself would be implemented as a post-concretization check on the ``Experiment`` object. The method signature for concretizing an ``ExperimentSpec`` could grow an additional argument to pass the ``SystemSpec`` that it would be used with, and the last step of concretization would be to check for compatibility between the two.

- [ ] Post-solve check
- [ ] After the check is implemented, we need docs to help people understand errors

Depends on #717.

scheibelp:
Fixes https://github.com/LLNL/benchpark/issues/724

An experiment can write:

~`system_depends("rocm", when="+rocm")`~
`requires("rocm", when="+rocm")`

and a system can write

~`system_provides("rocm")`~
`provides("rocm")`

`benchpark experiment init` can now take a `--system` arg (the directory generated by `system init` which:

- gives the experiment access to the system details (e.g. can look at `sys_gpus_per_node`, or ask for the biggest allocation that might reasonably fit on a single node)
- enforces system properties based on `system_depends` in the experiment (so you can make sure that when you create a `+rocm` experiment that you are doing so for a system with `rocm`)

### Example

```
bin/benchpark system init --dest elcap llnl-elcapitan cluster=elcapitan
bin/benchpark experiment init --dest amg2023 --system=elcap amg2023+rocm
```

### Things that could be supported

- ~`system_depends`~ `requires` is conditional, but right now ~`system_provides`~`provides` is not (e.g. rocm-based systems always provide rocm, it might make sense to make this conditional for gtl)
- This could suggest all systems which provide the set of `system_depends` needed for a given experiment configuration (e.g. all the systems that provide rocm)
- The reverse could be possible but would need some more changes: for a given system, configure an experiment "properly" (e.g. for a rocm-enabled system, you probably want to do `+rocm` experiments)

### Related items

- https://github.com/LLNL/benchpark/issues/336


lib/benchpark/system.py
```diff
@@ -78,6 +78,9 @@ def verify(self):
         for pm in self.programming_models:
             pm.verify(self)
 
+    def enforce(self, attr):
+        assert attr in self.system_provides
+
```
michaelmckinsey1:
```suggestion
    def enforce(self, attr):
        if attr not in self.system_provides:
            raise AttributeError(f'"{attr}" not valid for this system. Choose from {self.system_provides}')
        return True
```
Do you think this would be better @scheibelp ?

Current example trying to initialize cuda on rocm system:
```
$ benchpark experiment init --dest amg2023 --system=elcap amg2023+cuda
Traceback (most recent call last):
  File "/Users/mckinsey1/Documents/repos/benchpark/lib/main.py", line 334, in <module>
    exit_code = main()
                ^^^^^^
  File "/Users/mckinsey1/Documents/repos/benchpark/lib/main.py", line 102, in main
    exit_code = action(args)
                ^^^^^^^^^^^^
  File "/Users/mckinsey1/Documents/repos/benchpark/lib/benchpark/cmd/experiment.py", line 68, in command
    actions[args.experiment_subcommand](args)
  File "/Users/mckinsey1/Documents/repos/benchpark/lib/benchpark/cmd/experiment.py", line 40, in experiment_init
    experiment.write_ramble_dict(f"{destdir}/ramble.yaml")
  File "/Users/mckinsey1/Documents/repos/benchpark/lib/benchpark/experiment.py", line 481, in write_ramble_dict
    self.system_spec.system.enforce(need)
  File "/Users/mckinsey1/Documents/repos/benchpark/lib/benchpark/system.py", line 82, in enforce
    assert attr in self.system_provides
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError
Traceback (most recent call last):
  File "/Users/mckinsey1/Documents/repos/benchpark/bin/benchpark", line 23, in <module>
    main()
  File "/Users/mckinsey1/Documents/repos/benchpark/bin/benchpark", line 19, in main
    subprocess.run([sys.executable, main_py] + sys.argv[1:], check=True)
  File "/opt/anaconda3/envs/bp-3.11.11/lib/python3.11/subprocess.py", line 571, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['/opt/anaconda3/envs/bp-3.11.11/bin/python3', PosixPath('/Users/mckinsey1/Documents/repos/benchpark/lib/main.py'), 'experiment', 'init', '--dest', 'amg2023', '--system=elcap', 'amg2023+cuda']' returned non-zero exit status 1.
```

pearce8: yes, an error message will be useful

scheibelp: I have now updated the error message (but I used slightly different wording).

pearce8 (approved):

PR #853, opened 2025-06-26
Add RAJAPerf v2025.03.0

michaelmckinsey1:
## Description

`v2025.03.0` released last week.

PR #854, opened 2025-06-27
Version Benchmarks on Latest Release Dynamically

michaelmckinsey1:
## Description

- [x] Currently we version multiple benchmarks on `develop` instead of the latest release. This strategy enables using the spack mechanism to determine the latest release, only requiring the latest version to be added in the `package.py`. The default version of the experiment will hardcode the latest release version of the benchmark, which will have to be manually updated, but will not change under our feet.
- [x] CI changes depend on #883 
- [x] There is a helper function `determine_version` to enable the `latest` keyword to install the latest release.
- [x] Depends on #853
- [x] Needs to be applied to all experiments using develop:
   - [x] amg2023 has 0 releases - using commit
   - [x] kripke
   - [x] laghos pointing to fork with caliper instrumentation which needs to be upstreamed
   - [x] lammps
   - [x] remhos
   - [x] raja-perf
   - [x] stream
   - [x] bablestream
   - [x] ior
   - [x] osu-micro-benchmarks
   - [x] phloem
   - [x] gpcnet
   - [x] smb

Goals:
1. default hard coded last release, or commit if no release.
2. hardcoded last release version, `latest`, and `develop`.
3. Test both `develop` and last hardcoded release version in CI nightly test


michaelmckinsey1:
1. No `latest` version, only hardcoded last release version and `develop`.
2. Test both versions in CI

michaelmckinsey1:
@rfhaque Do we still want to use `gpu-opt` branch of remhos or use `1.0`?

michaelmckinsey1:
Waiting to check result of nightly tests

michaelmckinsey1:
Broken CI tests

pearce8 (approved):

PR #860, opened 2025-07-01
Add "Machine Up" Checks and Decouple Cluster Workflows

michaelmckinsey1:
## Description

- [x] Add stage dependencies using the `needs` clause, update rules, and refactor gitlab yaml syntax such that jobs complete independently for each cluster. Before these changes, one cluster being down could block the other clusters.
- [x] ~These changes also allow disabling the `resource_group` parameter, which limited jobs to a single runner at a time. Since all we are requesting is non-interactive shell runners that are mainly for submitting the shell script to the allocated node and watch for stdout, they do not use much resources and we can submit jobs in parallel.~ Removing this for now and tracking bug for this feature in #862 
image

~each of the `run_test*` runners are running in parallel, but are non-interactive shell runners on the login node **not** individual batch requests.~

- [x] Fix Dane nightly tests which use slurm (dailys use flux)

- [x] Update `.gitlab/**` for rework

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -33,10 +33,16 @@ include:
     file:
       - utilities/preliminary-ignore-draft-pr.yml
       - utilities/skip-branch-not-a-pr.yml
+  # Machine checks
+  - local: .gitlab/utils/machine_checks.yml
   # Daily tests (from PRs)
   - local: .gitlab/tests/shared_flux_clusters.yml
-  - local: .gitlab/tests/shared_slurm_clusters.yml
-  - local: .gitlab/tests/non_shared.yml
+  #- local: .gitlab/tests/shared_slurm_clusters.yml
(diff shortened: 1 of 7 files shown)


michaelmckinsey1:
~Ready if https://lc.llnl.gov/gitlab/benchpark/benchpark/-/pipelines/561116 and https://lc.llnl.gov/gitlab/benchpark/benchpark/-/pipelines/561119 passing other than CPX, TPX tests
I think I will have to restart these due to current downtime on CZ gitlab~

https://lc.llnl.gov/gitlab/benchpark/benchpark/-/pipelines/561435

michaelmckinsey1:
Appending additional example where tuolumne down 07/02/25. CI working properly
image

michaelmckinsey1:
@slabasan @pearce8  This is ready now

slabasan (approved):

PR #878, opened 2025-07-07
Replace oneapi 2022 with 2023

michaelmckinsey1:
## Description

- [x] There is a hypre build issue (https://lc.llnl.gov/gitlab/benchpark/benchpark/-/jobs/2730071) for oneapi2022, which prevents building amg2023. Using oneapi2023 works for all benchmarks, so changing to this version by default.
- [x] Disallow `oneapi@<=2022` by manual check in package.py to avoid hypre build error

- [x] Update `systems/llnl-cluster/system.py`


michaelmckinsey1:
dane kripke failure fixed by #880, lassen unrelated

pearce8 (approved):

PR #726, opened 2025-04-07
[WIP] Mpi-Pingpong Benchmark

michaelmckinsey1:
## Description

- [x] Add MPI pingpong benchmark to benchpark.
- [x] ~Figure out how to pass location of pingpong directory (since private repo) on command line, instead of hard coding~ Waiting for repo to be public so we can change the url to point at public repo
- [x] Set number of nodes instead of setting total ranks per machine
- [x] Set required variables for benchpark analyze https://github.com/LLNL/benchpark/blob/95b4534919ccb30358376e907115dd5f9d1867c7/experiments/kripke/experiment.py#L141-L145

- [x] Depends on #957

## Adding/modifying a system (docs: [Adding a System](https://software.llnl.gov/benchpark/add-a-system-config.html))

- [x] Update `llnl-*/system.py` with `sys_sockets_per_node` parameter

## Adding/modifying a benchmark (docs: [Adding a Benchmark](https://software.llnl.gov/benchpark/add-a-benchmark.html))

- [x] Add `mpi-pingpong` benchmark
- [x] Update package.py to work with Spack 1.0 @scheibelp -> The package works - michael

PR #888, opened 2025-07-10
Do Not Set CALI_CONFIG if Builtin Caliper

michaelmckinsey1:
## Description

- [x] Benchmarks with built in Caliper configurations do not want to set `CALI_CONFIG`, e.g. RAJAPerf and https://github.com/LLNL/benchpark/pull/726. This change enables the option of adding a tag `builtin-caliper`, which the Caliper modifier will check for and add the config as other environment variables `SPOT_CONFIG` and `OTHER_CALI_CONFIG`, which the application can reference if need be, but won't be used natively by Caliper.

- [x] Update `repo/raja-perf/application.py`

- [x] Update `modifiers/caliper/modifier.py`


michaelmckinsey1:
~Is it possible to read the tags from `repo/raja-perf/application.py` in `modifiers/caliper/modifier.py`? I want to read the `builtin-caliper` tag from specific applications. Currently I am writing a file in place of this functionality.~

pearce8 (approved):

PR #898, opened 2025-07-14
Analyze: Handle Case Where Name is None and New Intersection Option

michaelmckinsey1:
## Description

- [x] Currently, if profiles contain any different regions `benchpark analyze` will fail with error:
```
    return lib.map_infer(values, mapper, convert=convert)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "pandas/_libs/lib.pyx", line 2999, in pandas._libs.lib.map_infer
  File "/Users/mckinsey1/Documents/repos/benchpark/lib/benchpark/cmd/analyze.py", line 232, in <lambda>
    ".", lambda row: row["name"].apply(lambda n: "MPI_" not in n).all()
                                                 ^^^^^^^^^^^^^^^
TypeError: argument of type 'NoneType' is not iterable
```
because of the check in the Thicket Query. Modifiying the check to not check against `None` fixes this issue
- [x] Additionally, add the option to take the `intersection` of the calltree instead of the `union` which is an alternative way to address this issue, but also a feature.

- [x] Update `lib/benchpark/cmd/analyze.py


michaelmckinsey1:
~Make sure gitlab unit test for analyze passing (good if all dane tests passing)~ ready

pearce8 (approved):

PR #904, opened 2025-07-18
Automatically Enumerate Dryruns and MpiOnlyExperiment

michaelmckinsey1:
## Description

- [x] Supersedes #701. The ability to run newly added experiment.py's in CI without changes to any CI files (.github), using `benchpark list` which will automatically execute dryrun testing for:
   - [x] New benchmarks.
   - [x] New experiments on existing benchmarks
   - [x] New modifiers
   - [x] New systems
- [x] Introduces new management class for experiments to inherit from `MpiOnlyExperiment` if running the experiment without a programming model is not valid, e.g. `babelstream`. This will return a descriptive error during experiment initialization, instead of the user getting a confusing error during build/run:
```
$ benchpark experiment init --dest babelstream babelstream
Traceback (most recent call last):
  File "/Users/mckinsey1/Documents/repos/benchpark/lib/main.py", line 336, in <module>
    exit_code = main()
                ^^^^^^
  File "/Users/mckinsey1/Documents/repos/benchpark/lib/main.py", line 104, in main
    exit_code = action(args)
                ^^^^^^^^^^^^
  File "/Users/mckinsey1/Documents/repos/benchpark/lib/benchpark/cmd/experiment.py", line 70, in command
    actions[args.experiment_subcommand](args)
  File "/Users/mckinsey1/Documents/repos/benchpark/lib/benchpark/cmd/experiment.py", line 20, in experiment_init
    experiment = experiment_spec.experiment
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/mckinsey1/Documents/repos/benchpark/lib/benchpark/spec.py", line 380, in experiment
    return self.experiment_class(self)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/mckinsey1/Documents/repos/benchpark/lib/benchpark/experiment.py", line 276, in __init__
    raise NotImplementedError(
NotImplementedError: "babelstream" cannot run with MPI only without inheriting from MpiOnlyExperiment. Choose from ['openmp', 'rocm', 'cuda']
```
- [x] depends on #903 
- [x] depends on #921
- Open question on whether it is necessary to enumerate variants like `compiler`. These only result in branching, which outside of building we can only check for syntax.
- [x] Improve logs for failing dryruns

## Adding/modifying core functionality, CI, or documentation:

- [x] Update `.github`
- [x] Update dryrun docs
- [x] Update github dryrun developer docs once #892 merged

PR #931, opened 2025-07-29
Enable Initializing Second Experiment From Library

### LINKED ISSUE #932: Cannot initialize second experiment from library
- [x] Tracking in #931
 
- There was an issue discovered in #904, where trying to initialize an experiment other than `saxpy` (which is used in every current pytest), will take the form:
```
$ benchpark unit-test
_________________________________________________________________________________ test_programming_model_checks __________________________________________________________________________________

    def test_programming_model_checks():
        # babelstream mpi-only not valid
        with pytest.raises(NotImplementedError, match="cannot run with MPI only"):
            spec = benchpark.spec.ExperimentSpec("babelstream").concretize()
>           experiment = spec.experiment  # noqa: F841

lib/benchpark/test/experiment_errors.py:15: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
lib/benchpark/spec.py:380: in experiment
    return self.experiment_class(self)
lib/benchpark/experiment.py:225: in __init__
    super().__init__()
lib/benchpark/cuda.py:16: in __init__
    super().__init__()
lib/benchpark/rocm.py:16: in __init__
    super().__init__()
lib/benchpark/openmp.py:17: in __init__
    if self.spec.variants["openmp"][0]:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <benchpark.spec.ConcreteVariantMap object at 0x14a4e9590>, key = 'openmp'

    def __getitem__(self, key):
>       return self.dict[key]
E       KeyError: 'openmp'

../../../.benchpark/ramble/lib/ramble/llnl/util/lang.py:495: KeyError
==================================================================================== short test summary info =====================================================================================
FAILED lib/benchpark/test/experiment_errors.py::test_programming_model_checks - KeyError: 'openmp'
```
We know the issue is with multiple initialization because running the test on its own works, i.e. `$ benchpark unit-test lib/benchpark/test/experiment_errors.py`

The issue seems to stem from the `variant` directive not working correctly after the first initialization. This behavior does not affect normal benchpark use, as initializing experiments from the command line will never cause this issue.

michaelmckinsey1:
## Description

- [x] Refactor programming models to be defined like scaling options are. I noticed while testing that the scaling variants were not affected from the same bug as the programming models. I.e. this is an issue with importing the programming class multiple times, which results in weird behavior with directives. This change avoids this.
- [x] Depends on #904 
- [x] Issue described in #932
- [x] ~This can be worked around by running the pytests individually, but this complicates the code coverage.~
   - [x] Example of error code checking on command line tests here https://github.com/LLNL/benchpark/pull/982/
files#diff-a3ddd7238fc6e36daeb0ef76e93fe15cc87824bd3299888075210c5ca6a474d3 
   - [x] See removed tests in .github/workflows/run.yml -> moved to pytest now with this fix. Much cleaner tests in pytests.
- [x] Refactor all experiments
- [x] rename `models.py` to `programming_model.py`
- [x] Update docs/tutorial


.github/workflows/run.yml
```diff
@@ -5,43 +5,6 @@ on:
...
-          # 2) Multiple scaling options
-          set +e
-          # Cannot use both +strong and +weak
-          stderr_output="$({ ./bin/benchpark experiment init --dest kripke --system dane kripke+strong+weak; } 2>&1)"
-          status=$?
-          set -e
-          [[ $status -ne 0 ]] || fail "Expected non-zero exit status, got $status"
-          # Check for the specific BenchparkError message
-          expected="spec cannot specify multiple scaling options"
-          grep -Fq "$expected" <<<"$stderr_output" && echo "failed as expected: $expected" || fail "Expected error message not found in stderr. Got: $stderr_output"
```
michaelmckinsey1, line 44: These tests can now be moved to lib/benchpark/test/experiment_errors.py, which shows how much more concise the pytest format is

michaelmckinsey1:
Ready for a review, but definitely not merging because only 2 of the experiments are updated for this change

lib/benchpark/programming_model.py
```diff
@@ -0,0 +1,85 @@
+# Copyright 2023 Lawrence Livermore National Security, LLC and other
```
scheibelp, line 1: I think this module should be called `experiment_models`, or live in an an `experiments` subdir

lib/benchpark/programming_model.py
```diff
@@ -0,0 +1,85 @@
...
+    class BaseModel:
+        requires("mpi", when="+mpi")
+        requires("rocm", when="+rocm")
+        requires("cuda", when="+cuda")
+        requires("openmp", when="+openmp")
+
+        variant("mpi", default=True, description="Run with MPI")
+        variant("rocm", default=False, description="Build and run with ROCm")
+        variant("cuda", default=False, description="Build and run with CUDA")
+        variant("openmp", default=False, description="Build and run with OpenMP")
```
scheibelp, line 37: It looks like these variants are always defined on every experiment, regardless of whether it inherits the associated model types.

It should be allowed to split these variants into separate modules and imported classes (as they were originally); at the same time, customizing it all from one file is ok too, I wonder if it's possible to address the above concern though (it might not be possible to conditionally define the variants).

scheibelp (commented):
I have a question and a request

lib/benchpark/programming_model.py
michaelmckinsey1: This is true, regardless of which models are defined, if the experiment inherits from `Models(...)` all programming models will show in the spec `+mpi~rocm~cuda~openmp`. I would like to note that this is the same behavior for the Scaling, e.g. even if the experiment does not define Throughput scaling it will show `~throughput` in the spec. Not that this is correct either, but if we can find a solution for this it should be applied to both the programming models and scaling.

I don't know of a solution to the underlying bug that would enable separate classes. Perhaps something like
```
# mpi.py
def Mpi():
  class BaseMPI:
       variant("mpi")
```

```
# rocm.py
def Rocm():
  class BaseRocm:
       variant("rocm")
```

would provide the correct behavior, but the design seems maybe a bit convoluted.

It seems like the relevant piece for the behavior we desire is making sure the variant is not applied when the class is imported.

lib/benchpark/programming_model.py
michaelmckinsey1: Ok I will rename this to `experiment_models`

experiments/kripke/experiment.py
```diff
@@ -5,20 +5,14 @@
 
 from benchpark.directives import variant, maintainers
 from benchpark.experiment import Experiment
-from benchpark.mpi import MpiOnlyExperiment
-from benchpark.openmp import OpenMPExperiment
-from benchpark.cuda import CudaExperiment
-from benchpark.rocm import ROCmExperiment
+from benchpark.models import ModelsType, Models
```
michaelmckinsey1: Feedback from @pearce8 :

`benchpark.models` -> `benchpark.programming_model`

`ModelsType` -> `ProgrammingModelType`
`Models` -> `ProgrammingModel`

lib/benchpark/programming_model.py
michaelmckinsey1: https://github.com/llnl/benchpark/pull/931#discussion_r2842773812

repos/ramble_applications/commbench/application.py
```diff
@@ -14,7 +14,8 @@ class Commbench(ExecutableApplication):
 
     maintainers("arhag23")
 
-    software_spec("commbench", pkg_spec="commbench", package_manager="spack*")
+    with when("package_manager_family=spack"):
+        software_spec("commbench", pkg_spec="commbench")
```
michaelmckinsey1, line 18: These application changes are due to `package_manager` being deprecated, which will generate a warning and that actually errors in the new unit tests `test/experiment_errors.py`. So I went ahead and deprecated the functionality to avoid the error.

lib/main.py
```diff
@@ -374,7 +374,7 @@ def benchpark_tags_handler(args):
     """
     source_dir = paths.benchpark_root
     ramble_exe = paths.benchpark_home / "ramble/bin/ramble"
-    subprocess.run([ramble_exe, "repo", "add", "--scope=site", f"{source_dir}/repo"])
+    subprocess.run([ramble_exe, "repo", "add", "--scope=site", f"{source_dir}/repos/ramble_applications"])
```
michaelmckinsey1: This is something we missed when we refactored ramble and spack repos

lib/benchpark/experiment.py
```diff
@@ -255,32 +255,35 @@ def __init__(self, spec):
...
+            raise BenchparkError(
                 f'{invalid_models} are not valid programming models for "{self.name}". Choose from {self.programming_models}.'
             )
-        # Check if experiment is trying to run in MpiOnly mode without being an MpiOnlyExperiment
-        elif "mpi" not in str(self.spec) and not any(
-            self.spec.satisfies("+" + model) for model in models[1:]
-        ):
-            raise NotImplementedError(
-                f'"{self.name}" cannot run with MPI only without inheriting from MpiOnlyExperiment. Choose from {self.programming_models}'
-            )
```
michaelmckinsey1, line 283: This check is no longer applicable now that `mpi` is always in the spec. The case this check is trying to catch is now caught by the above check, as `mpi` will be an invalid model. See the new babelstream unit test for an example

pearce8 (approved):

scheibelp:
@michaelmckinsey1 

[lib/benchpark/rocm.py‎](https://github.com/llnl/benchpark/pull/931/changes#diff-47d028275a66adc292a1f510f3482f119c3f068d45f082ecab13a268c16b040b)

had 

```
        def get_spack_variants(self):
            return (
                "+rocm amdgpu_target={rocm_arch}"
                if self.spec.satisfies("+rocm")
                else "~rocm"
            )
```

Where did this move to? I feel like it was dropped when looking through these changes

michaelmckinsey1:
> @michaelmckinsey1
> 
> [lib/benchpark/rocm.py‎](https://github.com/llnl/benchpark/pull/931/changes#diff-47d028275a66adc292a1f510f3482f119c3f068d45f082ecab13a268c16b040b)
> 
> had
> 
> ```
>         def get_spack_variants(self):
>             return (
>                 "+rocm amdgpu_target={rocm_arch}"
>                 if self.spec.satisfies("+rocm")
>                 else "~rocm"
>             )
> ```
> 
> Where did this move to? I feel like it was dropped when looking through these changes

It looks like it did get dropped

PR #942, opened 2025-07-31
Adding More Hardware Information to LLNL Systems

### LINKED ISSUE #937: Add memory info to the systems
- [x] Test to make sure they get stored in caliper metadata after experiment runs - [look here](https://github.com/LLNL/benchpark/blob/fc3e74fc2200d2e3c20606bc3b5fa3b7754d9dd2/modifiers/caliper/modifier.py#L120-L130).
- [ ] What can you automatically extract from hwloc info?  Start with @amroakmal's script and write a helper script
- [ ] There may be more information in /etc/node_info.json on some of the LC machines

Units and how to organize the data
- [ ] Use the same units as hwloc, add them in the name of the variable, e.g., `sys_cpu_L1_KB_per_core`, `sys_mem_GB_per_node` 
- [ ] check that the units make sense for the cheapest/smallest CPU on AWS (we can't do fractions)
- [ ] If hwloc is reporting X L1s per node, we should also - so update our variables to `sys_cpu_L1_KB_per_core` (we already have `sys_num_cpu_cores`) and ` 

Types of counters
- [x] sys_mem_GB_per_node 
- [x] sys_cpu_mem_GB_per_node (same as sys_mem_per_node in CPU-only)
- [x] sys_gpu_mem_GB_per_node (if separate from CPU memory)
- [x] sys_cpu_L1_KB_per_core
- [x] sys_gpu_L1_KB_per_device
- [x] sys_gpu_num_L1_per_node
- [x] sys_cpu_L2_MB_per_node
- [x] sys_gpu_L2_MB_per_node
- [ ] sys_gpu_num_L2 
- [x] sys_cpu_L3_MB_per_node
- [x] sys_gpu_L3_MB_per_node
- [ ] sys_gpu_num_L3 
- [ ] sys_num_nodes_per_rack
- [ ] sys_num_racks
- [ ] sys_network_topology (what is available in hwloc on this?)

If you are finding that the counter names are not a good match for what we have on CPU clusters *and* on El Cap,  lets discuss here

stephanielam3211:
## Description

- [x] Added system memory information for dane, ruby, lassen, tioga, and tuolumne (el capitan).
- [x] Rename `sys_mem_per_node` to `sys_mem_per_node_GB`

- [x] Review the list of variables
- [x] Add/modify `systems/llnl-cluster/system.py`, `systems/llnl-elcapitan/system.py`, and `sysytems/llnl-sierra/system.py` files


systems/llnl-cluster/system.py
```diff
@@ -19,6 +19,12 @@ class LlnlCluster(System):
             "sys_cores_per_node": 56,
             "sys_cores_os_reserved_per_node": 0,  # No core or thread reservation
             "sys_cores_os_reserved_per_node_list": None,
+            "sys_mem_per_node": 94, #either 94GB or ~94981-96752MB
+            "sys_cpu_mem_per_node": 94,
+            "sys_cpu_L1_size": 32, #32KB for L1d and 32KB for L1i
+            "sys_cpu_L2_size": 1024,
+            "sys_cpu_L3_size": 39, #105MB
```
michaelmckinsey1: Should this be 39 or 105?

systems/llnl-elcapitan/system.py
```diff
@@ -46,6 +55,18 @@ class LlnlElcapitan(System):
                 88,
             ],  # 3 cores reserved per socket
             "sys_gpus_per_node": None,  # Determined by "gpumode" variant
+            "sys_sockets_per_node": 4,
+            "sys_mem_per_node": 1024, #either 125GB or ~128997MB
```
michaelmckinsey1: Is this comment incorrect?

systems/llnl-elcapitan/system.py
```diff
@@ -22,6 +22,15 @@ class LlnlElcapitan(System):
...
             "sys_cores_os_reserved_per_node_list": [0, 8, 16, 24, 32, 40, 48, 56],
             "sys_gpus_per_node": 8,
+            "sys_mem_per_node": 125, #either 125GB or ~128997MB
+            "sys_cpu_mem_per_node": 125,
+            "sys_gpu_mem_per_node": 68702699520, #in B
+            "sys_cpu_L1_size": 32, #32KB for L1d and 32KB for L1i
+            "sys_gpu_L1_size": 16,
+            "sys_gpu_num_L1": 1008,
+            "sys_cpu_L2_size": 512, #512 KB
+            "sys_gpu_L2_size": 8192,
```
michaelmckinsey1, line 32: The units are KB here but MB for Tuolumne

systems/llnl-sierra/system.py
```diff
@@ -26,6 +26,15 @@ class LlnlSierra(System):
...
             ],  # First two cores on each socket reserved.
             "sys_gpus_per_node": 4,
+            "sys_mem_per_node": 256, #either 256GB or ~128997MB
+            "sys_cpu_mem_per_node": 256,
+            "sys_gpu_mem_per_node": 16384, #in MiB Fb Memory?
+            "sys_cpu_L1_size": 32, #32KB for L1d and 32KB for L1i
+            "sys_gpu_L1_size": 16, #not done
+            "sys_gpu_num_L1": 1008, #not done
+            "sys_cpu_L2_size": 512, #512 KB
+            "sys_gpu_L2_size": 8192, #not done
```
michaelmckinsey1: Are these complete?

michaelmckinsey1 (changes requested):
Main issues:
1. Instead of `_size` in the variable names use `_KB`, `_MB`, `_GB`, should be standard across systems
2. `llnl-sierra` seems incomplete

systems/llnl-cluster/system.py
stephanielam3211: 105 i guess so it matches in terms of MB

systems/llnl-elcapitan/system.py
stephanielam3211: incorrect comment

systems/llnl-sierra/system.py
stephanielam3211: I don't remember, let me check

systems/llnl-cluster/system.py
michaelmckinsey1: Actually it was right https://www.intel.com/content/www/us/en/products/sku/192475/intel-xeon-platinum-8276l-processor-38-5m-cache-2-20-ghz/specifications.html. I changed it back to 38.5

michaelmckinsey1 (changes requested):
I reviewed/fixed some values and naming conventions. I think we need to decide whether we are proceeding with trying to automate populating these values with hwloc or not

michaelmckinsey1 (approved):

pearce8 (approved):

PR #944, opened 2025-07-31
Test Using system-aware Experiment Initialization and add MPISystem

### LINKED ISSUE #928: Switch dry runs and CI tests to use ``experiment init`` with ``--system=elcap``
- [ ] Depends on #849

michaelmckinsey1:
## Description

- [x] Use system-aware experiment initialization in GitHub and GitLab CI
- [x] Add `MPISystem` to provide `mpi` for enforcement check in `system.py`.
- [x] Fix issue where a system could only provide one type of prog model. 
- [x] Fixes #928

- [x] Move `OpenMPSystem` from multiple `system.py` to system base class 

- [x] Update `lib/benchpark/system.py` to add `MPISystem`
- [x] Update `lib/benchpark/directives` to have provides accept multiple values.
- [x] Update `dryrun.sh` and `run-experiment.sh` CI files


lib/benchpark/system.py
```diff
@@ -43,6 +53,9 @@ def __init__(self, spec):
         self.timeout = "120"
         self.queue = None
 
+        # Assume every system is an MPI/OpenMP system
+        self._programming_models = [MPISystem(), OpenMPSystem()]
```
michaelmckinsey1: @pearce8 I know we assume MPI, but is there an example of a system which **is not** also an `OpenMPSystem`? Currently we only mark a couple of systems as `OpenMPSystem` and do not mark systems like llnl-elcapitan

lib/benchpark/cmd/experiment.py
```diff
@@ -57,7 +57,9 @@ def setup_parser(root_parser):
     init_parser.add_argument(
         "--basedir", help="Generate a system dir under this, and place all files there"
     )
-    init_parser.add_argument("--system", help="The system this will be run on")
+    init_parser.add_argument(
+        "--system", help="The system this will be run on", required=True
+    )
```
michaelmckinsey1: Maybe we don't do this yet, but this would be a simple way to enforce the use of system-aware experiment init

michaelmckinsey1: I removed the change, just wanted to show it

michaelmckinsey1:
Ready for review if passing CI

lib/benchpark/system.py
pearce8: We might need 2 types of OpenMP systems:

- [ ] OpenMPSystem - with OpenMP that works on the CPU.  We should not make all systems by default OpenMPSystems because it depends on whether there is an OpenMP compiler.
- [ ] OpenMPOffloadSystem - if systems with GPUs have a compiler which is able to offload work to the GPU using an OpenMP directive.  Again, only add this if it indeed works - currently, we would only ever do that with Fortran, so we do not have any use cases for this.

pearce8 (changes requested):
@michaelmckinsey1 If we are going to make all systems OpenMPSystem, we need to add a short OpenMP test on each system, something like Stream.

lib/benchpark/system.py
michaelmckinsey1: Maybe name `OpenMPCPUOnlySystem` to be clear we are not doing OpenMP offload (if we are going that route)

michaelmckinsey1:
> @michaelmckinsey1 If we are going to make all systems OpenMPSystem, we need to add a short OpenMP test on each system, something like Stream.

todo: I need to figure out best way to add openmp tests **across systems** for gitlab (LC systems)

Edit: [8695b48](https://github.com/LLNL/benchpark/pull/944/commits/8695b48166d74557c3f9c0e0d88eeae96076444a) Although unsure of the value of testing openmp on these systems aside from Dane

lib/benchpark/system.py
michaelmckinsey1: [befd1f3](https://github.com/LLNL/benchpark/pull/944/commits/befd1f34372d5987e0dabdbe8844e70d62104cf3)

pearce8 (approved):

PR #945, opened 2025-08-04
[WIP] Use Enums to Enforce Syntax

amroakmal:
## Description

Currently checks such as `self.spec.satisfies("affinity=none"):` will not error if there is a spelling/syntax mistake within the argument, e.g. `self.spec.satisfies("affinity=non"):` (missing "e"). This will of course not be a syntax error, and the expected usage from the command line will not satisfy this condition. The changes in this PR uses Enums to avoid these programming mistakes, which are potentially difficult to debug.

- [x] Refactor `lib/benchpark/experiment.py`

### DIFF
--- a/lib/benchpark/experiment.py
+++ b/lib/benchpark/experiment.py
@@ -85,14 +85,38 @@ class SingleNode:
             return "single_node" if self.spec.satisfies("+single_node") else ""
 
 
+class AffinityVariantValues(str, Enum):
+    AFFINITY = "affinity"
+    NAME = "name"
+    MODE = "mode"
+    PKG_SPEC = "pkg_spec"
+    COMPILER = "compiler"
+
+    NONE = "none"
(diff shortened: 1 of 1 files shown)


michaelmckinsey1 (changes requested):
From @scheibelp :
1. Import respective enums from class (cuda enum should come from cuda.py)
2. make variable names more succinct where possible

lib/benchpark/experiment.py
```diff
@@ -85,31 +85,59 @@ def get_helper_name_prefix(self):
             return "single_node" if self.spec.satisfies("+single_node") else ""
 
 
+class AffinityVariantValues(str, Enum):
```
michaelmckinsey1, line 88: I think this could be `AffinityEnum` and we could make a base enum class for the `tuple_values` methods etc. that are shared across enums

PR #949, opened 2025-08-05
Debug, Bank, and Timeout as Variants

### LINKED ISSUE #946: Expose Setting `bank` as Parameter for Scheduler
Set bank for Flux or account for Slurm

michaelmckinsey1:
## Description

- [x] Add the following options available to all systems:
   - [x] `timeout=X` job timeout
   - [x] `bank=X` set account/bank
   - [x] `queue=pdebug` turn on debug mode
e.g. `benchpark experiment init ... timeout=60 bank=fractale job_queue=pdebug`
- [x] Add docs
.
- [x] Add allocation modifier check that number of requested nodes is under limit for selected queue `==> Error: 200 nodes is unsatisfiable for queue 'pdebug' (max 24).`
   - [ ] unit test? non-trivial for modifier 
- [x] check at `system init` that timeout is less than max limit for queue `ValueError: timeout=721 is unsatisfiable for the selected queue 'pdebug'. Max timeout is 720`
   - [x] Add unit test 

@michaelmckinsey1 question: What does debug mode do?  Do you mean submit to debug queue?  

@michaelmckinsey1 Changes requested, and additional tasks:

- [x] The system should have a `queue` variant, and whoever is writing the system.py can specify available queues.  
- [x] For each queue, need to specify `max time limit`, and `max number of nodes` allowed to allocate
- [x] Job timeout needs to be no greater than the queue limit for the specific queue -  it would be good to check it at experiment --system initiation time.  Not sure if this means queues need to be directives - let's discuss on Monday.
- [x] `bank`s should also be per system, and therefore variants on the system.  A user will then be able to use a subset of the banks.
- [x] Add queues, time limits, and banks for llnl systems


michaelmckinsey1:
change debug to `job_queue` that takes wildcard argument (for us `pdebug`)

pearce8 (changes requested):
See PR description for question and changes requested.

michaelmckinsey1:
@pearce8 banks/queues added for LC systems. Other systems will not set bank and queue (so use the default values)

pearce8 (approved):

PR #953, opened 2025-08-06
Spack 1 update

### LINKED ISSUE #918: Update to Spack 1.0
- [ ] Please push as a branch to our repo
- [ ] Shouldn't need the newest Ramble
- [ ] Handling compilers in system.py: move from compilers section to packages section.  Added a helper method which merges dictionaries
- [ ] Right now just testing generating configurations
- [ ] CSC Eigen does modules, may be broken
- [ ] What needs to be changed in benchpark functionality
- [ ] Which packages currently in benchpark need to also be updated?
- [ ] What additional changes?

scheibelp:
Fixes #918 

- [x] Make a list of systems which need their owners to insert paths @scheibelp  
  - See: https://github.com/LLNL/benchpark/pull/953#issuecomment-3268741193
- [x] Make a list of systems which have not been tested @scheibelp  
  - See: https://github.com/LLNL/benchpark/pull/953#issuecomment-3268741193
- [x] Check if Fortran is needed in any benchmarks alphabetically after GPCNet @nhanford 
- [x] Send a message to benchpark-support when this is merged @pearce8 

As of September 3rd this is now passing all pipeline benchmark builds/executions, and is ready for review.

At a high level:

* there is no more compilers section: the compilers must be expressed as packages
  * So every system.py has fairly extensive changes
    * but minimal changes to API:
      * no new functions (except helper functions for making confg dicts): `compute_compilers_section` is still responsible for describing compilers, but it produces a `packages:` config
      * There is a new helper function `merge_dicts` to recursively merge two configs (it works like Spack's config merge logic, and perhaps could be replaced with that, but for now entirely avoids any import needs and is ~20 lines)
        * This makes it easy for the base `system.py` to merge compiler config with the output of `compute_packages_section`)
* packages must declare language dependencies (c/c++/fortran)
  * so every package.py changes
    * A few of them did not need to add these because they inherit from builtin (in those cases, the way we reference packages in builtin changed, so those packages changed also)
  * in several cases I'm probably adding an unnecessary `fortran` dependency: in alphabetical order I've somewhat-carefully checked everything up to `gpcnet`
  * Assuming the concretization completes without issue, I generally expect it is not much of a problem to incorrectly add a fortran dependency

### Packages with additional changes

(all the other package changes follow the general description above)

* `hypre` I disabled a check entirely that appears to not ever be activated (and accessing the compiler attributes changed in Spack 1.0)
* `raja-perf`: there used to be a separate `rocmcc` and `llvm-amdgpu` concepts, but they merged: `llvm-amdgpu` is now the compiler, and library resources sit one directory lower in `llvm`

### Unfinished systems

See: https://github.com/LLNL/benchpark/pull/953#issuecomment-3268741193

### Follow-up items

* This PR modifies some experiments to add `+mpi` to the benchmark spec and to also influence how some dependencies are built. It would be good to consolidate these choices if possible (to help avoid other benchmarks "randomly" concretizing incorrectly, we should potentially be applying these constraints more consistently).


experiments/kripke/experiment.py
```diff
@@ -132,4 +132,4 @@ def compute_package_section(self):
             if self.spec.variants["single_memory"][0]
             else "~single_memory"
         )
-        self.add_package_spec(self.name, [f"kripke@{app_version} {single_memory} "])
+        self.add_package_spec(self.name, [f"kripke@{app_version} {single_memory} +mpi"])
```
michaelmckinsey1: ~`+mpi` can be removed as of https://github.com/LLNL/benchpark/pull/904~

michaelmckinsey1:
https://github.com/LLNL/benchpark/pull/953/files/fbfb1168eead0b3c049380f9b3a1eee0ddb98ca5..844b9007502ba1882c0086b7ef500103faf3aa94

I'm not sure why this only started failing recently again. I think the runner hashes may have gotten longer. But If this fixes it then that works.

pearce8:
@scheibelp How do you determine which languages need to be on this list for each package?

```
    depends_on("c", type="build")
    depends_on("cxx", type="build")
    depends_on("fortran", type="build")
```

lib/benchpark/runtime.py
```diff
@@ -68,29 +68,43 @@ def __init__(self, dest, upstream=None):
         self.dest = pathlib.Path(dest)
         self.upstream = upstream
 
-        self.ramble_location, self.spack_location = (
+        self.ramble_location, self.spack_location, self.pkgs_location = (
             self.dest / "ramble",
-            self.dest / "spack",
+            self.dest / "spk",
```
michaelmckinsey1: We should do https://github.com/LLNL/benchpark/pull/1009 to avoid needing to change the spack name

michaelmckinsey1 (dismissed):

repo/hypre/package.py
```diff
@@ -21,8 +21,7 @@ class Hypre(BuiltinHypre):
     def configure_args(self):
         configure_args = super().configure_args()
 
-        if (self.compiler.fc and 'xlf' in self.compiler.fc) or (self.compiler.f77 and 'xlf' in self.compiler.f77):
-          if not "+fortran" in self.spec:        
+        if False:
```
pearce8: Is this correct?  Or needs a different fix?

systems/llnl-elcapitan/system.py
```diff
@@ -625,15 +619,15 @@ def rocm_config(self):
...
-                "llvm-amdgpu": {
-                    "externals": [
-                        {
-                            "spec": f"llvm-amdgpu@{self.rocm_version}",
-                            "prefix": f"/opt/rocm-{self.rocm_version}/llvm",
-                        }
-                    ],
-                    "buildable": False,
-                },
+                # "llvm-amdgpu": {
```
pearce8: So how do we handle llvm on el cap?

scheibelp: This bit can be deleted: the compiler config now adds an `llvm-amdgpu` definition (and the `llvm-amdgpu` package was only used when compiling with `rocmcc`).

One important caveat with merging the package/compiler definitions is that there is only one prefix, so packages that need to access the `llvm` subdir need to change (`raja-perf` appears to be the only package that needs this).

scheibelp:
> How do you determine which languages need to be on this list for each package?

I pulled the source code and looked for c/c++/fortran files. It's generally fine if one is added and not needed (if one is needed and not present, there will generally be an error). I started going through packages alphabetically to verify and got as far as `gpcnet` (for everything after, I think I just added all 3 dependencies, and that seems to work for all the benchmark tests that are run).

experiments/kripke/experiment.py
michaelmckinsey1: Correction. We need +mpi here because https://github.com/LLNL/benchpark/blob/develop/lib/benchpark/mpi.py helper does not have `get_spack_variants` function, which we don't want as we may not always want +mpi on. We will continue adding +mpi in the spec manually. Also we cannot omit it as even if the variant is true in the package for mpi, spack may still turn it off, so can't rely on default

repo/hypre/package.py
scheibelp: I think we agreed this could be removed (I have since done so):  this logic was originally added in https://github.com/LLNL/benchpark/commit/5113774f4b94be4c4a9b5469679b7b34df13a872 by @rfhaque and I think he had some additional context for this.

rfhaque: @scheibelp @pearce8 Its ok to remove this code block

scheibelp:
* Make a list of systems which need their owners to insert paths
  * `cscs-daint`
  * `cscs-eiger`
* Make a list of systems which have not been tested
  * (everything from the first list plus:)
  * `aws-cluster`
  * `cscs-lumi`
  * `jsc-juwels`
  * `lanl-venado`
  * `llnl-matrix`
  * `riken-fugaku`

(for this second list, some of them have dry runs, but none of them are part of our pipelines that run benchmarks; all of them except `llnl-matrix` are outside of LLNL)

repo/gromacs/package.py
```diff
@@ -203,6 +205,10 @@ class Gromacs(CMakePackage, CudaPackage, ROCmPackage):
         msg="GROMACS requires oneMKL 2021.3 or later since version 2023",
     )
 
+    depends_on("c", type="build")
+    depends_on("cxx", type="build")
+    depends_on("fortran", type="build")
```
nhanford:
```suggestion
```

repo/quicksilver/package.py
```diff
@@ -29,6 +29,10 @@ class Quicksilver(MakefilePackage):
     variant("cuda", default=False, description="Build with CUDA support")
     variant("caliper", default=False, description="Build with Caliper support")
 
+    depends_on("c", type="build")
+    depends_on("cxx", type="build")
+    depends_on("fortran", type="build")
```
nhanford:
```suggestion
```

repo/remhos/package.py
```diff
@@ -31,6 +31,10 @@ class Remhos(MakefilePackage, CudaPackage, ROCmPackage):
     variant("metis", default=True, description="Enable/disable METIS support")
     variant("caliper", default=False, description="Enable/disable Caliper support")
 
+    depends_on("c", type="build")
+    depends_on("cxx", type="build")
+    depends_on("fortran", type="build")
```
nhanford, line 36: Perhaps the upstream should have a variant to disable? I see a `--disable-fortran` option mentioned in the repo...

repo/saxpy/package.py
```diff
@@ -28,6 +28,10 @@ class Saxpy(CMakePackage, CudaPackage, ROCmPackage):
     conflicts("+rocm", when="+cuda+openmp")
     conflicts("+openmp", when="+rocm+cuda")
 
+    depends_on("c", type="build")
+    depends_on("cxx", type="build")
+    depends_on("fortran", type="build")
```
nhanford:
```suggestion
```

repo/smb/package.py
```diff
@@ -19,7 +19,12 @@ class Smb(MakefilePackage):
 
     variant("mpi", default=False, description="Build with MPI support")
     variant("rma", default=False, description="Build RMA-MT variant")    
+
+    depends_on("c", type="build")
+    depends_on("cxx", type="build")
+    depends_on("fortran", type="build")
```
nhanford:
```suggestion
```

repo/stream/package.py
```diff
@@ -24,6 +24,10 @@ class Stream(CMakePackage):
 
     requires("@5.10-caliper", when="+caliper")
 
+    depends_on("c", type="build")
+    depends_on("cxx", type="build")
```
nhanford:
```suggestion
```

repo/remhos/package.py
scheibelp: If the `remhos` package does not have a `fortran` variant, then ideally we'd always be explicitly enabling or disabling Fortran support in the build. I don't see one; in general it's safer to assume it is enabled in that case (at least for modeling these language dependencies).

pearce8 (dismissed):

pearce8 (dismissed):

PR #957, opened 2025-08-08
Use different value for placeholder instead of relying on max value

michaelmckinsey1:
## Description

 - [x] Enable setting `n_ranks` > 1000 by removing the check in the allocation modifier. Replace the placeholder value with `-1`, which is more clearly invalid than the previous value. Currently, on develop, setting n_ranks > 1000 will result in `==> Error: Must specify one of: n_nodes, n_ranks, n_gpus` error, since the allocation modifier will set the value to `None`.

- [x] Update `lib/benchpark/system.py`
- [x] Update `modifiers/allocation/modifier.py`


michaelmckinsey1:
I want to check that CI passes

pearce8 (approved):

PR #969, opened 2025-08-12
Use Positive Placeholder for Ramble Check

michaelmckinsey1:
## Description

- [x] https://github.com/LLNL/benchpark/pull/957 used a negative placeholder value, which causes ramble to fail a check https://github.com/GoogleCloudPlatform/ramble/blob/c8376b60a00871429968c559fbeedad5a9e42fdb/lib/ramble/ramble/experiment_set.py#L223. This change sets the placeholder to a positive value, so this check will pass. This check runs before we set the value in the modifier to a valid value, which is why it is an issue in the first place.

- [x] Update `lib/benchpark/system.py` and `modifiers/allocation/modifier.py`


lib/benchpark/test/system.py
```diff
@@ -19,8 +19,8 @@ def test_system_compute_variables_section(monkeypatch):
             "timeout": "120",
             "scheduler": "flux",
             "sys_cores_per_node": 84,
-            "n_ranks": -1,
-            "n_nodes": -1,
+            "n_ranks": 9223372036854775807,
+            "n_nodes": 9223372036854775807,
```
scheibelp: What happens if `sys.maxsize != 9223372036854775807`?

You might want an explicit check for that number. I feel like there's a smaller number that would reasonably bound all requests. Also, I like `max_request` from before, at least the part that does

```
            if val > max_request:
                raise ValueError(f"Request exceeds maximum: {var}/{val}/{max_request}")
```

because if e.g. someone asks for more than `999999` of something, that's probably too much (if it isn't I think they're in a special case where they could bump `max_request`.)

michaelmckinsey1:
@scheibelp waiting on CI

scheibelp (dismissed):

lib/benchpark/cmd/setup.py
```diff
@@ -169,6 +169,7 @@ def find(d, tag):
 . {per_workspace_setup.spack_location}/share/spack/setup-env.sh
 
 export SPACK_DISABLE_LOCAL_CONFIG=1
+export PYTHONPATH="{benchpark.paths.benchpark_root}/lib:$PYTHONPATH"
```
michaelmckinsey1: @scheibelp  This is so the modifier can access `benchpark.accounting`, where I though it more fitting to define the constants. Do you think this overkill; should we keep them hardcoded everywhere instead?

scheibelp: This is for the allocation modifier to access it? I'm worried this could be fragile depending on PYTHONPATH manipulations. Duplication in this instance wouldn't bother me (along with a comment in both spots)

scheibelp (approved):

PR #995, opened 2025-08-26
Set GTL true by default in mpich package and set linker flags

michaelmckinsey1:
## Description

- [x] Set GTL library for packages building cray-mpich, so they do not have to set in package.py. Packages using mpich as external will still need to set linker flags (this is the most common case).

- [x] Update `repo/cray-mpich/package.py`


pearce8 (approved):

pearce8:
@michaelmckinsey1 please let me know when Tioga tests pass as we need to make sure they do before merging gtl-related fixes.

michaelmckinsey1:
tioga tests passing @pearce8

pearce8 (approved):

PR #1034, opened 2025-09-05
Temporarily submit Tuolumne Jobs to pbatch Instead of pci Queue

michaelmckinsey1:
## Description

- [x] Since a single benchpark pipeline submits 3 resource allocations that can take >1hr each, we end up dominating the pci queue on Tuolumne which is only 4 nodes. To be good citizens we should try using pbatch and see if the queue time is noticeable.


michaelmckinsey1:
Ignore dane flux failures

slabasan (approved):

PR #1042, opened 2025-09-09
Adding support for Commbench as a new experiment, also adding El Dorado as a system

simongdg:
## Adding El Dorado system (docs: [Adding a System](https://software.llnl.gov/benchpark/add-a-system-config.html))
- [ ] Add `systems/snl-eldorado/system.py` file - only if it is different from El Capitan software stack
- [ ] Should be able to use El Capitan's `systems/all_hardware_descriptions/hardware_name/hardware_description.yaml` which will appear in the [docs catalogue](https://software.llnl.gov/benchpark/system-list.html)
- [ ] Update system.py to use Spack 1.0 @scheibelp 

## Adding Commbench benchmark (docs: [Adding a Benchmark](https://software.llnl.gov/benchpark/add-a-benchmark.html))

- [ ] package.py
- [ ] update package.py to use Spack 1.0 @scheibelp 
- [ ] application.py
- [ ] Tags in Ramble's `application.py` or in `repo/benchmark_name/application.py` will appear in the [docs catalogue](https://software.llnl.gov/benchpark/benchmark-list.html)
- [ ] Add experiment.py to define a single node and multi-node experiments


repo/cray-mpich/package.py
```diff
@@ -16,7 +16,7 @@ def libs(self):
         libs = super().libs
 
         if self.spec.satisfies("+gtl"):
-            gtl_lib_prefix = self.spec.extra_attributes["gtl_lib_path"]
+            gtl_lib_prefix = self.spec.extra_attributes["gtl_path"] + "/lib"
```
pearce8: @nhanford Is this not standard?  If it is different on our systems and SNL systems, how should we be handling this?

systems/snl-eldorado/system.py
```diff
@@ -0,0 +1,730 @@
...
+                }
+            ]
+        }
+        if self.spec.satisfies("compiler=cce") or self.spec.satisfies(
+            "compiler=rocmcc"
+        ):
+            selections["compilers"] += self.rocm_cce_compiler_cfg()["compilers"]
+
+        # Note: this is always included for some low-level dependencies
+        # that shouldn't build with %cce
```
scheibelp: I think this note is supposed to attach to gcc

systems/snl-eldorado/system.py
```diff
@@ -0,0 +1,730 @@
...
+                        "environment": {},
+                        "extra_rpaths": [],
+                    }
+                }
+            ]
+        }
+        if self.spec.satisfies("compiler=cce") or self.spec.satisfies(
+            "compiler=rocmcc"
+        ):
+            selections["compilers"] += self.rocm_cce_compiler_cfg()["compilers"]
```
scheibelp: The specification for this changed: it looks like https://github.com/LLNL/benchpark/blob/develop/systems/llnl-elcapitan/system.py is a close match (uses all the same compilers). If you have questions about that, let me know.

systems/snl-eldorado/system.py
```diff
@@ -0,0 +1,730 @@
...
+                        }
+                    ]
+                }
+            }
+
+        selections["packages"] |= self.compiler_weighting_cfg()["packages"]
+
+        return selections
+
+    def compiler_weighting_cfg(self):
```
scheibelp, line 302: I think you can entirely replace this with the implementation of `compiler_weighting_cfg` from https://github.com/LLNL/benchpark/blob/develop/systems/llnl-elcapitan/system.py

systems/snl-eldorado/system.py
```diff
@@ -0,0 +1,730 @@
...
+                    "externals": [
+                        {
+                            "spec": "cray-libsci@23.05.1.4%gcc",
+                            "prefix": "/opt/cray/pe/libsci/23.05.1.4/gnu/10.3/x86_64/",
+                        }
+                    ]
+                }
+            }
+
+        selections["packages"] |= self.compiler_weighting_cfg()["packages"]
```
scheibelp: When combining separate config dicts you'll now want to use `merge_dicts` (example in https://github.com/LLNL/benchpark/blob/develop/systems/llnl-elcapitan/system.py)

scheibelp (commented):
This appears similar to https://github.com/LLNL/benchpark/blob/develop/systems/llnl-elcapitan/system.py, so I added some references about what needs to change with the merge of https://github.com/LLNL/benchpark/pull/953

(package recipe looks compatible, no notes there)

repo/cray-mpich/package.py
simongdg: So I have added logic to address this:

https://github.com/LLNL/benchpark/pull/1042/commits/ec22db11c30c9602b4c422a86c964a25c3bcb853 

I have tested it on SNL systems, it should work for LLNL, but needs to be tested.

systems/snl-eldorado/system.py
```diff
@@ -0,0 +1,730 @@
...
+            else:
+                gtl_spec = "~gtl"
+                gtl_cfg = dont_use_gtl
+
+            return {
+                "packages": {
+                    "cray-mpich": {
+                        "externals": [
+                            {
+                                "spec": f"cray-mpich@{self.mpi_version}%cce@{self.cce_version} {gtl_spec} +wrappers",
```
scheibelp: This has to rearrange

`cray-mpich@{self.mpi_version}%cce@{self.cce_version} {gtl_spec} +wrappers`

to

`cray-mpich@{self.mpi_version} {gtl_spec} +wrappers %cce@{self.cce_version}`

systems/snl-eldorado/system.py
```diff
@@ -0,0 +1,730 @@
...
+            40,
+            48,
+            56,
+            64,
+            72,
+            80,
+            88,
+            ],  # 3 cores reserved per socket
+        "sys_gpus_per_node": None,  # Determined by "gpumode" variant
+        "system_site": "llnl",
```
michaelmckinsey1: This would be `llnl` -> `snl`

simongdg:
I have updated my pull request so that the snl-eldorado system looks more similar to llnl-elcapitan system. I also undid the cray-mpich package.py to how it was originally, as I was able to add the gtl_lib_path to my snl-eldorado system.py.

PR #1046, opened 2025-09-10
move PR template, remove dry run task, dry runs are automatically added

slabasan:


michaelmckinsey1 (approved):
I removed reference to single_node while we are at it, which no longer exists

pearce8 (approved):

PR #1051, opened 2025-09-12
Add variance Option to Caliper Modifier

michaelmckinsey1:
## Description

- [x] enable `time.variance` in caliper modifier, e.g. `caliper=mpi,time,variance`


michaelmckinsey1:
variance test https://lc.llnl.gov/gitlab/benchpark/benchpark/-/jobs/2921980

michaelmckinsey1:
kripke/tuolumne failure is unrelated.

pearce8 (changes requested):
@michaelmckinsey1 The default Caliper mode (time) should record inclusive time, exclusive time, and variance.  There is no reason to have these as separate modes, it doesn't cost much to calculate these on the basis of the data that is being collected already.

michaelmckinsey1:
@pearce8 Ok that makes this change a one-liner then acutally. Inclusive time will already be there.

pearce8 (approved):

PR #1062, opened 2025-09-17
Rewrite System Documentation

### LINKED ISSUE #1049: Update docs for Spack 1.0
- [ ] docs/add-a-system-config.rst
- [ ] verify links to spack docs are still correct

michaelmckinsey1:
## Description

- [x] Rewrite system documentation, specific to spack1.0 changes
- [x] Fix `benchpark system external` for spack1.0
- [x] Improve instructions on package finding process -> moved docs from own page to appendix in new system docs.
- [x]  Should address #1049 

- [x] Update `systems/aws-tutorial`

- [x] move content from `update-a-system-config.rst` to appendix in `add-a-system-config.rst`


docs/add-a-system-config.rst
```diff
@@ -43,15 +44,23 @@ in :doc:`system-list`.
 If you are running on a system with an accelerator, find an existing system with the
 same accelerator vendor, and then secondarily, if you can, match the actual accelerator.
 
-1. accelerator.vendor
-2. accelerator.name
+1. ``accelerator.vendor`` - Company name
```
michaelmckinsey1, line 47: Do we think the hardware description deserves multiple sections? Or is it pretty self-explanatory from the existing examples (I think the latter).

docs/add-a-system-config.rst
```diff
@@ -74,373 +84,460 @@ where:
...
+In this example, we will recreate a fully-functional example of the AWS ``system.py``
+that we use for benchpark tutorials (see `aws-tutorial/system.py
+<https://github.com/LLNL/benchpark/blob/develop/systems/aws-tutorial/system.py>`_). To
+start, we import the base benchpark ``System`` class, which our ``AwsTutorial`` system
+will inherit from. We also import the maintainer and variant directives, which provide
+the utilities to track a maintainer by their GitHub username and variants to specify
+configurable properties of our system. We can specify the different AWS instances that
+share this same hardware and software specificaiton using the ``instance_type`` variant.
+We use ``instance_type`` here instead of ``cluster`` (you will see ``cluster`` in other
+systems), to differentiate a cloud ``instance`` from an HPC ``cluster``.
```
michaelmckinsey1: How do we want to explain why we use ``instance_type`` over ``cluster`` for cloud?

docs/add-a-system-config.rst
```diff
@@ -74,373 +84,460 @@ where:
...
-spack (``software.yaml``). The actual version will be rendered by Ramble when it is
-built.
+Here, we define the ``compute_packages_section()`` function, where you can include any
+package that you would like a package manager, such as spack, to find as an "external",
+meaning it will not build that package from source and use your system package instead.
+For each package that you include, you need to define its spec ``name@version`` and the
+system path ``prefix`` to the package. Additionally for spack, you need to set
+``buildable: False`` to use the package as an external.
+
+At minimum, we recommend you define externals for ``cmake``, ``mpi``, ``blas``, and
```
michaelmckinsey1: Any others?

docs/add-a-system-config.rst
```diff
@@ -74,373 +84,460 @@ where:
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+In the ``compute_compilers_section``, we define the compilers available on the system.
+For our AWS system, this is ``gcc@11.4.0``. We return a dictionary, with the helper
+``compiler_section_for()`` function, that formulates the compiler ``name`` and
+``entries`` for Spack, where the ``entries`` are a list of ``compiler_def()``. For the
+``compiler_def()``, we must at minimum specify the ``spec``, ``prefix``, and ``exes``:
+
+1. ``spec`` - Similar to package specs, ``name@version``. GCC in particular also needs
+   the ``languages=c,c++,fortran`` variant.
```
michaelmckinsey1: @scheibelp Maybe there is a better way to explain the languages?

systems/aws-tutorial/system.py
```diff
@@ -15,41 +15,39 @@ class AwsTutorial(System):
 
     maintainers("stephanielam3211")
 
+    common = {
+        "system_site": "aws",
+        "programming_models": [OpenMPCPUOnlySystem()],
```
scheibelp: It is different to instantiate a single programming model instance globally (vs. in `__init__`), but I don't think it will have a negative effect to do so.

docs/update-a-system-config.rst
```diff
@@ -1,70 +0,0 @@
...
-    Copyright 2023 Lawrence Livermore National Security, LLC and other
-    Benchpark Project Developers. See the top-level COPYRIGHT file for details.
-
-    SPDX-License-Identifier: Apache-2.0
-
-Updating a System
-=================
-
-If a system already exists, its external package definitions can be updated from the
-output of `benchpark system external`:
```
scheibelp, line 11: (question) it looks like everything about updating systems (and `benchpark system external`) that is relevant moved into `add-a-system-config.rst` - just wanted to confirm.

docs/add-a-system-config.rst
```diff
@@ -74,373 +84,461 @@ where:
...
 
-2. Adding or Parameterizing System Software Stack
+B. Creating the System Definition (``system.py``)
 -------------------------------------------------
 
-``system.py`` in Benchpark provides an API to represent a system software stack as a
-command line parameterizable object. If none of the available software stack
-specifications match your system, you may add a `new-system` directory in the `systems`
-directory where the `new-system` directory name follows the naming convention:
+Now that you have defined the hardware description for your system, you can now create
```
scheibelp, line 104: (question) Is it strictly-required to create the hardware description first?

docs/add-a-system-config.rst
scheibelp: `cluster` is more of an idiom than a requirement

In general, both `instance_type` and `cluster` are intended to distinguish between mostly-similar systems that only differ in a few details (but may use the same MPI, externals, architecture, etc.) so something like

I think this is also the first example of a variant, so you could say something like:

> There are many similar types of AWS nodes that differ only in  terms of the number of processors, memory (but otherwise have the same system packages available): this can be encoded with a variant in a benchpark system - the user can indicate what type of instance they are creating, and the system description will reflect the instance type chosen. 
>
> (Note: most system classes in Benchpark have a similar concept, but generally the idea is that there are a few different multi-node machines with very-similar configs, and so they typically use the term "cluster")

docs/add-a-system-config.rst
scheibelp: In terms of a tutorial, I think succinct is good vs. explaining the why of it. Note that if for some reason there is no gfortran, then `languages` should not include `,fortran` (that might not be worth mentioning here though)

I would recommend some word smoothing for the first paragraph though:

> Define compilers that are available on your system by implementing ``compute_compilers_section()``:
> - for each compiler, create the necessary config with `compiler_def`
> - for each type of compiler (gcc, intel, etc.), combine them with `compiler_section_for`
> - then merge them with `merge_dicts` (this part is unnecessary if you have only one type of compiler)
> - generally you will want to compose a minimal list of compilers: e.g. if you want to compile your benchmark with the oneAPI compiler, and have multiple versions to choose from, you'd add a variant to the system, and the config would expose only one of them

scheibelp (commented):
This looks generally good - thanks! I have a couple questions and a couple suggestions (but nothing that should block this).

docs/add-a-system-config.rst
pearce8: I think this is self-explanatory.

docs/add-a-system-config.rst
pearce8: Technically not required, but we want people to realize if there are similar systems specified already, and either not duplicate effort or use them as a starting point.

pearce8: Maybe we need to add a step to check for other systems that have the same description.  I wonder if this could be a yaml comparison script - of course we would need to define what "similar" is, and this may be plagued by different spelling and naming conventions humans will use here.  Thoughts?

pearce8: @scheibelp @michaelmckinsey1

docs/add-a-system-config.rst
pearce8: I like @scheibelp 's suggestion.  One modification: 

(Note: most system classes in Benchpark have a similar concept, but often they refer to physical (named) clusters with very-similar configs, and so they typically use the term "cluster" rather than "instance_type".)

docs/add-a-system-config.rst
pearce8: @michaelmckinsey1 We want to keep this list short, but I think we need to add a sentence on needing to define cuda or rocm runtimes for GPUs; look at specific systems to make sure you use the right handles and point at them as examples.

docs/add-a-system-config.rst
pearce8: I like @scheibelp 's suggestion.

docs/add-a-system-config.rst
```diff
@@ -74,373 +84,461 @@ where:
...
+the ``system.py``, which involves defining the software on your system. This includes
+defining compilers and pre-installed packages, which your package manager can use
```
pearce8: @michaelmckinsey We need to talk about the mandatory vs optional definitions here separately.  Tackle mandatory first.  Then the faster build process discussion needs to be in the compute packages section.  Please make sure we are not explaining this in multiple places.

@scheibelp Here I would like to tell people how they can generate the list of externals.  Is there a good place in Spack docs we should link?

michaelmckinsey1:
going to fail lint until https://github.com/LLNL/benchpark/pull/1077 merged

docs/add-a-system-config.rst
michaelmckinsey1: Changed in [feecc7c](https://github.com/LLNL/benchpark/pull/1062/commits/feecc7c50134069dd740ec670ff18aaf1dba4043)

docs/add-a-system-config.rst
michaelmckinsey1: 7cfa5e0

docs/add-a-system-config.rst
michaelmckinsey1: [f5f02a8](https://github.com/LLNL/benchpark/pull/1062/commits/f5f02a814b7b446ed855ff4f1c913be8986fdc2d)
I added Peters suggestion, but I think I need to try adding a clang compiler. This would be good to show an example of using the merge_dicts, and would not require the languages variant (we discussed this yesterday).

So I'm leaving this unresolved.

systems/aws-tutorial/system.py
michaelmckinsey1: Ah. This could be an issue I think in regards to a use case like https://github.com/LLNL/benchpark/pull/931. I will revert the programming_models back to the init function. Hopefully, it is clear what I am going for here. We should recommend non-objects are refactored for cleaner system definitions.

michaelmckinsey1: Reverted this line in [dee4ab0](https://github.com/LLNL/benchpark/pull/1062/commits/dee4ab0f239eddf0ff241fbd40beb5a356fcd265)

docs/update-a-system-config.rst
michaelmckinsey1: Yes. I moved this short page to the appendix of the system page, since that is where it is relevant.

docs/add-a-system-config.rst
michaelmckinsey1: We have our own command to generate and update the list of externals, which is in the appendix and referenced in the "Add a packages section". `benchpark system external` can compare the package defined in the system already against newly found externals. And if there are no defined packages, it works the same as `spack external find`

michaelmckinsey1: [61be5aa](https://github.com/LLNL/benchpark/pull/1062/commits/61be5aafd22d1058c140695c5baa8a20dce4b82f) separates mandatory vs optional steps. The package section now appears last, as it is the only optional step. We still recommend within that section that certain packages are defined for build/performance reasons depending on the package.

docs/add-a-system-config.rst
michaelmckinsey1: Upon discussing this, we decided to leave for future work #1079. While it may be feasible to make a script for this, I think the naming of the folders `[INTEGRATOR]-MICROARCHITECTURE[-ACCELERATOR][-NETWORK]` is quite descriptive to determine similarity. If a user is able to properly identify those four things, they can match their system to an existing description. And if they cannot correctly identify those, then the script would also fail.

docs/add-a-system-config.rst
michaelmckinsey1: This is now #1079

scheibelp (dismissed):

slabasan (approved):

pearce8 (approved):

Issue #1068, opened 2025-09-19
AMG2023 ROCm: Umpire build error on Tuolumne


pearce8:
- [ ] Reproducer for @scheibelp 
- [ ] What is the issue?
- [ ] Workaround is to remove .spack from home directory - how does it fix the issue?

pearce8:
Intermittent spack error, should be fixed with spack cache removals

PR #1071, opened 2025-09-19
Locate spack stages/cache in benchpark workspace

scheibelp:
* Setting SPACK_USER_CACHE_PATH will ensure isolation between different benchpark instances and benchpark workspaces (I thought setting `config:misc_cache` would achieve this, but people have been reporting problems like https://github.com/LLNL/benchpark/issues/1068)
* Moving the stage into the benchpark workspace means that spack build stages are no longer stored in a temporary dir (so can be examined after an allocation ends)

### DIFF
--- a/lib/benchpark/cmd/setup.py
+++ b/lib/benchpark/cmd/setup.py
@@ -171,6 +171,7 @@ def command(args):
 
     pkg_str = ""
     if pkg_manager == "spack":
+        spack_build_stage = experiments_root / "builds"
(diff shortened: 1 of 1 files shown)


pearce8 (changes requested):
@scheibelp Please add instructions on how to use this to the docs, probably in getting-started (depending when the user has to set this).

scheibelp:
This is always-on, not configurable.

pearce8 (approved):

PR #1074, opened 2025-09-22
Define MPI as Virtual in llnl-matrix

michaelmckinsey1:
## Description

- [x] I have not manually tested this yet, but this matches the syntax of other systems. @scheibelp Noticed this was defined incorrectly in ``llnl-matrix``
- [x] Manual tests pass for kripke and AMG

- [x] Update `llnl-matrix`


systems/llnl-matrix/system.py
```diff
@@ -163,15 +163,15 @@ def compute_packages_section(self):
                     "externals": [{"spec": "curl@7.61.1", "prefix": "/usr"}],
                     "buildable": False,
                 },
+                "mpi": {"buildable": False},
```
michaelmckinsey1, line 166: @scheibelp Can I define this here? If somehow MPI is not enabled, will defining the MPI virtual break things?

scheibelp: > Can I define this here? 

Sorry, by here I think you mean universally (and not along with the actual mpi-implementation definitions)? Yes that is fine

> If somehow MPI is not enabled, will defining the MPI virtual break things?

* This is fine to include if experiments don't need mpi
* Also fine - sort of - if no mpi implementation is defined
* But if this is present, and no external mpi implementations are defined, and an experiment (or rather its package.py) declares a dependency on mpi, then there would be a failure (although I would consider that a good thing)

If I misunderstood the question, let me know

michaelmckinsey1: Yes I meant outside of the MPI definitions. Putting it in the MPI definitions would avoid any confusion, but would require defining it for each MPI, so if that isn't necessary I think this is better.

scheibelp (approved):

PR #1080, opened 2025-09-24
Fix diffExperimentBuilds.py Script

michaelmckinsey1:
## Description

- [x] I noticed this script was no longer working. Underlying changes to the library broke the script. The cli is the same.

- [x] Update `lib/scripts/diffExperimentBuilds.py`


slabasan:
Is there a way to add this to CI so we can catch breaking changes from the core code?

pearce8 (approved):

PR #1088, opened 2025-09-25
Enable Persistent Alternative "benchpark home" Location

### LINKED ISSUE #1061: Enable Different bootstrap Location than Home
The benchpark bootstrap takes up too much space to expect all users to be able store this in the home directory. We should write a configuration file to the benchpark root directory that tracks the bootstrap location, enabling an alternate location, which we can check when needing to find the bootstrap.

Then, enable interface in `benchpark boostrap --location /custom/usr/path/` to specify alternate path.

michaelmckinsey1:
## Description

- [x] Add a new command `benchpark configure` that creates configuration file in the benchpark root, `configure.yaml`, which stores data that must persist after the process dies, such as the bootstrap location. Benchpark now checks for this file and uses the location, if it exists. If it does not exist, the default directory is still the home directory for the benchpark bootstrap.
- [x] Add FAQ docs for moving bootstrap location.
- [x] Resolves https://github.com/LLNL/benchpark/issues/1061

Usage:
```
# Create a `configure.yaml` configuration file in benchpark root dir, with location
benchpark configure --bootstrap-location /Users/mckinsey1/Documents/repos/benchpark
# Create the bootstrap at that location
benchpark bootstrap
```
Then use benchpark as normal

- [x] Update `FAQ.rst`
- [x] Update `lib/`
- [x] Added test


michaelmckinsey1:
from dryrunexperiments:
```
Cloning Ramble to /home/runner/.benchpark/ramble
Writing to /home/runner/work/benchpark/benchpark/benchpark.yaml
Cloning packages to /home/runner/.benchpark/spack-packages
Cloning Spack to /home/runner/.benchpark/spack
Cloning Ramble to /home/runner/.benchpark/.benchpark/ramble
Writing to /home/runner/work/benchpark/benchpark/benchpark.yaml
Cloning packages to /home/runner/.benchpark/.benchpark/spack-packages
Cloning Spack to /home/runner/.benchpark/.benchpark/spack
```
there is some bug where it is not finding the first path, it should not clone twice

pearce8 (changes requested):
@michaelmckinsey1 Is it possible to add a test for this feature?

michaelmckinsey1:
> @michaelmckinsey1 Is it possible to add a test for this feature?

@pearce8 There now exists a test for both the path location and that recursive paths are not created (https://github.com/LLNL/benchpark/pull/1088#issuecomment-3335873802).

michaelmckinsey1:
Waiting to verify CI

michaelmckinsey1:
Would be good to get @scheibelp opinion on the implementation

nhanford:
This looks really good. Consider doing it more like Spack with a `configure` command and a `configure.yaml`. So it could be something like `benchpark configure bootstrap --location /path/to/boostrap/location`. That could keep it more general so we could add more stuff to the configuration of the Benchpark deployment later.

michaelmckinsey1:
> This looks really good. Consider doing it more like Spack with a `configure` command and a `configure.yaml`. So it could be something like `benchpark configure bootstrap --location /path/to/boostrap/location`. That could keep it more general so we could add more stuff to the configuration of the Benchpark deployment later.

This sounds better

lib/benchpark/cmd/configure.py
```diff
@@ -0,0 +1,30 @@
...
+    )
+
+
+def command(args):
+    data = {}
+
+    if args.bootstrap_location:
+        data["bootstrap"] = {
+            "location": str(args.bootstrap_location).rstrip("/") + "/.benchpark/",
+        }
```
michaelmckinsey1, line 32: We can extend the `setup_parser` and this logic here to configure other aspects of benchpark

michaelmckinsey1:
> This looks really good. Consider doing it more like Spack with a `configure` command and a `configure.yaml`. So it could be something like `benchpark configure bootstrap --location /path/to/boostrap/location`. That could keep it more general so we could add more stuff to the configuration of the Benchpark deployment later.

@nhanford 
This is now implemented

lib/benchpark/cmd/configure.py
nhanford: Excellent

nhanford (approved):
This looks great Mike! Thanks!

pearce8 (approved):

PR #1090, opened 2025-09-25
Add CI Check to Validate Docs Scripts Before Build

### LINKED ISSUE #999: Add action to check if the experiments table got generated

michaelmckinsey1:
## Description

- [x] Resolves #999 
- [x] The tables at https://software.llnl.gov/benchpark/benchmark-list.html and https://software.llnl.gov/benchpark/system-list.html depend on python scripts, that sometimes fail when the underlying benchpark library changes. We do not catch this, because the CI that builds the docs only runs when merged into develop. So this is adding a CI check that the scripts are working, that will run on PRs.

- [x] Update `.github/workflows`

### DIFF
--- a/docs/generate-benchmark-list.py
+++ b/docs/generate-benchmark-list.py
@@ -9,6 +9,7 @@ sys.path.append("../lib/")
 import benchpark.accounting  # noqa: E402
 import benchpark.paths  # noqa: E402
 
+raise ValueError("test")
 
 def construct_tag_groups(tag_groups, tag_dicts, dictionary):
     # everything is a dict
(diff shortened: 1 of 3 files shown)


docs/generate-benchmark-list.py
```diff
@@ -9,6 +9,7 @@
 import benchpark.accounting  # noqa: E402
 import benchpark.paths  # noqa: E402
 
+raise ValueError("test")
```
michaelmckinsey1: this should fail the test

michaelmckinsey1: This failed correctly https://github.com/LLNL/benchpark/actions/runs/18103625793/job/51512677675?pr=1090

pearce8 (approved):

PR #1092, opened 2025-09-25
add local version of cray's libsci package

slabasan:
## Description

Local version of cray's libsci package

On perlmutter, we were hitting this error:
```
RuntimeError: cannot determine libsci libraries
```

In inspecting further, we see that `candidates = [name for name in self.canonical_names.values() if name.lower() in self.prefix.lower()]` resolves to `candidates = ["GNU", "CRAY"]`, but is looking for a list of length 1. So we take the first element (`GNU`) as it is the identifier to picking the right libsci library.


slabasan:
Thinking this is no longer needed..will re-open if this changes. The existing logic in the cray-libsci package seems to be correct.

PR #1100, opened 2025-10-01
llnl-elcapitan gcc missing +gtl

michaelmckinsey1:
## Description

We don't have any cases at the moment using gcc and GTL, but these changes enable it. The package would need to specify the compilers for gcc vs llvm-amdgpu, otherwise gcc will be used and you will get a HIP error.
- [x] Spec `llnl-elcapitan` gcc compiler so it can be `+gtl`
- [x] Add `llvm-amdgpu` to gcc config, such that `gcc+rocm` does not build `llvm-amdgpu`
   - [x] Specify `gcc` preference so `amdclang` is not used in this case 

- [x] Update `llnl-elcapitan/system.py`

### DIFF
--- a/systems/llnl-elcapitan/system.py
+++ b/systems/llnl-elcapitan/system.py
@@ -467,7 +467,7 @@ class LlnlElcapitan(System):
                 gtl_spec = "+gtl"
             else:
                 gtl_spec = "~gtl"
-    
+
             return {
                 "packages": {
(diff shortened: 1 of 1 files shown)


michaelmckinsey1:
Only merge if all CI passing

rfhaque (approved):

pearce8 (approved):

PR #1106, opened 2025-10-07
Add lammps pace workload

rfhaque:
## Description
This PR add the PACE workload to lammps

### DIFF
--- a/.gitlab/tests/nightly.yml
+++ b/.gitlab/tests/nightly.yml
@@ -31,12 +31,6 @@ include:
           - salmon-tddft
           - smb
           - stream
-      # LAMMPS ACE test
-      - HOST: dane
-        ARCHCONFIG: llnl-cluster
-        SCHEDULER_PARAMETERS: $DANE_PARAMS
-        BENCHMARK: [lammps]
-        VARIANT: [workload=pace]
       # +rocm tests
       - HOST: [tuolumne]
         ARCHCONFIG: llnl-elcapitan
@@ -100,11 +94,6 @@ include:
         SCHEDULER_PARAMETERS: $DANE_PARAMS
         BENCHMARK: [amg2023, kripke, lammps, raja-perf, remhos, stream]
         BENCHMARK_VERSION: develop
-      - HOST: dane
-        ARCHCONFIG: llnl-cluster
-        SCHEDULER_PARAMETERS: $DANE_PARAMS
-        BENCHMARK: [lammps]
-        VARIANT: [+rocm]
 nightly_test_run:
   extends: .nightly_rules
   resource_group: $HOST
--- a/.gitlab/tests/nightly.yml
+++ b/.gitlab/tests/nightly.yml
@@ -31,6 +31,12 @@ include:
           - salmon-tddft
           - smb
           - stream
+      # LAMMPS ACE test
+      - HOST: dane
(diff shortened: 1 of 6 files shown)


repo/lammps/package.py
```diff
@@ -9,15 +9,12 @@
...
 class Lammps(BuiltinLammps):
 
-  version("stable_29Aug2024_update3", tag="stable_29Aug2024_update3")
+  variant("pace", default=False, description="Enable the ML PACE module")
+  depends_on("pace", when="+pace")
 
   depends_on("kokkos+openmp cxxstd=17", when="+openmp")
-  depends_on("kokkos+rocm", when="+rocm")
-  depends_on("kokkos+cuda cxxstd=17", when="+cuda")
+  depends_on("kokkos+wrapper", when="+cuda")
```
rfhaque, line 17: @scheibelp @pearce8 How to rewrite this spack clause so that it satisfies the following condition in the `kokkos/package.py`. This is on matrix using the gcc compiler
```
if spec.satisfies("~wrapper+cuda") and not (spec.satisfies("%clang") or spec.satisfies("%cce") 
or spec.satisfies("+cmake_lang")):  
        raise InstallError("Kokkos requires +wrapper when using +cuda without %clang, %cce or +cmake_lang")   
```

scheibelp: I think this indicates missing constraints in `kokkos`, for now you can add something to our extension of kokkos like:

```
class Kokkos(BuiltinKokkos):
    requires("+wrapper", when="+cuda ~cmake_lang %gcc")
    # ... and so on for each compiler of interest other than cce and clang
```

(at which point the spec will automatically concretize to `+wrapper` on Matrix. If we are trying to avoid `+wrapper`, we can add `require: +cmake_lang` to kokkos on Matrix)

pearce8: @rfhaque Did you address this suggestion from @scheibelp ?

pearce8 (changes requested):
@rfhaque please let me know if you have addressed Peter's suggestion

pearce8:
@michaelmckinsey1 Please kick off CI.

repo/lammps/package.py
rfhaque: @pearce8 Need to give this a bit more thought. It's not clear how to enforce `+cmake_lang` in kokkos on Matrix as spack does not allow system-specific variants. It's possible we can get by with alwayss using `depends_on("+cmake_lang", when="+cuda")`, but that is not what the upstreamed Kokkos package.py currently does. sparta also faces the same issue

.gitlab/tests/nightly.yml
```diff
@@ -94,6 +100,11 @@ include:
         SCHEDULER_PARAMETERS: $DANE_PARAMS
         BENCHMARK: [amg2023, kripke, lammps, raja-perf, remhos, stream]
         BENCHMARK_VERSION: develop
+      - HOST: dane
+        ARCHCONFIG: llnl-cluster
+        SCHEDULER_PARAMETERS: $DANE_PARAMS
+        BENCHMARK: [lammps]
+        VARIANT: [+rocm]
```
michaelmckinsey1: Should this also be ran with `workload=pace`? In that case this line would be `VARIANT: ["+rocm workload=pace"]`

pearce8 (approved):

PR #1115, opened 2025-10-21
Fix gpcnet benchmark

scheibelp:
gpcnet 1.2 doesnt appear to build with gcc 10 or later; it is old, so replace it with gpcnet `master` version. Also it appears that the builtin definition of the gpcnet package is usable, so replace Benchpark's definition with that.


pearce8:
@scheibelp I think you need to resolve conflicts here anyways, which should re-run the CI.

pearce8 (approved):

PR #1116, opened 2025-10-21
Add Instructions for Fixing Linter Errors and Fix Linter Checks

pearce8:
- [x] Update `ci-developer-guide.rst` on how to fix each type of linter error
- [x] Fix `isort` and `flake8` checks, which were previously not being ran
- [x] Changes to `*.py` Files are from running isort


michaelmckinsey1 (dismissed):
I moved the docs to a more relevant section. I also noticed we were not even running `isort` correctly, so I fixed that.

PR #1125, opened 2025-10-28
Riken-fugaku update after spack1.0

SBA0486:
## Description
Updated Riken-fugaku system definition after spack1.0.
Experiment.py adds fujitsu-mpi and fujitsu-ssl2 to experiments' environment packages for fugaku if needed.

- [x] Modify `systems/system_name/system.py` file
- [x] Modify `lib/benchpark/experiment.py` file

- [x]  Modify 'repo/saxpy/package.py' to work with riken-fugaku system definition.


lib/benchpark/experiment.py
```diff
@@ -516,6 +516,14 @@ def compute_package_section_wrapper(self):
                 f"Package section must be defined for application package {self.name}"
             )
 
+        if hasattr(self, "system_spec"):
```
pearce8: @scheibelp Can you please propose how this could be moved to systems/riken-fugaku/system.py?

repo/saxpy/package.py
```diff
@@ -29,6 +29,8 @@ class Saxpy(CMakePackage, CudaPackage, ROCmPackage):
     conflicts("+openmp", when="+rocm+cuda")
 
     depends_on("c", type="build")
```
pearce8: @scheibelp Any ideas why SBA is running into needing these here (while we don't)?  Is something in the Fugaku system definition triggering this?  I'd like to understand as it may impact other benchmarks.

scheibelp: It looks like the saxpy source is c++ rather than c, so I think it is correct to have 

`depends_on("cxx", type="build")`

the c/fortran dependencies should not be needed.

lib/benchpark/experiment.py
scheibelp: These calls would be augmenting the config generated in the `riken-fugaku` system:

```
            "software": {
                "packages": {
                    "default-mpi": {"pkg_spec": "fujitsu-mpi"},
                    "lapack": {"pkg_spec": "fujitsu-ssl2"},
                }
            }
```

Effectively replacing them with

```
            "software": {
                "packages": {
                    "default-mpi": {},
                    "lapack": {},
                }
            }
```

@SBA0486 

* I assume you would want to use `fujutsu-mpi` and I think this change may avoid using it. Do you want to use mpi implementations other than `fujitsu-mpi` for some experiments?
* do you use a different lapack implementation with compilers other than gcc?

Maybe a better question is: what goes wrong without the changes in this function?

SBA0486: @scheibelp I want to rather use `fujitsu-mpi/fujitsu-ssl2` which are available on fugaku. These calls would add `default-mpi` and `lapack` packages to the experiment config generated in  `experiment/ramble.yaml` when `--system=riken-fugaku` like follows. 
```  software:
    environments:
      amg2023:
        packages:
        - amg2023
        - hypre
        - default-mpi
        - lapack 
```
If not added in here, it tries to install openmpi/openblas which tends to fail.

As for why I excluded gcc for lapack, the fujitsu-ssl2 package requires fj or clang compiler.

scheibelp: > If not added in here, it tries to install openmpi/openblas which tends to fail.

Ah got it, this can be controlled by adding

`"mpi": {"buildable": False},`

to `riken-fugaku`'s `compute_packages_section`: at that point anything that needs `mpi` will be redirected to use the external `fujitsu-mpi` entry. `llnl-elcapitan` has an example at https://github.com/LLNL/benchpark/blob/develop/systems/llnl-elcapitan/system.py#L328

You can do the same thing with `blas`/`lapack` (`llnl-elcapitan` has something a bit different because there's an option to choose between a couple providers, but it looks like you set things up so there's always one `fujitsu-ssl2` based on which compiler you pick, so it can be treated the same as `mpi`).

SBA0486: Adding `"mpi": {"buildable": False},` and `blas/lapack: {require: fujitsu-ssl2}` works with the side effect that packages that defaulted to +mpi now install without mpi. When mpi is required though it default to fujitsu-mpi.

pearce8 (dismissed):

pearce8 (approved):

PR #1127, opened 2025-10-28
Change `llnl-matrix` to use gcc as default compiler over oneapi

michaelmckinsey1:
## Description

- [x] nvcc does not work well with oneapi

- [x] Update `llnl-matrix`


michaelmckinsey1:
this should pass lammps and lammps pace workloads tests in gitlabs

michaelmckinsey1:
https://lc.llnl.gov/gitlab/benchpark/benchpark/-/jobs/3058519 should pass (if it gets an allocation)

pearce8:
Merging in develop to try to get more runs on Matrix to re-launch.

pearce8:
@michaelmckinsey1 I don't see any runs on Matrix?

michaelmckinsey1:
@pearce8 If they never get an allocation, they will never post their status here.

pearce8 (approved):

PR #1130, opened 2025-11-07
Experiment management: Breaks backward compatibility [Merge after 1151]

scheibelp:
- [x] Update docs to reflect the change @michaelmckinsey1 
- [ ] What needs to be updated in the tutorials to reflect the interface change?  @slabasan 

## 1: experiments are now placed in the associated system directory

go from

```
benchpark experiment init --dest=osu-micro-benchmarks --system=matrix osu-micro-benchmarks +cuda
benchpark setup osu-micro-benchmarks wkp/
```

to

```
benchpark experiment init matrix osu-micro-benchmarks +cuda
benchpark setup matrix/osu-micro-benchmarks wkp/

# and now the ramble workspace is
# wkp/matrix/osu-micro-benchmarks/workspace/
# vs.
# wkp/osu-micro-benchmarks/matrix/workspace/
```

i.e.

- `experiment init` does not need `--dest` if it's just the experiment name
- experiments are always stored in their relevant system dir
  - (a system dir will store all the experiments that are associated with it)


nhanford:
This looks great! I have a few (possibly naive) questions:

It's been a long time since I've done intense things with the argument parser. Is it smart enough to still work if a user reverses the positions of the system and experiment names? I don't expect these to ever collide (although I may eat my words later when someone names a system saxpy). Otherwise, I suppose it would still be a good idea to keep the `--system` flag to avoid confusion from users trying to memorize the command. Maybe also support `--sys` or `-s`, unless you have plans for `-s` to mean something else later on.

What if I wish to modify an experiment, not by modifying the `experiment.py`, but by simply changing arguments (say `workloads`) on this `init` command? Will I get the expected re-instantiation behavior? Perhaps the best way to do this would be to prompt the user with a `[y/N]` since they would be effectively overwriting their existing config.

Will this collide on an experiment being *added* to an existing system? This is a case we should be sure to test.

scheibelp:
> It's been a long time since I've done intense things with the argument parser. Is it smart enough to still work if a user reverses the positions of the system and experiment names?

No: I could update the parser to allow either order, but I'd have to inspect the arguments, e.g. figure out which one is a directory (which isn't that hard). FWIW

- `benchpark redo --help` could tell the user the order
- Before this, `--system` was a mandatory option (a bit contradictory, but users were forced to do `benchpark experiment init --system=...`
- In my experience relying on `--help` would be sufficient for new users, I don't know of commands that infer order, and I think for power users it would be convenient to be able to to omit `--system`

> What if I wish to modify an experiment, not by modifying the experiment.py, but by simply changing arguments (say workloads) on this init command?

If you didn't change any `package.py` files in the repo any `application.py`, `system.py`, or `experiment.py`, and only are submitting a new experiment spec, then FWIW cleaning up should be unnecessary: you can add the new experiment spec (with a distinct `--dest`) to the system.

If you wanted a "clean" slate for experiments, then as it is now, this command isn't good at suggesting the minimum changes necessary for that (although I could update it to do that fairly easily).

It's worth noting that at the moment this command outputs all the cleanup necessary to handle changes to all those files, but does not minimize cleanup for subsets of changes. Assuming the user does all their changes within the benchpark git repository, this could use Git hashing of files to potentially auto-detect the minimum cleanup (IMO that might be worth splitting off into a future effort though).

> Will this collide on an experiment being added to an existing system? 

With these new changes, you must create the new system directory first. You can add multiple experiments to it, although if you add the same experiment multiple times (specify the same `--dest` twice, or instantiate two experiments with the same name without specifying `--dest`) then that will cause a collision.

michaelmckinsey1:
The dryruns are failing (https://github.com/LLNL/benchpark/actions/runs/19155191267/job/54754462496?pr=1130) because it is expecting a directory that doesnt exist. It's expecting the form `benchmark/system` where it is now `system/benchmark`

lib/benchpark/cmd/experiment.py
```diff
@@ -18,29 +18,27 @@ def experiment_init(args):
...
     else:
-        raise ValueError("Must specify one of: --dest, --basedir")
+        experiment_component = experiment_spec.name
+
+    destdir = os.path.join(args.system, experiment_component)
 
     try:
         os.makedirs(destdir)
         experiment.write_ramble_dict(f"{destdir}/ramble.yaml")
+        print(f"Run `benchpark setup {destdir} ...` to generate Ramble workspace")
```
michaelmckinsey1: Since the only option is the `experiments_root`, I think this is clearer than `...` (I think ellipsis is fine for system since there can be varying amount of variants). Or maybe `benchpark setup {destdir} EXPERIMENTS_ROOT`?
```suggestion
        print(f"Run `benchpark setup {destdir} <experiments_root>` to generate Ramble workspace")
```

michaelmckinsey1 (changes requested):

michaelmckinsey1 (dismissed):

pearce8 (approved):

PR #1134, opened 2025-11-10
cray-mpich: set ldflags

slabasan:
## Description

Sets ldflags to properly set libmpi_gtl_cuda.so on perlmutter


michaelmckinsey1:
Before this change, we were not passing the `ld_flags` that we were defining in the system.py to cray-mpich package, so those flags were never being used, resulting in the "gtl requested but not linked" error

michaelmckinsey1:
waiting to see if kripke on elcap passes with this change

pearce8 (approved):

PR #1136, opened 2025-11-12
Cray mpich wrappers

scheibelp:
How do deal with GPU aware MPI (maybe this needs to be recorded somewhere):

- [x] MPI is where GPU awareness needs to live, so the package links to the correct stuff (using gtl, etc.) and sets the appropriate environment variable for the MPI library if needed (these are not standardized).
- [x] If building application+cuda, will build mpi+cuda.
- [x] To turn off gpu-aware MPI, the application+cuda package needs to build mpi~cuda.
- [x] application+cuda or application~cuda will also set the internal variables for the app source (i.e., MFEM_GPU_AWARE) which will be used in ifdefs. 

In this PR:

- [x] Adds a `cray-mpich-gtl` package that copies everything from `cray-mpich` except for `mpicc/mpicxx/mpif90/mpif77` - for those wrappers, it creates a secondary wrapper script that inserts `-lmpi_gtl_hsa`.
- [x] This should make it so any package that would use `cray-mpich+gtl` does not have to link this library itself (I've only tested the substitution effects where packages that would normally use `cray-mpich` instead use this).
- [ ] Add an example+rocm which builds with MPI+rocm on Tioga or Tuolumne, verify it uses gtl @nhanford 
- [ ] Add an example+cuda which builds works with GPU-unaware MPI on matrix (openMPI or mvapich) @nhanford 
- [ ] Add or modify a test in CI @nhanford @michaelmckinsey1

### DIFF
--- a/repo/cray-mpich-gtl/package.py
+++ b/repo/cray-mpich-gtl/package.py
@@ -12,7 +12,7 @@ from spack.package import LinkTree
 from spack_repo.builtin.packages.mpich.package import MpichEnvironmentModifications
 
 
-class CrayMpichWrappers(MpichEnvironmentModifications, BundlePackage):
+class CrayMpichGtl(MpichEnvironmentModifications, BundlePackage):
 
     version("1.0.0")
 
--- a/repo/cray-mpich-gtl/package.py
+++ b/repo/cray-mpich-gtl/package.py
@@ -23,8 +23,14 @@ class CrayMpichGtl(MpichEnvironmentModifications, BundlePackage):
     def libs(self):
         return self.spec["cray-mpich"].libs
 
-    def setup_dependent_run_environment(self, env, dependent_spec):
-        env.set("MPICH_GPU_SUPPORT_ENABLED", "1")
+    def setup_dependent_build_environment(self, env, dependent_spec):
+        if dependent_spec.satisfies("+rocm"):
+            env.set("SPACK_GTL", "hip")
+        elif dependent_spec.satisfies("+cuda"):
+            env.set("SPACK_GTL", "cuda")
+
+    # def setup_dependent_run_environment(self, env, dependent_spec):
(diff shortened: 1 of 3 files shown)


scheibelp:
I think based on https://docs.nersc.gov/development/programming-models/mpi/cray-mpich/, if the original `cray-mpich` package just sets the `CRAY_ACCEL_TARGET` environment variable, that this will be unnecessary. This is being tested out in https://github.com/spack/spack-packages/pull/2097. I will reopen this PR if other efforts do not settle this matter.

repo/cray-mpich-gtl/package.py
```diff
@@ -0,0 +1,49 @@
...
+import stat
+from spack_repo.builtin.packages.mpich.package import MpichEnvironmentModifications
+from spack.package import LinkTree
+
+class CrayMpichWrappers(MpichEnvironmentModifications, BundlePackage):
+
+    version("1.0.0")
+
+    depends_on("cray-mpich", type="build")
+    provides("mpi")
```
nhanford:
```suggestion
    provides("mpi@3")
```
To match the underlying package.

repo/cray-mpich-wrappers/package.py
```diff
@@ -0,0 +1,49 @@
...
+
+        mkdir(self.prefix.bin)
+        for target in os.listdir(dep.prefix.bin):
+            if target in ["mpicc", "mpicxx", "mpif90", "mpif77"]:
+                fpath = os.path.join(self.prefix.bin, target)
+                with open(fpath, "w") as f:
+                    f.write(f"""\
+#!/bin/bash
+
+{dep.prefix.bin}/{target} -lmpi_gtl_hsa "$@"
```
nhanford: Should accommodate `-lmpi_gtl_cuda` as well...

pearce8 (approved):

PR #1159, opened 2025-12-10
Add support for PBS/Torque scheduler

Sbozzolo:
I'd like to try `benchpark` on a system with OpenPBS, but `benchpark` did not support such scheduler. This PR adds compatibility with it (to the best of my understanding of `benchpark`).


modifiers/allocation/modifier.py
```diff
@@ -466,13 +470,67 @@ def pjm_instructions(self, v):
         v.batch_submit = "pjsub {execute_experiment}"
         v.allocation_directives = "\n".join(batch_directives)
 
+    def pbs_instructions(self, v):
+        batch_opts, cmd_opts = Allocation._init_batch_and_cmd_opts(v)
+
+        if v.n_ranks and v.n_nodes and v.n_ranks_per_node:
+            expected_ranks = v.n_nodes * v.n_ranks_per_node
+            if v.n_ranks != expected_ranks:
```
scheibelp: Might get a complaint given that you are

```
        if not v.n_ranks_per_node and v.n_ranks and v.n_nodes:
            v.n_ranks_per_node = math.ceil(v.n_ranks / v.n_nodes)
```

modifiers/allocation/modifier.py
```diff
@@ -297,6 +297,10 @@ def determine_allocation(self, v):
         if not v.n_threads_per_proc:
             v.n_threads_per_proc = 1
 
+        # Calculate n_ranks_per_node if not explicitly set
+        if not v.n_ranks_per_node and v.n_ranks and v.n_nodes:
+            v.n_ranks_per_node = math.ceil(v.n_ranks / v.n_nodes)
```
scheibelp: If `pbs` always requires setting `n_ranks_per_node` (or in particular when `n_ranks` and `n_nodes` are set) I think that might best be done in `pbs_instructions` (based on reading it I assume it is not required, and in that case as well I think it might be best to skip this): this function implicitly ensures that either `n_nodes` or `n_ranks` is set by the end of it, and scheduler option functions tend to work with that; if this also sets n_ranks_per_node itself, it could over-specify (e.g. the potential mismatch from calling `ceil`).

Perhaps if `n_nodes` is set but not `n_ranks`/`n_ranks_per_node` then I should be setting `n_ranks_per_node` = `sys_cores_per_node` in this function. I assumed schedulers were doing that automatically; nearly all experiments in benchpark set `n_ranks`.

modifiers/allocation/modifier.py
```diff
@@ -466,13 +470,67 @@ def pjm_instructions(self, v):
...
+        if v.n_ranks:
+            cmd_opts.append(f"-np {v.n_ranks}")
+
+        if v.n_nodes:
+            node_spec = f"nodes={v.n_nodes}"
+            if v.n_ranks_per_node:
+                node_spec += f":ppn={v.n_ranks_per_node}"
+
+            if v.n_gpus and v.sys_gpus_per_node:
+                if v.n_gpus > v.n_nodes * v.sys_gpus_per_node:
```
scheibelp: This seems like a check I should have added to `determine_allocation` - would you mind moving this there?

Sbozzolo:
Sorry for the slow turnaround time. I wanted to implement the review comments and verify that the updated PR works as expected, but I haven't had time for that yet.

Sbozzolo:
I took a second look and read the manual more carefully (https://2026.help.altair.com/2026.0/PBS%20Professional/PBSUser2026.0.pdf, in particular section 5.5)

I tested this and it seem to work. The PBS model seems to be slightly different from other ones, as there's more per-node specification.

pearce8:
@scheibelp Please review the revisions.

modifiers/allocation/modifier.py
```diff
@@ -287,11 +287,17 @@ def determine_allocation(self, v):
             if v.n_gpus:
                 if v.sys_gpus_per_node:
                     gpus_node_request = math.ceil(v.n_gpus / float(v.sys_gpus_per_node))
+                    if gpus_node_request > v.sys_gpus_per_node:
```
scheibelp: `gpus_node_request` looks like the number of nodes you want (based on the number of gpus you ask for, and the number of gpus per node on the system)

In that case, I'm not sure what this comparison means: I read it as "if you are requesting more system nodes than there are gpus per node, then...", but can you let me know if that's off track?

I think this was adapted from

```
            if v.n_gpus and v.sys_gpus_per_node:
                if v.n_gpus > v.n_nodes * v.sys_gpus_per_node:
```

which I think I originally asked to move here, but perhaps the need for it was removed if because that particular check only arises when the allocation request is over-specified (e.g. you ask for x gpus, y nodes, and z gpus per node where you could have just asked for any two).

scheibelp (changes requested):
This looks good except for one issue I see with the check that was moved to `determine_allocation`. I think it would be ok to remove it if you find tweaking it to be more trouble than it's worth.

modifiers/allocation/modifier.py
```diff
@@ -466,13 +472,56 @@ def pjm_instructions(self, v):
...
+        batch_opts, cmd_opts = Allocation._init_batch_and_cmd_opts(v)
+
+        if not v.n_ranks_per_node:
+            v.n_ranks_per_node = math.ceil(v.n_ranks / v.n_nodes)
+
+        node_spec = [f"select={v.n_nodes}"]
+
+        if v.n_ranks:
+            cmd_opts.append(f"-np {v.n_ranks}")
+            node_spec.append(f"mpiprocs={v.n_ranks_per_node}")
```
scheibelp: Question: if the user sets n_ranks_per_node but not n_ranks, should part of this still execute?

modifiers/allocation/modifier.py
Sbozzolo: Yes, you are right. I think I got confused by what I meant with `gpus_node_request`

modifiers/allocation/modifier.py
Sbozzolo: Right, thank you!

scheibelp (approved):
Thanks!

PR #1167, opened 2025-12-15
Add amg performance testing params

rfhaque:
## Description
This PR
1. Adds performance testing parameters for amg2023 for the different testing modes
2. Changes the benchmarking scaling function to allow:
    a. All dimensions to be modified in a single scaling iteration, by returning a list of values from the scaling function
    b. No dimensions to be modified in the scaling iterations, by returning `None` from the scaling function. In this mode, it is presumed that the problem specification will provide the actual list of values for the variable being scaled. Variables `scaling-factor` and `scaling-iterations` are ignored in this mode.


pearce8 (changes requested):
@rfhaque 
- [ ] Are these parameters what we are using going forward?
- [ ] Please remove the parameters in the comments if they are not needed - or make them separate problems that are not comments
- [ ] Please fix lint

pearce8 (approved):

PR #1179, opened 2026-01-05
Fix spack buildcache push in github action

### LINKED ISSUE #1177: Broken spack buildcache push in github action
https://github.com/llnl/benchpark/actions/runs/20718270125/job/59474836532

michaelmckinsey1:
## Description

- [x] The saxpy action tries to push to the buildcache at ghcr.io/llnl/benchpark-binary-cache to speedup future CI runs. This is broken because the newer versions of spack no longer allow passing the password on the command line and instead expect an env var that contains the password (ref: https://github.com/spack/spack/blob/develop/lib/spack/docs/containers.rst?utm_source=chatgpt.com)

### DIFF
--- a/.github/workflows/run.yml
+++ b/.github/workflows/run.yml
@@ -103,11 +103,13 @@ jobs:
           path: ./workspace/generic-x86/saxpy/openmp/workspace/archive/**
       - name: Upload Binaries to CI Cache
         if: github.ref == 'refs/heads/develop'
+        env:
+          SPACK_OCI_USER: ${{ github.actor }}
+          SPACK_OCI_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          spack mirror set \
-            --push \
-            --oci-username ${{ github.actor }} \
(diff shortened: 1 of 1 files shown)


michaelmckinsey1:
Upload binaries step should pass in https://github.com/llnl/benchpark/actions/runs/20727740615/job/59508038953?pr=1179

.github/workflows/run.yml
```diff
@@ -102,12 +102,13 @@ jobs:
           name: workspace-archive
           path: ./workspace/generic-x86/saxpy/openmp/workspace/archive/**
       - name: Upload Binaries to CI Cache
-        if: github.ref == 'refs/heads/develop'
```
michaelmckinsey1, line 105: this needs to be put back before merging

michaelmckinsey1: Actually I cant test this because I will be `SPACK_OCI_USER: michaelmckinsey1`.

pearce8 (approved):

PR #1171, opened 2025-12-16
Cleanup RAJAPerf Package

michaelmckinsey1:
## Description

- [x] ~Add rocm `6.4.3` and `7.0.1` versions to `llnl-elcapitan`~ done in https://github.com/llnl/benchpark/pull/1174
- [x] Remove a majority of our `repo/raja-perf` in favor of upstream for future maintainability
- [x] requires updating `spack-packages` tag 
- [x] blocked by #1186

PR #1031, opened 2025-09-05
ScaFFold Benchmark and `spack-pip` Package Manager

michaelmckinsey1:
## Description

This PR is blocked by:
- [x] Application needs to be public https://github.com/LBANN/ScaFFold
- [x] Depends on https://github.com/LLNL/Adiak/pull/38 
   - [x] upstream the adiak package https://github.com/spack/spack-packages/pull/2300
- [x] (not mandatory) Needing to set `MPICH_GPU_SUPPORT_ENABLED=0` on coral2 otherwise GTL linker error
- [x] (ignore using roctx for now) -> (not mandatory) `ScaFFold+rocm caliper=rocm` hangs (for rocprof)
   - [x] ~Test spack package with rocprof on another benchmark (we were using roctracer before so maybe it is this)~ 
   - [x] ~run debugger~ 
    

Summary:
- [x] Add `spack-pip` package manager
   - [x] requires updating ramble tag to dev once https://github.com/GoogleCloudPlatform/ramble/pull/1248 merged
   - [x] Add infrastructure for new `spack-pip` mgr to benchpark
- [x] Add `py-scaffold` python benchmark, which has Caliper annotations via python bindings.
- [x] supersedes #687 
- [x] Update spack-packages tag to get updated version of pybind11 package.

Reproducers:
```
# Tuolumne profile rocm kernels and mpi functions
$ rm -rf tuolumne/ && benchpark system init --dest tuolumne llnl-elcapitan cluster=tuolumne && rm -rf py-scaffold-sp && benchpark experiment init --dest py-scaffold-sp      --system tuolumne py-scaffold+rocm scaffold_path=/usr/WS1/mckinsey/ScaFFold-spack/      package_manager=spack-pip caliper=mpi,time,rocm
$ rm -rf wkp && benchpark setup py-scaffold-sp wkp && . wkp/setup.sh && . wkp/.latest-experiment.sh 

# Matrix profile cuda kernels and mpi functions
$ rm -rf matrix/ && benchpark system init --dest matrix llnl-matrix && rm -rf py-scaffold-sp-mat/ && benchpark experiment init --dest py-scaffold-sp-mat      --system matrix py-scaffold+cuda scaffold_path=/usr/WS1/mckinsey/ScaFFold-spack/      package_manager=spack-pip caliper=mpi,time,cuda
$ rm -rf wkp2 && benchpark setup py-scaffold-sp-mat wkp2 && . wkp2/setup.sh  && . wkp2/.latest-experiment.sh 
```

[...]

PR #1186, opened 2026-01-13
Bump spack-packages version

michaelmckinsey1:
## Description

- [x] Update `spack-packages` version for #1171 and #1031
- [x] Make changes to fix broken tests due to the update
- [x] Add latest raja version `v2025.12.0`
- [x] Remove compiler flags from systems

### DIFF
--- a/repo/lammps/package.py
+++ b/repo/lammps/package.py
@@ -14,6 +14,9 @@ class Lammps(BuiltinLammps):
 
   depends_on("kokkos+openmp cxxstd=17", when="+openmp")
   depends_on("kokkos+wrapper", when="+cuda")
+
+  # Kokkos 5 not building
+  depends_on("kokkos@:4", when="@:20250722 +kokkos")
   
   flag_handler = build_system_flags
 
--- a/repo/sparta-snl/package.py
+++ b/repo/sparta-snl/package.py
@@ -53,7 +53,9 @@ class SpartaSnl(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("kokkos+openmp cxxstd=17", when="+openmp")
     depends_on("kokkos+rocm", when="+rocm")
     depends_on("kokkos+wrapper+cuda cxxstd=17", when="+cuda")
-  
+    # Kokkos 5 not building
+    depends_on("kokkos@:4", when="+kokkos")
+
     depends_on("jpeg", when="+jpeg")
     depends_on("libpng", when="+png")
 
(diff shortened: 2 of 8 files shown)


michaelmckinsey1:
Only expected failures are `[matrix, llnl-matrix, lammps, +cuda]` and `[matrix, llnl-matrix, sparta-snl, +cuda~gpu-aware-mpi]`. Will need to look into other failures

systems/llnl-elcapitan/system.py
```diff
@@ -770,7 +770,7 @@ def rocm_cce_compiler_cfg(self):
                 },
                 flags={
                     "cflags": "-g -O2",
-                    "cxxflags": "-g -O2 -std=c++14",
+                    "cxxflags": "-g -O2 -std=c++17",
```
michaelmckinsey1: @scheibelp This change was necessary to fix https://lc.llnl.gov/gitlab/benchpark/benchpark/-/jobs/3264976, since the new version of `camp` requires `c++17`. But it does not seem correct to set this here when this is package dependent right? What should we do? Should we just remove it and let packages specify?

note `snl-eldorado` has the same issue

repo/lammps/package.py
```diff
@@ -14,6 +14,9 @@ class Lammps(BuiltinLammps):
 
   depends_on("kokkos+openmp cxxstd=17", when="+openmp")
   depends_on("kokkos+wrapper", when="+cuda")
+
+  # Kokkos 5 not building
+  depends_on("kokkos@:4", when="@:20250722 +kokkos")
```
michaelmckinsey1, line 19: @rfhaque Is this expected to be necessary?

repo/sparta-snl/package.py
```diff
@@ -53,7 +53,9 @@ class SpartaSnl(CMakePackage, CudaPackage, ROCmPackage):
     depends_on("kokkos+openmp cxxstd=17", when="+openmp")
     depends_on("kokkos+rocm", when="+rocm")
     depends_on("kokkos+wrapper+cuda cxxstd=17", when="+cuda")
-  
+    # Kokkos 5 not building
+    depends_on("kokkos@:4", when="+kokkos")
```
michaelmckinsey1, line 57: @rfhaque Is this expected to be necessary?

systems/llnl-elcapitan/system.py
scheibelp: > Should we just remove it and let packages specify?

I think this is the best option (and run a pipeline): It might some constraints/options need to be added to the package definitions but I expect that would be a better way to manage standards.

michaelmckinsey1: We decided to remove the compiler flags from systems and let packages set flags only

pearce8 (approved):

PR #1006, opened 2025-09-02
WIP: Add Lanl CTS-2 system rocinante to system definitions

sriram-LANL:
## Adding/modifying a system (docs: [Adding a System](https://software.llnl.gov/benchpark/add-a-system-config.html))
- [x] Add/modify `systems/lanl-rocinante/system.py` file
- [x] Add/modify `systems/all_hardware_descriptions/hardware_name/hardware_description.yaml` which will appear in the [docs catalogue](https://software.llnl.gov/benchpark/system-list.html)

PR #1188, opened 2026-01-14
Add system config for the LANL Rocinante cluster

rfhaque:
## Description
This PR add the system configuration for the LANL Rocinante cluster. Supersedes https://github.com/llnl/benchpark/pull/1006

- [x] Add/modify `systems/all_hardware_descriptions/hardware_name/hardware_description.yaml` which will appear in the [docs catalogue](https://software.llnl.gov/benchpark/system-list.html)


sriram-LANL:
@rfhaque  I was able to configure and run `kripke+strong` from the tutorial.  I think this works and can be merged.

kripke_mpi_strong_raw_exc

pearce8 (dismissed):

systems/lanl-rocinante/system.py
```diff
@@ -0,0 +1,264 @@
...
+    maintainers("sriram")
+
+    id_to_resources = {
+        "rocinante": {
+            "sys_cores_per_node": 112,
+            "sys_cores_os_reserved_per_node": 0,  # No core or thread reservation
+            "sys_cores_os_reserved_per_node_list": None,
+            "system_site": "lanl",
+            "hardware_key": str(hardware_descriptions)
+            + "HPECray-sapphirerapids-Slingshot/hardware_description.yaml",
```
sriram-LANL: @rfhaque This should start with a `/` i.e. `"/HPECray-..."` for the tests to pass.   I apologize, this is a mistake I made in the original file.

sriram-LANL:
@pearce8 @rfhaque the tests are still failing because `id_to_resource` at line 23 doesn't have entries for "tycho" and "crossroads".  These entries are essentially identical to the rocinante entry except for memory per node and memory type which is not really captured in the structure.

I'm not sure how to proceed here - duplicating the rocinante entry works but I believe is the wrong answer.  Should we just call this variant "CTS2" or is there a better way?

pearce8:
@sriram-LANL I just added memory entries to the system description, and duplicated them for the other 2 systems.  We are tracking memory sizes and cache sizes, in case some benchmarks are defined in terms of those (e.g., Stream needs to be run at 3x the size of the last level cache).  Please take a look and update the numbers if you can.  We do not currently differentiate between memory types, maybe that's something to consider for the future.

pearce8 (approved):

PR #1190, opened 2026-01-16
Custom repos in benchpark

scheibelp:
Define a `repos.yaml` like:

```
repos:
  experiments:
  - ../experiments
  - ../experiments/test
  # and systems, applications, and packages, see config/repos.yaml added in this PR for full example
```

in `<benchpark_root>/config` to customize where experiment definitions are pulled from.

Benchpark can pull config from one location, with the following priority (highest first):

* `benchpark -C <dir>...`
* If CWD where you call `benchpark` has `benchpark-config` directory
* If `<benchpark_root>/config` is a directory (this must exist if the first two don't)

More generally:

* Add a rudimentary YAML-based config system to benchpark
* Define a `Repos` section
  * Benchpark now uses this repos config to instantiate experiments and systems repos for benchpark itself, and packages/applications repos for Spack/Ramble
* Accommodates other possible sections
  * These sections support property-like access in python (but that assume the property exists, which they all should in the case of repos config)

TODOs

- [x] several files are called `repos.yaml` or `repo.yaml` w/ different purposes, maybe I should call this new one something different
- [ ] schema validation
- [x] Consolidate with benchpark bootstrapping config (currently gets consumed via a different mechanism, lives in a file in `configure.yaml`)


config/repos.yaml
```diff
@@ -0,0 +1,5 @@
+repos:
```
michaelmckinsey1, line 5:
```suggestion
# Copyright 2023 Lawrence Livermore National Security, LLC and other
# Benchpark Project Developers. See the top-level COPYRIGHT file for details.
#
# SPDX-License-Identifier: Apache-2.0
repos:
```

lib/main.py
```diff
@@ -7,48 +7,23 @@
...
-    info                Get information about Systems and Experiments
-    show-build          Show how spack built a benchmark
-    list                List experiments, systems, benchmarks, and modifiers
-    bootstrap           Bootstrap benchpark or update an existing bootstrap
-    analyze             Perform pre-defined analysis on the performance data (caliper files) after 'ramble on'
-    configure           Configure options relating to the Benchpark environment
-    """
-if len(sys.argv) == 1 or "-h" == sys.argv[1] or "--help" == sys.argv[1]:
-    print(helpstr)
-    exit()
```
michaelmckinsey1, line 47: This is to prevent benchpark from bootstrapping when running `help` or `version`

https://github.com/llnl/benchpark/pull/822#issuecomment-2960605283

michaelmckinsey1 (changes requested):
From our discussion

Regarding the `benchpark configure` comand - take `benchpark.yaml` rename it to `bootstrap.yaml` benchpark configure will write to the `config` directory.

lib/main.py
```diff
@@ -39,19 +40,42 @@
     list                List experiments, systems, benchmarks, and modifiers
     bootstrap           Bootstrap benchpark or update an existing bootstrap
     analyze             Perform pre-defined analysis on the performance data (caliper files) after 'ramble on'
+    redo                Re-instantiate all experiments in a system
```
michaelmckinsey1: I know redo is unrelated here, but doesn't it also re-instantiate the system (not just the experiments)?

config/repos.yaml
michaelmckinsey1, line 1: Regarding the naming of this file - it makes sense to not be `repo.yaml`, as that would conflate with the fact that this is pointing at multiple `repo.yaml`'s. Therefore, I think `repos.yaml` is a decent name, and I can't think of a better one.

lib/main.py
```diff
@@ -39,19 +40,42 @@
...
+# Later imports initiate bootstrapping, and the configure command may want
+# to configure this behavior, so we handle it before doing remaining imports
+if "configure" in sys.argv:
     import benchpark.cmd.configure  # noqa: E402
 
     parser = argparse.ArgumentParser(description="Benchpark")
+    parser.add_argument(
+        "-C",
+        "--config",
+        help="use config related to system/experiment/workspace generation",
```
michaelmckinsey1:
```suggestion
        help="Use config related to system/experiment/workspace generation. Default is 'benchpark/config/'",
```

lib/main.py
```diff
@@ -93,6 +120,11 @@ def main():
     parser.add_argument(
         "-V", "--version", action="store_true", help="show version number and exit"
     )
+    parser.add_argument(
+        "-C",
+        "--config",
+        help="use config related to system/experiment/workspace generation",
+    )
```
michaelmckinsey1, line 134: This is the second `-C` (one above). So the user can either: 
1. `benchpark -C config/ configure ...` configure a specific config other than the default
2. `benchpark -C config/ ...` use a config
right? Then maybe these should have different help strings

michaelmckinsey1 (changes requested):
(1) I think we need docs to instruct users to add the path to their `experiments`/`systems`/`repo` to `config/repos.yaml`. And that if they wish for their addition to be used over the existing version, it should go first in the config.

(2) Can you merge develop and run a gitlab pipeline?

lib/main.py
scheibelp: This one is kind of weird: we actually handle `-C` above in the import section, but if we don't include this here, then `argparse` will consider it an invalid option: I should add a comment about that.

scheibelp: (comment now added)

docs/configuration.rst
```diff
@@ -0,0 +1,53 @@
+..
+    Copyright 2023 Lawrence Livermore National Security, LLC and other
+    Benchpark Project Developers. See the top-level COPYRIGHT file for details.
+
+    SPDX-License-Identifier: Apache-2.0
+
+###############
+ Configuration
```
michaelmckinsey1:
```suggestion
 Configuring Benchpark
```

docs/index.rst
```diff
@@ -14,6 +14,7 @@
     getting-started
     benchpark-commands
     benchpark-workflow
+    configuration
```
michaelmckinsey1: In my opinion, this should be at the top of the `Using Benchpark` section, as it is not basic usage.

docs/configuration.rst
```diff
@@ -0,0 +1,53 @@
...
+###############
+
```
michaelmckinsey1, line 21:
```suggestion

Benchpark offers several options to configure usage. This includes:
   1. Configuring the location of the bootstrap directory (default is ``~/.benchpark``). This is useful to change from the default value if you do not have much space in your home directory, as the bootstrap contains the ``spack``, ``spack-packages``, and ``ramble`` repositories.
   1. Configuring which repositories benchpark uses for objects in the ``experiments/systems/repos`` directories (``experiment.py``, ``system.py``, ``package.py``, and ``application.py``).

```

docs/configuration.rst
```diff
@@ -0,0 +1,53 @@
...
+Benchpark clones Ramble to use as a library automatically as part of running. It clones
+both Ramble and Spack to allow building and running benchmarks. Benchmarks are organized
+into workspaces, which have their own clones of Spack/Ramble, and these are staged from
+a centralized location. The default is in ~/.benchpark, but you can pick another
+location if desired. This default can be set in
+``<benchpark_root>/config/bootstrap.yaml``.
+
+*******
+ Repos
```
michaelmckinsey1:
```suggestion
 Configuring Which Repositories Benchpark Uses
```

docs/configuration.rst
```diff
@@ -0,0 +1,53 @@
...
+    Benchpark Project Developers. See the top-level COPYRIGHT file for details.
+
+    SPDX-License-Identifier: Apache-2.0
+
+###############
+ Configuration
+###############
+
+********************
+ Bootstrap Location
```
michaelmckinsey1:
```suggestion
 Configuring the Benchpark Bootstrap Location
```

docs/configuration.rst
```diff
@@ -0,0 +1,53 @@
...
+********************
+
+Benchpark clones Ramble to use as a library automatically as part of running. It clones
+both Ramble and Spack to allow building and running benchmarks. Benchmarks are organized
+into workspaces, which have their own clones of Spack/Ramble, and these are staged from
+a centralized location. The default is in ~/.benchpark, but you can pick another
+location if desired. This default can be set in
+``<benchpark_root>/config/bootstrap.yaml``.
```
michaelmckinsey1:
```suggestion
Benchpark clones ``ramble``, ``spack``, and ``spack-packages`` into a centralized location (by default this is ``~/.benchpark``) to enable building and running benchmarks. Benchpark workspaces copy over these repositories from the centralized location to avoid downloading from the internet when initializing a Benchpark workspace. The bootstrap location can be configured in ``<benchpark_root>/config/bootstrap.yaml`` or by running ``benchpark configure --bootstrap-location <location>``.
```

docs/configuration.rst
```diff
@@ -0,0 +1,53 @@
...
+
+*******
+ Repos
+*******
+
+See ``<benchpark_root>/config/repos.yaml``: with this you can fully customize
+system/experiment repos used by Benchpark, application repos used by Ramble, and package
+repositories used by Spack. Note in the case of Spack, the builtin repository is always
+used (in addition to whatever package repositories are specified).
```
michaelmckinsey1:
```suggestion
The ``<benchpark_root>/config/repos.yaml`` file is used to fully customize
``system``/``experiment`` repositories used by Benchpark, ``application`` repositories used by Ramble, and package repositories used by Spack. For Spack and Ramble, the builtin repository is always
used (in addition to whatever package repositories are specified). The order in which repositories appear in ``<benchpark_root>/config/repos.yaml`` will determine their precedence, so if you desire to use a custom repository, it should be listed before any other repositories.
```

Is the last sentence true (double checking)?

docs/configuration.rst
```diff
@@ -0,0 +1,53 @@
...
+********
+ Scopes
+********
+
+Benchpark can pull config from one location, with the following priority (highest
+first):
+
+- ``benchpark -C <dir>...``
+- If CWD where you call benchpark has benchpark-config directory
+- If ``<benchpark_root>/config`` is a directory (this must exist if the first two don't)
```
michaelmckinsey1:
```suggestion
1. ``benchpark -C <dir>...``
1. If the CWD where you invoke the benchpark executable has ``benchpark-config`` directory
1. If ``<benchpark_root>/config`` is a directory (this must exist if the first two don't)
```

is number 2 ``benchpark-config`` or just ``config`` directory?

docs/configuration.rst
```diff
@@ -0,0 +1,53 @@
...
+
+Right now it can only generate bootstrap config. If no bootstrap config is detected in
+the chosen scope, this will auto-generate it.
+
+********
+ Scopes
+********
+
+Benchpark can pull config from one location, with the following priority (highest
+first):
```
michaelmckinsey1:
```suggestion
Benchpark can pull configurations from one location, with the following priority (highest
first):
```

docs/configuration.rst
```diff
@@ -0,0 +1,53 @@
...
+used (in addition to whatever package repositories are specified).
+
+*************************
+ ``benchpark configure``
+*************************
+
+Command that can generate config for you.
+
+Right now it can only generate bootstrap config. If no bootstrap config is detected in
+the chosen scope, this will auto-generate it.
```
michaelmckinsey1:
```suggestion
```
IMO we don't need to explain this command on this page.

docs/configuration.rst
```diff
@@ -0,0 +1,53 @@
...
+
+********
+ Scopes
```
michaelmckinsey1:
```suggestion
 Configuration Scopes
```

michaelmckinsey1 (changes requested):
Some docs edits

docs/configuration.rst
scheibelp: If its in the CWD, it must be called `benchpark-config`, if you put that directory in `benchpark_root`, then that will have precedence (but I should probably disallow that)

docs/index.rst
scheibelp: I put it in the `Using Benchpark` section, but a few items down (under items I think would be more-commonly of interest)

docs/configuration.rst
scheibelp: I think this is a useful context to explain that not all config needs to be manually generated, I think I avoided duplicating explanations. I could add a pointer?

docs/index.rst
michaelmckinsey1: Ok that works too

docs/configuration.rst
michaelmckinsey1: As long as it is accurate, I think this is fine

docs/configuration.rst
```diff
@@ -0,0 +1,68 @@
...
+listed before any other repositories.
+
+*************************
+ ``benchpark configure``
+*************************
+
+Command that can generate config for you.
+
+Right now it can only generate bootstrap config. If no bootstrap config is detected in
+the chosen scope, this will auto-generate it.
```
michaelmckinsey1: rephrasing

```suggestion
*************************
 Automatically Generating Configurations with ``benchpark configure``
*************************

``benchpark configure`` is designed to create the ``yaml`` configurations for you. As of now, it can only generate bootstrap config (``bootstrap.yaml``). If no bootstrap config is detected in the chosen scope, this will auto-generate it. ``benchpark configure`` cannot generate a ``repos.yaml`` file at the moment, this must be written manually.


```

docs/configuration.rst
michaelmckinsey1: Maybe I will rephrase so it fits better on the page then

michaelmckinsey1 (approved):
Good now

PR #1217, opened 2026-01-30
Enable CHAI only when running Kripke on GPU

rfhaque:
## Description
This PR adds the check to only enable CHAI in Kripke when running on the GPU.
CHAI is disabled when running on CPU

## Adding/modifying a benchmark (docs: [Adding a Benchmark](https://software.llnl.gov/benchpark/add-a-benchmark.html))

- [x] If package.py upstreamed to Spack is insufficient, add/modify `repo/benchmark_name/package.py` plus: create, self-assign, and link here a follow up issue with a link to the PR in the Spack repo.

PR #1219, opened 2026-01-30
Caliper: Use libpfm over papi for topdown and add error message if wrong uarch

michaelmckinsey1:
## Description

- [x] This is recommended in latest caliper release https://github.com/llnl/Caliper/blob/master/doc/sphinx/Topdown.rst

`The Linux perf based method is the preferred approach. It requires Caliper to be built with the -DWITH_LIBPFM=On switch or the +libpfm Spack variant.`

- [x] Depends on #1217 
- [x] hardcode uarch and add error message
   
   
Example error if running topdown on elcap
```
ValueError: Topdown analysis is not supported on the provided system with uarch='zen4'. Must be one of:
        ['sapphirerapids', 'emeraldrappids']
```

### DIFF
--- a/systems/llnl-cluster/system.py
+++ b/systems/llnl-cluster/system.py
@@ -3,6 +3,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+import yaml
 
 from benchpark.directives import maintainers, variant
 from benchpark.openmpsystem import OpenMPCPUOnlySystem
@@ -46,7 +47,6 @@ class LlnlCluster(System):
             "mount_points": ["/l/ssd", "/p/lustre1", "/p/lustre2", "/p/lustre3"],
         },
         "dane": {
-            "cpu_arch": "sapphirerapids",
             "sys_cores_per_node": 112,
             "sys_cores_os_reserved_per_node": 0,  # No explicit core reservation, first thread on each core reserved (2 threads per core)
             "sys_cores_os_reserved_per_node_list": None,
@@ -142,6 +142,9 @@ class LlnlCluster(System):
         attrs = self.id_to_resources.get(self.spec.variants["cluster"][0])
         for k, v in attrs.items():
             setattr(self, k, v)
+
+        with open(self.hardware_key, "r") as f:
+            self.hardware_dict = yaml.safe_load(f)
 
         mount_point = self.spec.variants["mount_point"][0]
         if mount_point not in self.mount_points + ["none"]:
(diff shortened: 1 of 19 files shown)


pearce8:
@daboehme What are the trade offs here?  Are we expecting this to work on more systems, or at least not fewer?  Is the overhead guaranteed to be not higher than PAPI?  Is the data/measurements the same as the PAPI counters?

@michaelmckinsey1 We need to at least see which of our systems this works on.  And compare rajaperf topdown with papi and with libpfm.

daboehme:
> @daboehme What are the trade offs here? Are we expecting this to work on more systems, or at least not fewer? Is the overhead guaranteed to be not higher than PAPI? Is the data/measurements the same as the PAPI counters?

The perf topdown service does essentially the same thing that the PAPI topdown component does. The reason we added this was because (a) certain individuals don't like PAPI in their codes, and (b) in the 7.2.0 PAPI release the topdown component is incomplete/broken and doesn't allow for attribution of topdown values to regions.

The measurements are the same we get with the PAPI component, and the supported CPUs are also the same (Ivy Bridge or higher for level 1 metrics, Sapphire Rapids or higher for level 2 metrics). Both use the fixed counters in those CPUs. The aggregation on the Caliper side is slightly better for perf than for PAPI (essentially, with PAPI we just average the percentages for all measurements per region, whereas with perf we scale by number of slots, which should be more accurate).

Overheads are also roughly the same and generally manageable as Michael showed yesterday.

systems/llnl-cluster/system.py
```diff
@@ -46,7 +47,6 @@ class LlnlCluster(System):
             "mount_points": ["/l/ssd", "/p/lustre1", "/p/lustre2", "/p/lustre3"],
         },
         "dane": {
-            "cpu_arch": "sapphirerapids",
```
michaelmckinsey1, line 49: Loading the yaml in each system.py is necessary to avoid recording this duplicate information. I had only hardcoded this for dane, but at least `cpu_arch` and `vendor` would be necessary to hardcode for every system, if not for loading the yamls

pearce8: What error will people get if the yaml load fails?

I think it is better to require setting of the `cpu_arch` and `vendor` and check this when committing hardware specs and output an appropriate error.

pearce8 (approved):
Please rework this to require CPU arch and vendor in the hardware description.

PR #1226, opened 2026-02-03
only set MPICH_GPU_SUPPORT_ENABLED to 1 when a package linked gtl lib

scheibelp:
The `llnl-elcapitan` system uses `cray-mpich-gtl` as the `mpi` provider. It was always setting `MPICH_GPU_SUPPORT_ENABLED=1`, even for packages where it did not `-lmpi_gtl_hsa` (for example `hpl`). This PR only activates that env var for packages that are `+rocm` (the same gate that controls whether `-lmpi_gtl_hsa` is linked by the wrapper).

### DIFF
--- a/repo/cray-mpich-gtl/package.py
+++ b/repo/cray-mpich-gtl/package.py
@@ -30,7 +30,7 @@ class CrayMpichGtl(MpichEnvironmentModifications, BundlePackage):
             env.set("SPACK_GTL", "cuda")
 
     def setup_dependent_run_environment(self, env, dependent_spec):
-        if dependent_spec.satisfies("+rocm"):
+        if dependent_spec.satisfies("+rocm") or dependent_spec.satisfies("+cuda"):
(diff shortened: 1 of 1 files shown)


repo/cray-mpich-gtl/package.py
```diff
@@ -30,7 +30,8 @@ def setup_dependent_build_environment(self, env, dependent_spec):
             env.set("SPACK_GTL", "cuda")
 
     def setup_dependent_run_environment(self, env, dependent_spec):
-        env.set("MPICH_GPU_SUPPORT_ENABLED", "1")
+        if dependent_spec.satisfies("+rocm"):
```
nhanford:
```suggestion
        if dependent_spec.satisfies("+rocm") or dependent_spec.satisfies("+cuda"):
```

nhanford (changes requested):

pearce8 (approved):

PR #1266, opened 2026-02-27
[Do not merge] Hip wrapper

scheibelp:
- [ ] Intended to be deprecated 
- [ ] https://github.com/spack/spack-packages/pull/4411
- [ ] https://github.com/spack/spack-packages/pull/2965
- [ ] https://github.com/spack/spack-packages/pull/2499
- [X] Fixes https://github.com/llnl/benchpark/issues/1259

Add a mechanism to perform spack compiler wrapper modifications to hip compiler invocations. Right now, depends on packages explicitly shelling out to an artificial `hip-wrapper` package. That can potentially be hidden from users.

PR #1278, opened 2026-03-04
[1277] - Add remaining Laghos problems to benchpark

### LINKED ISSUE #1277: Add remaining Laghos experiments to Benchpark with any parameters
The goal is to add all the eight problems that Laghos has to Benchpark along with any configurations/arguments that we could benefit from when benchmarking.

amroakmal:
Resolves #1277

### DIFF
--- a/experiments/laghos/experiment.py
+++ b/experiments/laghos/experiment.py
@@ -24,7 +24,7 @@ class Laghos(
     variant(
         "workload",
         default="sedov",
-        values=("sedov", "triplept"),
+        values=("taylor-green", "sedov", "1D-sod-shock", "triple-point", "gresho-vortex", "2D-riemann-12", "2D-riemann-6", "2D-rayleigh-taylor"),
         description="problem type",
     )
 
@@ -55,6 +55,46 @@ class Laghos(
         description="nonconforming or conforming",
     )
 
+    variant(
+        "vis",
+        default=False,
+        values=(True, False),
+        description="Enable or disable GLVis visualization",
+    )
+
+    variant(
+        "s",
+        default="4",
+        values=("1", "2", "3", "4", "5", "6", "7"),
+        description="ODE solver type",
+    )
+
+    variant(
+        "dim",
+        default="3",
+        values=("1", "2", "3"),
+        description="Dimension of the problem",
+    )
+
+    variant(
+        "cfl",
+        default="0.5",
+        description="CFL-condition number",
+    )
+
+    variant(
+        "cgm",
+        default="0.1",
+        description="Maximum number of CG iterations (velocity linear solve)",
+    )
+
+    variant(
+        "cgt",
+        default="1e-8",
+        description="Relative CG tolerance (velocity linear solve)",
+    )
+
+
     maintainers("wdhawkins")
 
     def generate_perf_specs(self):
@@ -69,6 +109,12 @@ class Laghos(
             "strong": None,
             "weak": None,
             "throughput": None,
+            "s": 4,
+            "dim": 3,
+            "cfl": 0.5,
+            "cgm": 300,
+            "cgt": 1e-8,
+            "vis": False
         }
         # Add problem specs as needed here
         if self.spec.satisfies("+throughput"):
@@ -120,6 +166,14 @@ class Laghos(
         self.add_experiment_variable("ms", problem_spec["ms"], True)
         self.add_experiment_variable("tf", problem_spec["tf"], True)
 
+        self.add_experiment_variable("vis", problem_spec["vis"], True)
+        self.add_experiment_variable("s", problem_spec["s"], True)
+        self.add_experiment_variable("dim", problem_spec["dim"], True)
+        self.add_experiment_variable("cfl", problem_spec["cfl"], True)
+        self.add_experiment_variable("cgm", problem_spec["cgm"], True)
+        self.add_experiment_variable("cgt", problem_spec["cgt"], True)
+        
+
         self.add_experiment_variable(
             "resource_count", problem_spec["resource_count"], True
         )
@@ -143,6 +197,7 @@ class Laghos(
                 },
             }
         )
+
 
     def compute_applications_section(self):
         if self.spec.satisfies("exec_mode=perf"):
@@ -229,6 +284,15 @@ class Laghos(
                 self.add_experiment_variable("gam", "--no-gpu-aware-mpi")
         else:
             self.add_experiment_variable("n_ranks", "{n_resources}", True)
+
+        if self.spec.satisfies("vis=True"):
+            self.add_experiment_variable("vis", True, True)
+
+        params_to_inspect = ["s", "dim", "cfl", "cgm", "cgt"]
+        for param in params_to_inspect:
+            if param in self.spec.variants:
+                self.add_experiment_variable(param, self.spec.variants[param][0], True)
+
 
     def compute_package_section(self):
         gam = "~gpu-aware-mpi"
--- a/experiments/laghos/experiment.py
+++ b/experiments/laghos/experiment.py
@@ -197,7 +197,7 @@ class Laghos(
             self.add_experiment_variable("nx", 1, True)
             self.add_experiment_variable("ny", 1, True)
             self.add_experiment_variable("nz", 1, True)
-            
+
             self.add_experiment_variable("rs", 3, True)
             self.add_experiment_variable("rp", 2, True)
             self.add_experiment_variable(
--- a/experiments/laghos/experiment.py
+++ b/experiments/laghos/experiment.py
@@ -112,9 +112,6 @@ class Laghos(
 
     def generate_perf_specs(self):
         problem_spec = {
-            "nx": 1,
-            "ny": 1,
-            "nz": 1,
             "pool_size": 16,
             "resource_count": 4,
             "strong": None,
@@ -198,9 +195,6 @@ class Laghos(
             self.generate_perf_specs()
         else:
             # "zones" defined from mesh file, we are hardcoding it here
-            self.add_experiment_variable("nx", 1, True)
-            self.add_experiment_variable("ny", 1, True)
-            self.add_experiment_variable("nz", 1, True)
             self.add_experiment_variable("rs", 3, True)
             self.add_experiment_variable("rp", 2, True)
             self.add_experiment_variable(
--- a/experiments/laghos/experiment.py
+++ b/experiments/laghos/experiment.py
@@ -72,13 +72,6 @@ class Laghos(
     )
 
     variant(
-        "s",
-        default="4",
-        values=("1", "2", "3", "4", "5", "6", "7"),
-        description="ODE solver type",
-    )
-
-    variant(
         "dim",
         default="3",
         values=("1", "2", "3"),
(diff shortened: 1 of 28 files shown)


amroakmal:
Update: I confirm testing the two remaining problems and they both finished successfully. I tested with a backup branch forked from this feature branch and this forked branch has been rebased on `hip-wrapper` branch. But this feature branch here does not have `hip-wrapper` feature branch logic/changes as that should be handled in its own PR.

=====

Update: Rebased on hip-wrapper, still facing same issue:
```
laghos: error while loading shared libraries: libunwind.so.1: cannot open shared object file: No such file or directory
flux-job: task(s) exited with exit code 127
```
=====

This has been tested for 6/8 different Laghos problems with different configs and params from what we added. It's ready for review, but I'll make sure its not merged until the remaining two problems are tested, blocked by the `hip-wrapper` PR.

experiments/laghos/experiment.py
```diff
@@ -55,6 +64,45 @@ class Laghos(
         description="nonconforming or conforming",
     )
 
+    variant(
+        "vis",
+        default=False,
+        values=(True, False),
+        description="Enable or disable GLVis visualization",
+    )
```
michaelmckinsey1, line 79: Did you test this for `vis=True`? What is the result?

repos/ramble_applications/laghos/application.py
```diff
@@ -18,8 +18,65 @@ class Laghos(ExecutableApplication):
...
+                       ' -d {device}' +
+                       ' -s {s}' +
+                       ' -dim {dim}' +
+                       ' -cfl {cfl}' +
+                       ' -cgm {cgm}' +
+                       ' -cgt {cgt}' +
+                       ' {vis}' +
+                       ' {assembly}',
+                       use_mpi=True)
+
```
michaelmckinsey1: Is the only thing being varied across workloads `-p`?

Take a look at https://github.com/llnl/benchpark/blob/develop/repos/ramble_applications/ior/application.py at how the common params can be a shared variable.

So I think every workload should be `'laghos' + ' -p [some_value]' + shared_args`, where `shared_args` is everything else

experiments/laghos/experiment.py
```diff
@@ -55,6 +64,45 @@ class Laghos(
...
+        "cgm",
+        default="300",
+        description="Maximum number of CG iterations (velocity linear solve)",
+    )
+
+    variant(
+        "cgt",
+        default="1e-8",
+        description="Relative CG tolerance (velocity linear solve)",
+    )
```
michaelmckinsey1, line 98: I'm not convinced these should be variants. Variants are parameters that are problem independent, like you did for `vis` (`vis` should correctly be a variant).

Parameters that impact performance should be set for the different modes `exec_mode=`. In laghos there are two, `exec_mode=perf` and `exec_mode=test` (`test` is the default). Some parameters will be the same for the two modes, but something like the problem size should be larger in the `perf` mode.

note how `nx`, `ny`, `nz` are not variants

experiments/laghos/experiment.py
```diff
@@ -69,6 +117,12 @@ def generate_perf_specs(self):
             "strong": None,
             "weak": None,
             "throughput": None,
+            "s": 4,
+            "dim": 3,
+            "cfl": 0.5,
+            "cgm": 300,
+            "cgt": 1e-8,
+            "vis": False,
```
michaelmckinsey1: since `vis` is a variant this is unecessary to set here (`"vis": False,`)

experiments/laghos/experiment.py
```diff
@@ -120,6 +174,13 @@ def generate_perf_specs(self):
         self.add_experiment_variable("ms", problem_spec["ms"], True)
         self.add_experiment_variable("tf", problem_spec["tf"], True)
 
+        self.add_experiment_variable("vis", problem_spec["vis"], True)
```
michaelmckinsey1: This will be overwritten by line 294, so it can be removed anyway

experiments/laghos/experiment.py
```diff
@@ -230,6 +291,16 @@ def compute_applications_section(self):
...
 
+        if self.spec.satisfies("+vis"):
+            self.add_experiment_variable("vis", "-vis", True)
+        else:
+            self.add_experiment_variable("vis", "-no-vis", True)
+
+        params_to_inspect = ["s", "dim", "cfl", "cgm", "cgt"]
+        for param in params_to_inspect:
+            if param in self.spec.variants:
+                self.add_experiment_variable(param, self.spec.variants[param][0], True)
```
michaelmckinsey1: This will not exist once these are removed as variants. I'm not sure if `s` should be a variant, as the solver. My understanding is the solver is completely independent of the other parameters correct (i.e. we shouldn't need to change the solver at a certain problem size)?  If this is the case this loop still doesn't need to exist, you can instead do
```
self.add_experiment_variable("s", self.spec.variants["s"][0], True)
```

repos/ramble_applications/laghos/application.py
```diff
@@ -18,8 +18,65 @@ class Laghos(ExecutableApplication):
...
+                       ' -dim {dim}' +
+                       ' -cfl {cfl}' +
+                       ' -cgm {cgm}' +
+                       ' -cgt {cgt}' +
+                       ' {vis}' +
+                       ' {assembly}',
+                       use_mpi=True)
 
-    executable('triplept', 'laghos' +
+    executable('triple-point', 'laghos' +
```
michaelmckinsey1: I agree that `triple-point` is better than `triplept`.

michaelmckinsey1 (changes requested):
Since this is working for everything but rocm/7, I would say this is not blocked by the hip-wrapper branch. Because we can still use it on the other architectures and rocm6.

So once this PR is ready, I think we can merge it independent of https://github.com/llnl/benchpark/pull/1266. Also @scheibelp said 1266 won't likely get merged anyway, it will be fixed in an upstream spack PR (#49673), which will take longer than the benchpark PR to merge.

experiments/laghos/experiment.py
amroakmal: Yes, I tested, what I looked for making it enabled and made sure its reflected correctly in the generated experiment script, but no files were generated as output. My understanding if this flag is enabled, then there could a like a live-visualization for the experiment running. I will look if there is a way to persist the visualization, I'm not sure about it though.

experiments/laghos/experiment.py
amroakmal: Can you elaborate more, I dont understand how doing that will make users be able to modify the values they want such as cgt param for example? I mean, for exec=test I can set the default values (following what Laghos provide), but if someone wants to modify a certain value, then you mean they should do that manually after the script is generated, correct? I thought we need to offer this to users to do so at `benchpark experiment init ...`
Please let me know what you prefer and what matches BP standards.

pearce8: Lets not make mathematical changes, so lets not turn tolerances or condition numbers into variants.  Also max number of cg iterations.

We definitely need to study all 3 problem dimensions (they will have dramatically different performance) and all polynomial degrees (the `-p`), so please make those variants.

Numbers of steps, final time, should all be variants with good defaults.

As for ODE solver type, we need to find out if all types are valid for all problems.  If you are unable to find answers to this, please remind me to figure out who to ask.

experiments/laghos/experiment.py
```diff
@@ -120,6 +174,13 @@ def generate_perf_specs(self):
         self.add_experiment_variable("ms", problem_spec["ms"], True)
         self.add_experiment_variable("tf", problem_spec["tf"], True)
 
+        self.add_experiment_variable("vis", problem_spec["vis"], True)
+        self.add_experiment_variable("s", problem_spec["s"], True)
+        self.add_experiment_variable("dim", problem_spec["dim"], True)
+        self.add_experiment_variable("cfl", problem_spec["cfl"], True)
+        self.add_experiment_variable("cgm", problem_spec["cgm"], True)
+        self.add_experiment_variable("cgt", problem_spec["cgt"], True)
+
```
pearce8, line 167: Per above, we don't need cfl, cgm, cgt

experiments/laghos/experiment.py
```diff
@@ -120,6 +174,13 @@ def generate_perf_specs(self):
         self.add_experiment_variable("ms", problem_spec["ms"], True)
```
pearce8: what is ms?

repos/ramble_applications/laghos/application.py
pearce8: There are at the very least the following parameters that can be varied independently of workloads:
- p
- dim
- nx,  ny, nz
- rs, rp (these are tied to the number of ranks though)

And of course turning vis on or off.

I agree with @michaelmckinsey1 that we should make this code less redundant and therefore error prone.

repos/ramble_applications/laghos/application.py
```diff
@@ -42,11 +169,24 @@ class Laghos(ExecutableApplication):
...
+                       ' -cgt {cgt}' +
+                       ' {vis}' +
+                       ' {k}' +
                        ' {assembly}',
                        use_mpi=True)
 
-    workload('triplept', executables=['triplept'])
+    workload('taylor-green', executables=['taylor-green'])
     workload('sedov', executables=['sedov'])
+    workload('1D-sod-shock', executables=['1D-sod-shock'])
```
pearce8: wait, is this one only valid for 1D?  Does it use a file as input?

If that is the case, that the workloads without a 1D or 2D in the name have the 1D, 2D, and 3D variants - while the others do not allow the variants.  We need a clean way to specify that.

experiments/laghos/experiment.py
amroakmal: Done, and now we can persist visualization result

experiments/laghos/experiment.py
amroakmal: Dr. @pearce8 : max steps: "Maximum number of steps (negative means no restriction)."

pearce8: Thanks, lets make this one a variant.

repos/ramble_applications/laghos/application.py
amroakmal: I dont think so, I mean I'm able to run 2D-reimann-12 problem in 3D, i.e. the problems can run with different values for -dim {1, 2, 3} and I'll experiment with others.

Also, we should add the "s" parameter (controls which ODE solver to be used) as variant, correct?

experiments/laghos/experiment.py
```diff
@@ -55,16 +64,62 @@ class Laghos(
...
+        default="2",
+        description="Elements in y-dimension (do not specify mesh_file). Note: this is mutually-exclusive with -epm. Use -epm 0 to use -nx, -ny, and -nz.",
+    )
+
+    variant(
+        "nz",
+        default="2",
+        description="Elements in z-dimension (do not specify mesh_file). Note: this is mutually-exclusive with -epm. Use -epm 0 to use -nx, -ny, and -nz.",
+    )
+
```
amroakmal: Remove

pearce8:
@michaelmckinsey1 please review and let me know when this is ready for my review.

experiments/laghos/experiment.py
```diff
@@ -55,16 +64,62 @@ class Laghos(
...
+        "ny",
+        default="2",
+        description="Elements in y-dimension (do not specify mesh_file). Note: this is mutually-exclusive with -epm. Use -epm 0 to use -nx, -ny, and -nz.",
+    )
+
+    variant(
+        "nz",
+        default="2",
+        description="Elements in z-dimension (do not specify mesh_file). Note: this is mutually-exclusive with -epm. Use -epm 0 to use -nx, -ny, and -nz.",
+    )
```
michaelmckinsey1: remove these

experiments/laghos/experiment.py
```diff
@@ -55,16 +64,62 @@ class Laghos(
...
+        "ms",
+        default="-1",
+        description="Maximum number of steps (negative means no restriction).",
+    )
+
+    variant(
+        "tf",
+        default="0.6",
+        description="Final time; start time is 0.",
+    )
```
michaelmckinsey1, line 98: If these are going to be variants, I think we should keep the default values that we had 
```
            "ms": 250,
            "tf": 10000,
```
not the laghos defaults. There was likely a reason they are already set this way

experiments/laghos/experiment.py
```diff
@@ -230,6 +278,22 @@ def compute_applications_section(self):
...
+            )
+        else:
+            self.add_experiment_variable("vis", "-no-vis", True)
+            self.add_experiment_variable("vs", "-no-visit", True)
+            self.add_experiment_variable("k", "", False)
+
+        params_to_check = ["nx", "ny", "nz", "dim", "ms", "tf", "s"]
+        for param in params_to_check:
+            if param in self.spec.variants:
+                self.add_experiment_variable(param, self.spec.variants[param][0], True)
```
michaelmckinsey1: Once the above changes are made, this should just be 
```
self.add_experiment_variable("dim", self.spec.variants["dim"][0], True)
...
```
not a loop

michaelmckinsey1 (changes requested):
`application.py` looks fine now that we have `common_args`, much more succinct.

pearce8:
@michaelmckinsey1 please re-review, and remove the "changes requested" label when you resolve all requests and approve the PR.

michaelmckinsey1 (approved):
Looks good, given the Laghos gitlab tests that are running are successful

pearce8 (approved):

PR #1281, opened 2026-03-09
Remove `rocm/7.1.0` on `llnl-elcapitan`

michaelmckinsey1:
## Description

- [x] Recommended to use `rocm/7.2.0` over `rocm/7.1.0` and `rocm/7.1.0` will be deprecated https://lc.llnl.gov/confluence/spaces/ELCAP/pages/899809690/2026-02-26+User+Meeting

### DIFF
--- a/systems/llnl-elcapitan/system.py
+++ b/systems/llnl-elcapitan/system.py
@@ -140,7 +140,6 @@ class LlnlElcapitan(System):
             "6.4.1",
             "6.4.2",
             "6.4.3",
-            "7.1.0",
             "7.2.0",
         ),
         description="ROCm version",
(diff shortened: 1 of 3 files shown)


systems/llnl-elcapitan/system.py
```diff
@@ -223,7 +222,7 @@ def __init__(self, spec):
                 f"{self.gcc_version.major}.{self.gcc_version.minor}"
             )
         else:
-            if self.rocm_version >= Version("7.1.0"):
+            if self.rocm_version >= Version("7.0.0"):
```
pearce8, line 225: Shouldn't this be 7.2.0 @michaelmckinsey1 ?

pearce8 (changes requested):

systems/llnl-elcapitan/system.py
michaelmckinsey1: The logic for that check applies to all `7.0` versions, not just `7.2` and beyond. Since we are only using `7.2` it would have the same effect, but I think that dilutes the point of the check.

pearce8 (approved):

PR #1294, opened 2026-03-25
Add riken-cloud system support

yoshifuminakamura:
## Description

Add support for the RIKEN cloud system.

This PR introduces a new system definition in `systems/riken-cloud/system.py` and includes the minimal supporting changes needed to use the system in Benchpark:
- update `modifiers/allocation/modifier.py` to handle the `--gpus 0` Slurm case used by this system
- adjust `repos/spack_repo/benchpark/packages/qws/package.py` for compatibility with the RIKEN cloud environment

This change is being submitted as a focused PR split out from a larger fork-local branch so the system support can be reviewed independently.

- [ ] Add/modify `systems/all_hardware_descriptions/hardware_name/hardware_description.yaml` which will appear in the [docs catalogue](https://software.llnl.gov/benchpark/system-list.html)

### DIFF
--- a/systems/riken-cloud/system.py
+++ b/systems/riken-cloud/system.py
@@ -1,0 +1,1816 @@
+# Copyright 2023 Lawrence Livermore National Security, LLC and other
+# Benchpark Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+from packaging.version import Version
+
+from benchpark.cudasystem import CudaSystem
+from benchpark.directives import maintainers, variant
+from benchpark.openmpsystem import OpenMPCPUOnlySystem
+from benchpark.paths import hardware_descriptions
+from benchpark.system import (
+    System,
+    compiler_def,
+    compiler_section_for,
+    merge_dicts,
+)
+
+
+class RikenCloud(System):
+
+    maintainers("jdomke", "SBA0486")
+
+    id_to_resources = {
+        "fx700": {
+            "sys_cores_per_node": 48,
+            "sys_mem_per_node_GB": 32,
+            "system_site": "rccs",
+            "hardware_key": str(hardware_descriptions)
+            + "/Fujitsu-A64FX-TofuD/hardware_description.yaml",
+        },
+        "gh200": {
+            "sys_cores_per_node": 72,
+            "sys_gpus_per_node": 1,
+            "sys_mem_per_node_GB": 512,
+            "system_site": "rccs",
+            "queue": "qc-gh200",
+            "hardware_key": str(hardware_descriptions)
+            + "/NVIDIA-neoverse-GH200-Infiniband/hardware_description.yaml",
+        },
+        "dgx": {
+            "sys_cores_per_node": 20,
+            "sys_gpus_per_node": 1,
+            "sys_mem_per_node_GB": 128,
+            "system_site": "rccs",
+            "queue": "ng-dgx-3h",
+            "hardware_key": str(hardware_descriptions)
+            + "/NVIDIA-cortex-GB10-Ethernet/hardware_description.yaml",
+        },
+        "genoa": {
+            "sys_cores_per_node": 96,
+            "sys_mem_per_node_GB": 768,
+            "system_site": "rccs",
+            "queue": "genoa",
+            "hardware_key": str(hardware_descriptions)
+            + "/AMD-zen4-EPYC-Ethernet/hardware_description.yaml",
+        },
+    }
+
+    variant(
+        "cluster",
+        default="fx700",
+        values=("fx700", "gh200", "genoa", "dgx"),
+        description="Which cluster to run on",
+    )
+    variant(
+        "compiler",
+        default="fj",
+        values=("clang", "gcc", "fj", "nvhpc", "cuda"),
+        description="Which compiler to use",
+    )
+    variant(
+        "cuda",
+        default="12.9",
+        values=("12.3", "12.4", "12.5", "12.6", "12.8", "12.9", "13.0", "13.1", "13.2"),
+        description="CUDA version",
+    )
+    variant(
+        "nvhpc",
+        default="25.7",
+        values=("26.3", "25.9", "25.7", "24.9", "24.3"),
+        description="NVHPC version",
+    )
+    variant(
+        "gtl",
+        default=False,
+        values=(True, False),
+        description="Use GTL-enabled MPI",
+    )
+
+    def __init__(self, spec):
+        super().__init__(spec)
+        if self.spec.variants["cluster"][0] == "fx700":
+            self.programming_models = [OpenMPCPUOnlySystem()]
+            self.scheduler = "slurm"
+
+        if self.spec.variants["cluster"][0] == "genoa":
+            self.programming_models = [OpenMPCPUOnlySystem()]
+            self.scheduler = "slurm"
+
+        if self.spec.variants["cluster"][0] == "gh200":
+            self.programming_models = [CudaSystem(), OpenMPCPUOnlySystem()]
+            self.cuda_version = Version(self.spec.variants["cuda"][0])
+            self.gtl_flag = self.spec.variants["gtl"][0]
+            self.nvhpc_version = Version(self.spec.variants["nvhpc"][0])
+            if str(self.nvhpc_version) == "26.3":
+                self.cuda_version = "13.1"
+            if str(self.nvhpc_version) == "25.9":
+                self.cuda_version = "13.0"
+            if str(self.nvhpc_version) == "25.7":
+                self.cuda_version = "12.9"
+            if str(self.nvhpc_version) == "24.9":
+                self.cuda_version = "12.6"
+            if str(self.nvhpc_version) == "24.3":
+                self.cuda_version = "12.3"
+            self.scheduler = "slurm"
+
+        if self.spec.variants["cluster"][0] == "dgx":
+            self.programming_models = [CudaSystem(), OpenMPCPUOnlySystem()]
+            self.cuda_version = Version(self.spec.variants["cuda"][0])
+            self.gtl_flag = self.spec.variants["gtl"][0]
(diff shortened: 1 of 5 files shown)


modifiers/allocation/modifier.py
```diff
@@ -346,6 +346,12 @@ def slurm_instructions(self, v):
 
         sbatch_opts.append("--exclusive")
```
pearce8, line 348: @scheibelp @michaelmckinsey1 Please check the changes to the allocation modifier.

michaelmckinsey1: Ok if I'm understanding correctly: `--gpus 0` is valid as a slurm directive on `riken-cloud` for the `gh200` cluster. @yoshifuminakamura Is the intention here to remove any options that begin with `--gpu` aside from `--gpus 0`? If so, I would recommend checking for other valid syntax:
```
if any(opt in ('--gpus 0', '--gpus=0') for opt in srun_opts):
    srun_opts = [
        opt for opt in srun_opts
        if not opt.startswith('--gpu') or opt in ('--gpus 0', '--gpus=0')
    ]
```

@yoshifuminakamura Could you elaborate more on what is expected to happen when setting `--gpus 0` on `riken-cloud` `gh200` cluster (i.e. is it a special gpu mode)? Right now this change would affect any of our other slurm systems.

yoshifuminakamura: Thanks for the question.

On the riken-cloud gh200 queue, there is no GRES configuration defined in Slurm, so we need to explicitly set --gpus 0 as a workaround.

After syncing with upstream, the changes in riken-cloud/system.py and modifier.py at the time of the PR no longer handle this correctly. As a result, we are currently addressing it as shown here:
https://github.com/RIKEN-RCCS/benchpark/blob/96bed8662f317a1180813c313481dfd6d213e721/modifiers/allocation/modifier.py#L349

Ideally, it would be preferable to handle this in system.py without modifying modifier.py.

Do you think it would make sense to withdraw this PR for now?

systems/riken-cloud/system.py
```diff
@@ -0,0 +1,1354 @@
...
+
+        return cfg
+
+    def system_specific_variables(self):
+        if self.spec.variants["cluster"][0] == "gh200":
+            return {
+                "cuda_arch": "90",
+                "queue": "qc-gh200",
+                "pre_exec_cmds": "export SLURM_MPI_TYPE=pmix",
+                "extra_cmd_opts": "--gpus 0",
```
michaelmckinsey1: This key should be `extra_batch_opts`
```suggestion
                "extra_batch_opts": "--gpus 0",
```

modifiers/allocation/modifier.py
```diff
@@ -346,6 +346,12 @@ def slurm_instructions(self, v):
...
 
+        if '--gpus 0' in srun_opts:
+            srun_opts = [
+                opt for opt in srun_opts
+                if not (opt.startswith('--gpu') and opt != '--gpus 0')
+            ]
```
michaelmckinsey1: Can you remove this piece, and instead add this function (outside of `def slurm_instructions`):
```suggestion
    def riken_slurm_instructions(self, v):
    
        self.slurm_instructions(v)
        
        if '--gpus 0' in v.allocation_directives:
            if v.n_gpus:
                v.mpi_command.remove(f"--gpus {v.n_gpus}")
                v.allocation_directives.remove(f"--gpus {v.n_gpus}")
```

and then in this same file, add "riken-slurm" to `determine_scheduler_instructions` where
```
"riken-slurm": self.riken_slurm_instructions
```
This will make it such that the scheduler behavior for this system does not impact other slurm systems.

systems/riken-cloud/system.py
```diff
@@ -0,0 +1,1354 @@
...
+            self.programming_models = [OpenMPCPUOnlySystem()]
+            self.scheduler = "slurm"
+
+        if self.spec.variants["cluster"][0] == "gh200":
+            self.programming_models = [CudaSystem(), OpenMPCPUOnlySystem()]
+            self.cuda_version = Version(self.spec.variants["cuda"][0])
+            self.gtl_flag = self.spec.variants["gtl"][0]
+            self.nvhpc_version = Version(self.spec.variants["nvhpc"][0])
+
+            self.scheduler = "slurm"
```
michaelmckinsey1, line 126:
```suggestion
            self.scheduler = "riken-slurm"
```

modifiers/allocation/modifier.py
michaelmckinsey1: I think the PR is ok to proceed. @scheibelp and I have some small suggestions in the PR to get this to work with our other slurm systems. Basically we think the `gh200` system, based on this unique requirement, should use a different scheduler `riken-slurm` that is identical to `slurm` aside from the `--gpus 0` specific logic. Are the proposed changes amenable to you?

systems/riken-cloud/system.py
pearce8: @yoshifuminakamura have you had a chance to try the suggestion?  We would like to merge this PR if you have it working.

yoshifuminakamura: @pearce8 @michaelmckinsey1 @scheibelp 
Thanks for the follow-up, and sorry for the delay.

I have not fully tried the `riken-slurm` suggestion yet. I am checking whether we can fix the underlying issue on the RIKEN Cloud Slurm side by configuring GRES for the GPU nodes. If that works, we should be able to keep using Benchpark's standard `slurm` scheduler and avoid adding RIKEN-specific scheduler logic.

To avoid blocking this PR longer, I can compact the PR by Friday, June 12, 2026 JST: remove the unrelated `qws/package.py` and common Slurm modifier changes, keep the PR focused on `systems/riken-cloud/system.py`, and leave the `--gpus 0` workaround out of the active path for now while the GRES configuration is being checked.

One more question before I push that update: what is the preferred approach for the hardware description yaml files? The current fork-local files use local queue/cluster-style names, so I would like to either reuse existing Benchpark hardware descriptions if appropriate, or rename/update the new files to match the upstream catalogue style.

The two GPU systems I need to describe are:

- NVIDIA GH200 Grace Hopper Superchip, 512 GB, InfiniBand HDR 200 Gbps
- NVIDIA GB10 Grace Blackwell Superchip, 128 GB, Ethernet 10 Gbps

If there are existing hardware descriptions that should be reused for either of these, please let me know. Otherwise, I can add new hardware descriptions using more hardware-oriented names and corrected contents.

Does that sound reasonable?

michaelmckinsey1: @yoshifuminakamura Yes this approach sounds reasonable for us.

https://software.llnl.gov/benchpark/add-a-system-config.html#a-adding-system-hardware-specs provides the template for naming a hardware spec.

~We don't have any GH200 or Blackwell examples, I think the closest example would be our H100 sapphirerapids definition https://github.com/llnl/benchpark/blob/develop/systems/all_hardware_descriptions/DELL-sapphirerapids-H100-Infiniband/hardware_description.yaml~
Correction: `lanl-venado` has `GH200`, so that would be an example. https://github.com/llnl/benchpark/tree/develop/systems/all_hardware_descriptions/HPECray-neoverse-GH200-Slingshot

yoshifuminakamura: @michaelmckinsey1  Thank you for the guidance.

I updated this PR to follow the suggested approach. The R-CCS Cloud Slurm configuration has now been updated so that the GH200 nodes expose the GPU through GRES correctly, so the previous `--gpus 0` workaround is no longer needed. I removed that workaround from this PR and did not add any changes to the common Slurm modifier or a custom `riken-slurm` scheduler.

I also compacted the PR so that it now only adds the RIKEN Cloud system definition and the associated hardware descriptions. The hardware descriptions were renamed following the Benchpark hardware-spec naming template, and I updated the network/spec details based on the current systems:

- `NVIDIA-neoverse-GH200-Infiniband`: NVIDIA GH200 Grace Hopper Superchip, 512GB, InfiniBand HDR 200Gbps
- `NVIDIA-cortex-GB10-Ethernet`: NVIDIA GB10 Grace Blackwell Superchip, 128GB, Ethernet 10Gbps
- `AMD-zen4-EPYC-Ethernet`: AMD EPYC Zen4/Genoa system

I verified the updated branch with:

- `git diff --check upstream/develop...HEAD`
- `python3 -m py_compile systems/riken-cloud/system.py`

Please let me know if you would prefer different names for the hardware descriptions, but I think this should now be much easier to review and merge.

michaelmckinsey1 (changes requested):
@yoshifuminakamura It looks better, thanks. A couple of things to get the unit tests passing:
1. could you add `"stream riken-cloud cluster=dgx"` to the `SKIP_EXPR` list in https://github.com/RIKEN-RCCS/benchpark/blob/2b1d10c6c14294681811b9d7aa8044e205d4292a/.github/utils/dryruns.py#L15? The default stream config tries to use 32 cores but dgx only has 20, so we can skip it in our tests.
2. can you `pip install black` and run `black systems/riken-cloud/system.py` to format the code for your system? It is currently not formatted.

@pearce8  I checked the rest of the dryruns are passing (they look like failures because the code coverage step fails on forks of benchpark). So by fixing the above things there should be no issues with the tests.

yoshifuminakamura:
@michaelmckinsey1 
Thanks for checking the dryruns. I addressed both requested items:

- Added `stream riken-cloud cluster=dgx` to `SKIP_EXPR`
- Ran `black systems/riken-cloud/system.py`

I also verified the updated branch with `git diff --check`, `python3 -m py_compile`, and `black --check`.

michaelmckinsey1:
@yoshifuminakamura It seems the tests won't run because there is a merge conflict. I'm guessing with the one line change in the dryruns file, so should be easy to fix.

Can you merge develop into your branch?

yoshifuminakamura:
@michaelmckinsey1 Thanks for the heads up. I rebased this PR branch on the latest `upstream/develop` and resolved the conflict in `.github/utils/dryruns.py`. I also compacted the branch into a single commit on top of the current develop branch.

PR #1307, opened 2026-04-06
avoid inserting -lmpi_gtl_hsa for caliper build

scheibelp:
@rfhaque noticed the benchpark/spack build of amg2023 with Caliper ROCm profiling on was hanging

I was able to reproduce that, and it seems that building Caliper with `-lmpi_gtl_hsa` is causing this hang. I don't know enough about Caliper to know why that would be a problem.

### DIFF
--- a/repos/spack_repo/benchpark/packages/caliper/package.py
+++ b/repos/spack_repo/benchpark/packages/caliper/package.py
@@ -14,5 +14,10 @@ class Caliper(BuiltinCaliper):
     directly into applications and activate them at runtime.
     """
 
+    def setup_build_environment(self, env):
+        # Do not insert -lmpi_gtl_hsa, this seems to cause a hang when doing
+        # rocm profiling in caliper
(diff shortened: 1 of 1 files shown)


pearce8:
@daboehme 

- [X] @rfhaque has verified that removing `-lmpi_gtl_hsa` resolves the hang
- [ ] Why does linking Caliper with `-lmpi_gtl_hsa` cause a hang?
- [ ] Does Caliper ever need to link with `-lmpi_gtl_hsa`?

pearce8:
@nhanford Are we supposed to use `-lmpi_gtl_hsa` on MI300A?  Any downsides to not using it?

nhanford:
@scheibelp Please prepend [`on_hang_stat_and_kill`](https://hpc.llnl.gov/software/development-environment-software/stat-stack-trace-analysis-tool) to your hanging `flux run` and report the anonymized results if possible.

pearce8 (approved):

PR #1317, opened 2026-04-20
Update .gitignore

michaelmckinsey1:
## Description

- [x] When `benchpark configure` is run, this file is updated and registers in git as a modified file.
   - [x] Modified files report erroneously as revisions on the dashboard https://my.cdash.org/index.php?project=Benchpark 

So this PR will ignore local changes to this file, but we should keep it in the repo

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ configure.yaml
 .spack-env
 .env
 .DS_Store
+config/bootstrap.yaml


michaelmckinsey1:
@scheibelp Is the pattern of ignoring local changes, but keeping the file in the repo make sense to you for this file?

scheibelp:
If it works for the specific purpose you described, it's ok with me

I thought generally that commands like `git commit -a` would include changes to ignored files if they are tracked. So maybe something like the following would be more-generally useful:

* benchpark copies that config into an untracked+ignored directory (and it uses that) 
* or maybe `benchpark configure` shouldn't alter the file at all if it detects that it already has default settings

michaelmckinsey1:
@scheibelp 

>  or maybe `benchpark configure` shouldn't alter the file at all if it detects that it already has default settings

I believe Benchpark does not alter the file if the defaults are there. This issue is occurring from trying to set different settings than default, specifically on the CI.

> benchpark copies that config into an untracked+ignored directory (and it uses that)

Are you recommending this be implemented in this PR? I think this PR would work as a starting point and if we need to make changes to this file we can force add it.

scheibelp:
Apologies but I'm not sure how this works. For example if I apply this change, then manually edit `config/bootstrap.yaml`, then it still appears as a changed file in `git status` and `git diff`, but is cdash more-aggressive in filtering out files in `.gitignore`?

michaelmckinsey1:
> Apologies but I'm not sure how this works. For example if I apply this change, then manually edit `config/bootstrap.yaml`, then it still appears as a changed file in `git status` and `git diff`, but is cdash more-aggressive in filtering out files in `.gitignore`?

@scheibelp 
The issue is that for CI we update `config/bootstrap.yaml` (because we change the bootstrap location), but then the test essentially registers wrong (it appears essentially as a failure) because there were unstaged changes on top of the commit (it tracks the commit at which the test was ran. So by omitting `config/bootstrap.yaml` from the git, this issue wouldn't happen anymore.

scheibelp (approved):

PR #1323, opened 2026-04-25
`benchpark setup`: use preexisting spack

scheibelp:
Add a `--spack` option to `benchpark setup` 

Like 

```
./bin/benchpark system init --dest=tioga llnl-elcapitan cluster=tioga compiler=rocmcc
./bin/benchpark experiment init --dest=kripke tioga kripke +rocm
./bin/benchpark setup --spack=`pwd`/spack tioga/kripke wkp/
^
(this last command is what changed)
```

when setting up a second experiment, the user would not specify `--spack`:

```
./bin/benchpark experiment init --dest=amg2023 tioga amg2023 +rocm
./bin/benchpark setup tioga/amg2023 wkp/
(no --spack on this)
```

In this case Benchpark does no configuration of the Spack instance (so for example has no awareness of the benchpark repository). It creates a symlink to the specified Spack instance. The idea is that someone can use any checkout they want of Spack (e.g. a personal branch) and configure it however they like. 

If this were used to share installs with several workspaces, or access a binary cache, the experiment builds would not be repeatable unless this Spack instance configured `concretizer:reuse:false`.

TODOs

- [x] This is probably not working yet, also consider semantics
- [x] Docs for the feature


scheibelp:
@michaelmckinsey1 I've confirmed that the spack instance designated in `benchpark setup --spack=...` does not get updated based on `checkout-versions.yaml`. In low-level terms: this is because `RuntimeResources.bootstrap()` is only called for Ramble/Spack/spack-packages inside of `~/.benchpark` (i.e. `main.py` is the only caller of that function).

(that also means for example that a user can modify the Spack instance in `experiments_root` after it is generated, and Benchpark would not undo those changes)

lib/benchpark/cmd/setup.py
```diff
@@ -207,7 +210,23 @@ def include_fn(fname):
...
+        # if the user says "use my specific spack instance" then I want to
+        # minimize changes to it. Note that Ramble may perform config commands
+        spack_user_cache_path = experiments_root / "spack-cache"
+        spack_location = str(experiments_root / "spack")
+        os.symlink(args.spack, spack_location)
+        pkg_str = f"""\
+export SPACK_USER_CACHE_PATH={spack_user_cache_path}
+export SPACK_DISABLE_LOCAL_CONFIG=1
+. {spack_location}/share/spack/setup-env.sh
+"""
```
michaelmckinsey1, line 281: small nit. The other branch of this if statement basically does the same logic, can we refactor these lines to be after the if-elif?

michaelmckinsey1 (changes requested):
Are you still looking to complete these todo's from the description?

- [ ] If we use a newer Ramble, then benchpark can add ~spack_install_compilers to the ramble.yaml package manager config, and then I think we can avoid any modifications of the shared spack instance
- [x] Docs for the feature

pearce8 (approved):

PR #1324, opened 2026-04-26
Benchpark gets its first Fluxtainer

nhanford:
## Fluxtainer System

This `system.py` describes the el10 container based on what is distributed by the Flux Framework project. The external GCC 14 and MPICH are included in the description.

## OSU Microbenchmarks Experiment Modification

An issue was uncovered in the `experiment.py` for `osu-micro-benchmarks` where only 1 rank was being specified across 2 nodes. I have changed this to 2 ranks across 2 nodes for now just to have a minimum viable product. I will deal with scaling and non-scaling benchmarks in a future PR.

## `checkout-versions.yaml`

I ran into an old bug with Ramble reporting results that appears to have been fixed recently, so I have bumped this version to the one that works.

## New Dockerfile in a new `containerfiles` directory

I created a directory structure for containerfiles in the root of the project. I tried to make this as extensible as possible. As I see it, the 2 main invariant factors requiring separate containers are the resource manager and the OS. Compilers, MPIs, and other system software should be able to coexist.

The Dockerfile itself simply clones Benchpark, installs the dependencies via PIP, and then runs a simple command to induce bootstrapping. Perhaps I should have used `benchpark bootstrap`?

## New Workflow to build and push containers to the GHCR for this project

I set this up to build a new container each week or when changes are made to the containerfile itself, or to be triggered manually. We probably don't need this rebuilding all the time, but it could be a good way to test basic functionality.

Please let me know what you think.

Thanks,
Nate

### DIFF
--- a/systems/fluxtainer/system.py
+++ b/systems/fluxtainer/system.py
@@ -15,12 +15,8 @@ class Fluxtainer(System):
     maintainers("nhanford")
 
     id_to_resources = {
-        "arm": {
-            "cpu_arch": "arm64"
-        },
-        "x86": {
-            "cpu_arch": "x86_64_v3"
-        },
+        "arm": {"cpu_arch": "arm64"},
+        "x86": {"cpu_arch": "x86_64_v3"},
     }
 
     variant(
--- a/systems/fluxtainer/system.py
+++ b/systems/fluxtainer/system.py
@@ -17,13 +17,13 @@ class Fluxtainer(System):
     id_to_resources = {
         "arm": {
             "cpu_arch": "arm64",
-            "sys_cores_per_node": 32,
+            "sys_cores_per_node": 8,
             "sys_mem_per_node_GB": 1,
             "n_nodes": 4,
         },
         "x86": {
             "cpu_arch": "x86_64_v3",
-            "sys_cores_per_node": 32,
+            "sys_cores_per_node": 8,
             "sys_mem_per_node_GB": 1,
             "n_nodes": 4,
         },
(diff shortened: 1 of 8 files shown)


systems/fluxtainer/system.py
```diff
@@ -0,0 +1,179 @@
...
+        default="x86",
+        description="Target Architecture",
+    )
+
+    def __init__(self, spec):
+        super().__init__(spec)
+        self.programming_models = [OpenMPCPUOnlySystem()]
+
+        self.scheduler = "flux"
+        setattr(self, "sys_cores_per_node", 32)
```
pearce8: This seems repetitive

systems/fluxtainer/system.py
```diff
@@ -0,0 +1,179 @@
...
+        setattr(self, "sys_mem_per_node_GB", 2)
```
pearce8: Hmmm this overwrites the id_to_resources struct

pearce8 (changes requested):
We need docs!

systems/fluxtainer/system.py
michaelmckinsey1: These are actually overwritten _by_ id_to_resources. @nhanford_ I think you intend to delete these two lines and then decide between `sys_mem_per_node_GB` 1 or 2?

nhanford: I decided to put the common elements in the `setattr` and use the `id_to_resources` struct only where necessary.

PR #1326, opened 2026-04-30
Copy repos.yaml into user-config

michaelmckinsey1:
## Description

- [x] Bug introduced in #1322, where `user-config` would only contain `bootstrap.yaml` but not the other required file, `repos.yaml`


michaelmckinsey1:
https://lc.llnl.gov/gitlab/benchpark/benchpark/-/pipelines/699098

scheibelp (approved):

PR #1327, opened 2026-05-04
Fail in CI if experiment was skipped

michaelmckinsey1:
## Description

- [x] Currently, the CI will not fail if more resources were requested than available than one node (will happen with current `osu-micro-benchmarks`). Ramble will just skip the experiment, and pass because no experiments failed. This PR checks if the output files from running the experiments are empty and will fail if they are empty, which indicates failure.
- [x] This will not work if the test needs to fail and 1+ experiments out of X executed.


michaelmckinsey1:
https://lc.llnl.gov/gitlab/benchpark/benchpark/-/jobs/3614560
this test should fail now because it will execute 0 out of 1 experiments

michaelmckinsey1:
@nhanford See https://lc.llnl.gov/gitlab/benchpark/benchpark/-/jobs/3614560#L233 where test now detects false positive success because the experiment never ran, and correctly fails.

scheibelp (approved):

PR #1330, opened 2026-05-05
Fix gcc on `llnl-elcapitan`

michaelmckinsey1:
## Description

- [x] User reported that gcc compiler option was not working. Fix using GCC and add test


michaelmckinsey1:
https://lc.llnl.gov/gitlab/benchpark/benchpark/-/pipelines/700739

pearce8 (approved):

PR #1337, opened 2026-05-06
Set pool size 0 for cpu as umpire not used

michaelmckinsey1:
## Description

- [x] set argument for clarity


rfhaque (commented):
@michaelmckinsey1 pool size is ignored if umpire is not used

michaelmckinsey1:
> @michaelmckinsey1 pool size is ignored if umpire is not used

That's why I think we should either set it to some value like 0 or not set the argument at all. Setting the pool size to a value for cpu runs is potentially confusing for users.

pearce8 (approved):

PR #1342, opened 2026-05-08
Fix Gitlab Failures

michaelmckinsey1:
## Description

- `exec_mode=test` resource requests should satisfy CI single node requirements. Other exec modes can implement different resource requests.
   - [x] Fix osu failure which skips experiment because it requests 2 nodes by default, where CI only has 1 node.
   - [x] AMG resource request 8 ranks, only 4 on tuolumne

### DIFF
--- a/experiments/amg2023/experiment.py
+++ b/experiments/amg2023/experiment.py
@@ -505,8 +505,8 @@ class Amg2023(
         if self.spec.satisfies("exec_mode=perf"):
             self.generate_perf_specs()
         else:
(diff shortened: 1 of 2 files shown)


michaelmckinsey1:
https://lc.llnl.gov/gitlab/benchpark/benchpark/-/pipelines/702634

michaelmckinsey1:
Relevant tests are fixed in gitlab

pearce8 (approved):

PR #1349, opened 2026-05-12
py-scaffold rocm/7 fixes

michaelmckinsey1:
## Description

- [x] Change strong and weak scaling definitions
- [x] Consolidate config url
- [x] Add `roctx` service as option for profiling
- [x] fix run cmd
- [x] Add `rccl` profiling option for caliper


lib/benchpark/cmd/analyze.py
```diff
@@ -595,6 +614,13 @@ def prepare_data(**kwargs):
         logger.info(f"Normalizing '{kwargs['yaxis_metric']}' by '{norm_col}'")
         tk.dataframe[kwargs["yaxis_metric"]] /= tk.dataframe[norm_col]
 
+    print(tk.profile_mapping)
+    print(tk.tree(kwargs["yaxis_metric"],indices=14162220873700))
+    print(tk.tree(kwargs["yaxis_metric"],indices=6334389100413))
+    print(tk.tree(kwargs["yaxis_metric"],indices=3650960442712))
+    print(tk.tree(kwargs["yaxis_metric"],indices=13284072939493))
+    print(tk.tree(kwargs["yaxis_metric"],indices=15700773896140))
```
michaelmckinsey1: Need to undo this, but would be nice as a feature to visualize the tree _after_ the query instead of only before

lib/benchpark/cmd/analyze.py
```diff
@@ -99,10 +100,28 @@ def set_metrics(self):
...
+        self.tk = tk
+        # Matches application_name column in metadata
+        self.name = "scaffold"
+
+    def set_metrics(self):
+        self.tk.dataframe["TFLOPs (BF16)"] = (
+            # https://github.com/ROCm/rocm-systems/blob/8bb3b73c117e5630106540447268ccad771906a4/projects/rocprofiler-compute/src/rocprof_compute_soc/analysis_configs/gfx90a/0400_roofline.yaml#L51
+            self.tk.dataframe["Total SQ_INSTS_VALU_MFMA_MOPS_BF16 (exc)"] * 512
+            / self.tk.dataframe["Avg GPU time (E)"]
+            / 10**12
```
michaelmckinsey1: This definition is technically agnostic to ScaFFold and could be applied to any application with these two columns

michaelmckinsey1: ~also change avg to max~

repos/spack_repo/benchpark/packages/py_scaffold/package.py
```diff
@@ -68,10 +68,10 @@ def setup_run_environment(self, env):
...
                     env.prepend_path("LD_LIBRARY_PATH", self.spec['mpi'].extra_attributes["gtl_lib_path"])
 
-        # if self.spec.satisfies("+caliper"):
-        #     if self.spec.satisfies("+rocm"):
-        #         # Need to set this to libcaliper.so to avoid rocprofiler context error
-        #         env.set("ROCP_TOOL_LIBRARIES", os.path.join(self.spec["caliper"].prefix, "lib64", "libcaliper.so"))
+        if self.spec.satisfies("+caliper"):
+            if self.spec.satisfies("+rocm"):
+                # Need to set this to libcaliper.so to avoid rocprofiler context error
+                env.set("ROCP_TOOL_LIBRARIES", os.path.join(self.spec["caliper"].prefix, "lib64", "libcaliper.so"))
```
pearce8, line 74: ROCP_TOOL_LIBRARIES -> ROCM_TOOL_LIBRARIES?

lib/benchpark/cmd/analyze.py
```diff
@@ -99,10 +100,35 @@ def set_metrics(self):
             * self.tk.dataframe["Reps"]
             * self.tk.metadata["mpi.world.size"]
         )
+        self.tk.exc_metrics.append("FLOP Rate (GFLOPS)")
 
         return ["Memory Bandwidth (GB/s)", "FLOP Rate (GFLOPS)"]
 
 
+class ScaFFold:
```
pearce8: Why are we setting metrics here?  Should we have a category for applications that use BF16?

repos/spack_repo/benchpark/packages/py_scaffold/package.py
michaelmckinsey1: No it is ROCP for rocprofiler

lib/benchpark/cmd/analyze.py
michaelmckinsey1: Yes that is what I was thinking here https://github.com/llnl/benchpark/pull/1349#discussion_r3263050750. Currently this applies to no other benchmark in benchpark (BF16), but I can try to make in generalized still.

michaelmckinsey1: @pearce8 Ok I have now made it such that any application with the required columns will compute the derived metric, not just ScaFFold.

pearce8 (approved):

PR #1356, opened 2026-05-15
Laghos changes

rfhaque:
## Description
This PR
- [x] Adds a variant `mesh-strategy` to laghos `experiment.py`, with possible values `epm`, `refinement` and `meshfile`
- [x] Setting `mesh-strategy` to `epm` allows users to specify the number of zones per rank which is used by mfem to construct the global mesh and partition it across the available ranks. The required input parameter `epm` has been added to `application.py`
- [x] Setting `mesh-strategy` to `refinement` uses the existing mechanism of mesh construction using serial and parallel refinement via the input parameters `nx, ny, nz, rs, rp`
- [x] `mesh-strategy` set to `meshfile` will use a user-specified mesh file for the problem (This PR does not implement this option in `experiment.py`)
- [x] Sets the benchpark variables `process_problem_size` and `total_problem_size` in terms of `qpts` (quadrature points) instead of `zones` (number of mesh elements). 
`qpts = zones * quad`, where `quad` is the order-dependent mesh quadrature (8 for linear, 64 for quadratic and 216 for cubic order)


experiments/laghos/experiment.py
```diff
@@ -108,141 +108,105 @@ class Laghos(
...
+                problem_spec["epm"] = 19652
         else:
-            problem_spec["rs"] = 4
-            problem_spec["rp"] = 1
-
-        self.add_experiment_variable("nx", problem_spec["nx"], True)
-        self.add_experiment_variable("ny", problem_spec["ny"], True)
-        self.add_experiment_variable("nz", problem_spec["nz"], True)
-        self.add_experiment_variable("rs", problem_spec["rs"], True)
-        self.add_experiment_variable("rp", problem_spec["rp"], True)
```
michaelmckinsey1, line 166: I get removing nx, ny, nz because they are exlusive with epm, but we still need to be able configure the serial and parallel refinement in my opinion. Or are we looking to scale from epm only?

rfhaque: @michaelmckinsey1 The problem can still be manually configured using `nx, ny, nz, rs, rp` if `epm` is set to 0. Scaling with epm is much easier than configuring the mesh dimensions and refinement.
I believe, for now, we should scale using `epm` only. Should there be a use case for scaling with nx,ny,nz, rs, rp, we can handle that in a separate PR

pearce8: @michaelmckinsey1 What parameters do you need for the workflow you are running?  We could have 2 workloads: 1) epm based, 2) using the original parameters

pearce8: @michaelmckinsey1 are you also passing in the mesh file as a parameter for your workflow?

pearce8 (approved):

experiments/laghos/experiment.py
michaelmckinsey1: Correction: `nx, ny, nz, rs, rp` are all mutually exclusive with `epm`. I thought it was just `nx, ny, nz` because that is what is in the laghos description.

PR #1376, opened 2026-06-05
removing affinity

pearce8:


pearce8:
@michaelmckinsey1 I am trying to see if we need the copy of the affinity package.py, but the one test I see using it is failing.  Can you please figure out if the modifier works with the upstream package?

michaelmckinsey1:
> @michaelmckinsey1 I am trying to see if we need the copy of the affinity package.py, but the one test I see using it is failing. Can you please figure out if the modifier works with the upstream package?

That test is failing because of the chai failure, so I can't tell until that is reverted.

pearce8:
@michaelmckinsey1 Please restart CI and check if the affinity test works.

michaelmckinsey1:
https://lc.llnl.gov/gitlab/benchpark/benchpark/-/jobs/3732755

PR #1387, opened 2026-06-12
skip coverage on forks

michaelmckinsey1:
## Description

- Currently, dryruns always appear as failure on forks because the coverage step will always fail. This skips coverage on forks, so we can still easily tell if the unit tests are passing.

I made a demo PR https://github.com/llnl/benchpark/pull/1388 to show this works (this is the same PR but from a fork) and the tests are skipped there.


michaelmckinsey1:
example of skipped action https://github.com/llnl/benchpark/actions/runs/27431826487/job/81084394453?pr=1388

pearce8 (approved):

PR #1390, opened 2026-06-12
Fix `py-scaffold` unit test

michaelmckinsey1:
## Description

- [x] Depends on #1349
- [x] Test was missing `rocm=7.1.1`, default rocm=6.4.3 won't work.
- [x] Remove tioga test since we are focusing on tuolumne recently
- [x] Need to check if gitlab test is passing on this branch after #1349 is merged


michaelmckinsey1:
The test actually runs as expected (https://lc.llnl.gov/gitlab/benchpark/benchpark/-/jobs/3819673). The issue is that after completion of the benchmark, there is always this error `corrupted size vs. prev_size in fastbins`, specifically on ROCm. We typically ignore this because it does not affect the benchmark run, however it fails the CI test because it then returns an exit code 134.

@joshuamvining Any ideas how to force this to exit code 0? The only thing I can think of is checking in `.gitlab/utils/run-experiment.sh` for exit code 134 and instead returning 0.

joshuamvining:
> The test actually runs as expected (https://lc.llnl.gov/gitlab/benchpark/benchpark/-/jobs/3819673). The issue is that after completion of the benchmark, there is always this error `corrupted size vs. prev_size in fastbins`, specifically on ROCm. We typically ignore this because it does not affect the benchmark run, however it fails the CI test because it then returns an exit code 134.
> 
> @joshuamvining Any ideas how to force this to exit code 0? The only thing I can think of is checking in `.gitlab/utils/run-experiment.sh` for exit code 134 and instead returning 0.

I think you could probably modify [.gitlab/bin/exit-codes](https://github.com/llnl/benchpark/blob/develop/.gitlab/bin/exit-codes) to have an exception for error code 134 whenever the benchmark is py-scaffold. That may be a cleaner place to put the exception than in `run-experiment.sh`

michaelmckinsey1:
Test is passing now https://lc.llnl.gov/gitlab/benchpark/benchpark/-/jobs/3826538

pearce8 (approved):

pearce8 (changes requested):
@michaelmckinsey1  Please resolve the conflict.

michaelmckinsey1:
> @michaelmckinsey1 Please resolve the conflict.

@pearce8 Fixed

pearce8 (approved):

PR #1401, opened 2026-06-17
Refactor `riken-cloud` system

michaelmckinsey1:
## Description

- [x] Refactor of PR #1294. Closes #1294.

We are recommending refactoring the `riken-cloud` system into 4 separate systems, because the clusters do not share a common software/compiler stack. Compilers are different, package versions are different, etc. The `riken-cloud` system (#1294) has independent package/compiler/software definitions for each cluster, which is a good indication that the clusters should be separate systems.

Systems like `llnl-cluster` contain multiple clusters, because these clusters share a software stack and compilers. It is not mandatory they share the same hardware, but your system must contain the same underlying environment, and hardware from different vendors will of course have different environments.


pearce8 (dismissed):

yoshifuminakamura:
@michaelmckinsey1 @pearce8 
Thank you for the refactor. Splitting `riken-cloud` into `riken-fx700`, `riken-gh200`, `riken-dgx`, and `riken-genoa` makes sense to us. We were finding it increasingly difficult to maintain a 2,000-line file that relied heavily on if-else branching.

One small question: in `systems/riken-dgx/system.py`, OpenMPI still uses `cuda_arch=90`, while `system_specific_variables()` sets `cuda_arch` to `100`. This comes from our original code, but should the OpenMPI setting also be updated to `100` for GB10?

Aside from that, I am OK with this refactor. Thank you for cleaning this up.

michaelmckinsey1:
> @michaelmckinsey1 @pearce8 Thank you for the refactor. Splitting `riken-cloud` into `riken-fx700`, `riken-gh200`, `riken-dgx`, and `riken-genoa` makes sense to us. We were finding it increasingly difficult to maintain a 2,000-line file that relied heavily on if-else branching.
> 
> One small question: in `systems/riken-dgx/system.py`, OpenMPI still uses `cuda_arch=90`, while `system_specific_variables()` sets `cuda_arch` to `100`. This comes from our original code, but should the OpenMPI setting also be updated to `100` for GB10?
> 
> Aside from that, I am OK with this refactor. Thank you for cleaning this up.

Yes I would image the cuda_arch should match. I changed it to 100 github.com/llnl/benchpark/pull/1401/commits/b3ceb6b5081d86f81222498037c14cc95dd65909

PR #1412, opened 2026-06-22
fix branson failure

### LINKED ISSUE #1286: branson+rocm `fatal error: 'mpi.h' file not found`
https://lc.llnl.gov/gitlab/benchpark/benchpark/-/jobs/3478454

We noticed this fails
```
benchpark experiment init --dest=branson tuolumne branson +rocm
```
while this works (does not incur the build failure linked above)
```
benchpark experiment init --dest=branson2 tuolumne branson +rocm caliper=time,mpi
```
The only difference is whether we also add caliper, but the same MPI is used in both cases.

michaelmckinsey1:
## Description

- [x] This is a temporary fix by building with Caliper. This should fix 2 unit test failures
- [ ] https://github.com/spack/spack-packages/pull/4411 should render this unnecessary. 

@michaelmckinsey1 is there an easy way to undo after the permanent fix if we merge this now?

### DIFF
--- a/.gitlab/tests/nightly.yml
+++ b/.gitlab/tests/nightly.yml
@@ -57,13 +57,12 @@ include:
           - remhos
           - sparta-snl
         VARIANT: [+rocm]
-      # GTL experiments
+      # Needs caliper for mpi
       - HOST: tuolumne
         ARCHCONFIG: llnl-elcapitan
         SCHEDULER_PARAMETERS: $ELCAP_PARAMS
         BENCHMARK: [branson]
-        VARIANT: [+rocm]
-        SYSTEM_ARGS: [+gtl]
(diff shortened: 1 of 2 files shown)


michaelmckinsey1:
@scheibelp My understanding is the permanent fix for this issue is a PR on spack?

scheibelp:
Yes I anticipate that https://github.com/spack/spack-packages/pull/4411 (when it is complete) will make it so this isn't needed (but I think it makes sense to merge this PR for now).

scheibelp (approved):

michaelmckinsey1:
fixed in github.com/lanl/branson/pull/72

michaelmckinsey1:
Fixed by https://github.com/lanl/branson/pull/72

PR #1414, opened 2026-06-23
Ci resource allocation/release job fix

joshuamvining:
## Description

- This patch should change CI logic to only create jobs for touched benchmarks/systems, instead of always creating jobs for every benchmark/system.

- [x] Update docs

## Tested Success Pipeline

- [Change to a ramble application](https://lc.llnl.gov/gitlab/benchpark/benchpark/-/commit/d2e5cf0fc5e5f991823ccc5871eb1b8252cb3542)
- [Change to a spack package](https://lc.llnl.gov/gitlab/benchpark/benchpark/-/commit/e5c657694e80d40960f3148099d5824a024f2068)
- [Change to a experiment](https://lc.llnl.gov/gitlab/benchpark/benchpark/-/commit/64618dacd4714aa57e13c1a25fd9b74f20bd4211)

## Tested Failure Pipeline

- [osu-mb failure](https://lc.llnl.gov/gitlab/benchpark/benchpark/-/commit/a22b29ae924b80ea74a1eb51cee929e80d96b1d7)

### DIFF
--- a/.gitlab/tests/shared_flux_clusters.yml
+++ b/.gitlab/tests/shared_flux_clusters.yml
@@ -41,6 +41,7 @@ allocate_resources_tioga:
 .release_resources_flux:
   extends: .on_host_template
   stage: release-resources
+  when: always
   # Do not cleanup dir yet
   script: [. .gitlab/utils/cancel-flux.sh --no-clean]
 release_resources_flux_tuolumne:
--- a/.gitlab/tests/shared_flux_clusters.yml
+++ b/.gitlab/tests/shared_flux_clusters.yml
@@ -19,15 +19,15 @@ workflow:
       if [[ "$CI_JOB_STATUS" == "canceled" ]]; then
         . .gitlab/utils/cancel-flux.sh
       fi
-allocate_resources_tuolumne:
-  extends: .allocate_resources_flux
-  parallel:
-    matrix:
-      - HOST: tuolumne
-        SHARED_ALLOC: $TUOLUMNE_SHARED_ALLOC
-        GPUMODE: SPX
-  tags: [shell, tuolumne]
-  needs: [tuolumne-up-check]
+# allocate_resources_tuolumne:
+#   extends: .allocate_resources_flux
+#   parallel:
+#     matrix:
+#       - HOST: tuolumne
+#         SHARED_ALLOC: $TUOLUMNE_SHARED_ALLOC
+#         GPUMODE: SPX
+#   tags: [shell, tuolumne]
+#   needs: [tuolumne-up-check]
 allocate_resources_tioga:
   extends: .allocate_resources_flux
   variables:
@@ -44,14 +44,14 @@ allocate_resources_tioga:
   when: always
   # Do not cleanup dir yet
   script: [. .gitlab/utils/cancel-flux.sh --no-clean]
-release_resources_flux_tuolumne:
-  extends: .release_resources_flux
-  needs: [allocate_resources_tuolumne, run_tests_flux_tuolumne]
-  parallel:
-    matrix:
-      - HOST: tuolumne
-        GPUMODE: SPX
-  tags: [shell, tuolumne]
+# release_resources_flux_tuolumne:
+#   extends: .release_resources_flux
+#   needs: [allocate_resources_tuolumne, run_tests_flux_tuolumne]
+#   parallel:
+#     matrix:
+#       - HOST: tuolumne
+#         GPUMODE: SPX
+#   tags: [shell, tuolumne]
 release_resources_flux_tioga:
   extends: .release_resources_flux
   needs: [allocate_resources_tioga, run_tests_flux_tioga]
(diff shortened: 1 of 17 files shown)


.gitlab/tests/shared_flux_clusters.yml
```diff
@@ -39,21 +40,29 @@ allocate_resources_tioga:
...
   # Do not cleanup dir yet
   script: [. .gitlab/utils/cancel-flux.sh --no-clean]
 release_resources_flux_tuolumne:
   extends: .release_resources_flux
-  needs: [allocate_resources_tuolumne, run_tests_flux_tuolumne]
+  rules: !reference [.resource_rules_tuolumne, rules]
+  needs:
+  - job: allocate_resources_tuolumne
+  - job: run_tests_flux_tuolumne
+    optional: true
```
michaelmckinsey1: Why would the release job be optional?

.gitlab/utils/rules.yml
```diff
@@ -17,6 +13,102 @@
...
+        - bin/**/*
+        - "**/amg2023/**/*"
+        - "**/kripke/**/*"
+        - "**/laghos/**/*"
+        - "**/py-scaffold/**/*"
+        - "**/raja-perf/**/*"
+        - "**/remhos/**/*"
+        - systems/llnl-elcapitan/**/*
+        - modifiers/**/*
+        - var/**/*
```
michaelmckinsey1, line 82: Why does this list vary per system? Shouldn't this be the same list of changed directories across all systems?

michaelmckinsey1 (changes requested):
1. Can you fix lint?
1. Can you comment in this PR which pipeline where you were testing this?
1. Will tests now only run if experiments/systems/lib/modifiers are changed? I think we still need to be running tests when repos/ changes for both spack packages and ramble applications.

.gitlab/tests/shared_flux_clusters.yml
joshuamvining: This is saying the `run_tests_...` job is optional for being needed for the `release_resources_...` job. Because the `run_tests_...` job is created with a matrix, without the `optional: true`, the release job would require every job in the matrix exist, which would be in direct conflict with the new rules because the point is to only create jobs in the matrix if that benchmark/system was touched by changes.

.gitlab/utils/rules.yml
joshuamvining: This list directly corresponds with what benchmarks are currently being tested on each system. It is basically a one to one to the `run_tests_...` job for a system.

If we did just have the list include all benchmarks across the board, then a change to an untested benchmark would still trigger the allocate and release jobs, but not a test job.

michaelmckinsey1: Ah I see. I'm thinking of the pain of maintaining both the lists of tests and these file lists now. Do you think there could be a better way to configure this? At minimum, we would need to update the CI documentation.

.gitlab/utils/rules.yml
```diff
@@ -1,22 +1,113 @@
...
+        - '**/branson/**/*'
+        - '**/hpcg/**/*'
+        - '**/hpl/**/*'
+        - '**/ior/**/*'
+        - '**/kripke/**/*'
+        - '**/laghos/**/*'
+        - '**/lammps/**/*'
+        - '**/osu-micro-benchmarks/**/*'
+        - '**/phloem/**/*'
+        - '**/raja-perf/**/*'
```
michaelmckinsey1: For any benchmark that is hyphenated, spack moved to underscores awhile ago, so this rule would not catch changes to the spack package version `raja_perf`. I don't recall if changing our experiments to be underscores will work, but the spack packages definitely have to be underscores.

.gitlab/utils/rules.yml
```diff
@@ -1,22 +1,113 @@
-.on_host_template:
+.run_test_rules:
   rules:
     - if: $CI_PIPELINE_SOURCE == "schedule" || $ADVANCED_JOB == "ON" && $CI_COMMIT_BRANCH
         != "main" && $CI_COMMIT_BRANCH != "develop" && $ALL_TARGETS != "ON"
       when: never
-    - if: $CI_JOB_NAME =~ /release_resources|cleanup_pipeline_dir/ && $CI_PIPELINE_SOURCE
-        != "schedule"
-      when: always
```
michaelmckinsey1, line 8: from codex, it seems like the release job will be skipped if all of the tests fail? 
 
- [P1] Preserve when: always on release-resource jobs — /Users/mckinsey1/Documents/repos/benchpark/.gitlab/utils/rules.yml:42-42
    These rule sets are now reused by `release_resources_*`, but their only positive branch sets when: on_success. In GitLab the matched rule's when overrides the job-level when: always, so a
    failing `run_tests_*` job will skip the corresponding release job instead of canceling the shared Slurm/Flux allocation. This shows up whenever a cluster test fails, which is exactly when
    the cleanup stage needs to run.

michaelmckinsey1 (changes requested):
@joshuamvining How does this break your modifier, since you rely on running all of the tests on the develop branch? Wouldn't none of the tests run with these changes?

joshuamvining:
> @joshuamvining How does this break your modifier, since you rely on running all of the tests on the develop branch? Wouldn't none of the tests run with these changes?

#1410 would prevent this from breaking my modifier. That PR allows my modifier to comb through many previous develop pipelines to find the latest versions of each job artifact it is trying to compare.

.gitlab/utils/rules.yml
joshuamvining: This should be fixed in latest commit.

Tested here: [Pipeline 724564](https://lc.llnl.gov/gitlab/benchpark/benchpark/-/pipelines/724564)

.gitlab/utils/rules.yml
joshuamvining: Fixed in latest commit, I changed the glob pattern to catch both hyphens and underscores.

.gitlab/utils/rules.yml
joshuamvining: Yeah I wasn't the biggest fan of the solution either but it is the only thing I could think of. Ideally the matrix and the watch rules could both inherit their benchmarks from the same source, I just don't know how I could make that work with yaml.

And yes if this solution is the one we go with, I will update docs.

michaelmckinsey1 (approved):

PR #1415, opened 2026-06-24
Add Frontier system support

matthewc2003:
## Description

- [x] Add system.py for ORNL's Frontier system
- [x] tested amg2023, kripke, and raja-perf all +rocm on Frontier


modifiers/allocation/modifier.py
```diff
@@ -325,6 +325,10 @@ def slurm_instructions(self, v):
                 srun_opts.append(f"-n {v.n_ranks_per_node}")
             else:
                 srun_opts.append(f"-n {v.n_ranks}")
+
+            if v.n_nodes:
+                sbatch_opts.append(f"-N {v.n_nodes}")
```
michaelmckinsey1: Was this from the original PR? This should be redundant to line 339 https://github.com/llnl/benchpark/blob/8dc25069a90caf5f9af3652d0ac6da3bc9a242ed/modifiers/allocation/modifier.py#L339

systems/olcf-frontier/system.py
```diff
@@ -0,0 +1,774 @@
...
+            elif self.spec.satisfies("gpumode=TPX"):
+                gpu_factor = 3
+            elif self.spec.satisfies("gpumode=CPX"):
+                gpu_factor = 6
+            extra_batch_opts += f"--setattr=gpumode={self.spec.variants['gpumode'][0]}\n--conf=resource.rediscover=true"
+
+            # Rabbits
+            mt_point = self.spec.variants["mount_point"][0]
+            if mt_point != "none" and "rabbits" in mt_point:
+                extra_batch_opts += f"\n-S dw={mt_point.lstrip('rabbits_')}"
```
michaelmckinsey1: These should be removed I think. The first part is specific to MI300A, and the second is specific to rabbit storage which I don't think Frontier has.

systems/olcf-frontier/system.py
```diff
@@ -0,0 +1,774 @@
...
+
+            # opts.update(
+            #     {
+            #         "gpu_factor": gpu_factor,
+            #         "extra_batch_opts": extra_batch_opts,
+            #     }
+            # )
```
michaelmckinsey1: Remove

michaelmckinsey1 (changes requested):
Can you list which benchmarks you have tested on Frontier?

modifiers/allocation/modifier.py
matthewc2003: Without this, when I tried testing on Frontier, the generated slurm script does not also specify the nodes parameter which is required to submit a job.

matthewc2003: I changed this because the nodes section was not applying the change for sbatch, only srun

systems/olcf-frontier/system.py
matthewc2003: removed

systems/olcf-frontier/system.py
matthewc2003: removed

michaelmckinsey1 (approved):

PR #1435, opened 2026-07-07
Add xsbench spack

### LINKED ISSUE #1417: Add XSBench to benchpark
https://github.com/ANL-CESAR/XSBench
https://github.com/spack/spack-packages/blob/develop/repos/spack_repo/builtin/packages/xsbench/package.py

@matthewc2003 As you work through HecBench, please also integrate OpenMC.  I think the two integrations should be very similar.  Bonus points if you can speed up the integration with any of the generative AI tools - if you do, please write up the process so we can consider using it or building on it in the future.

@RanivG This is a miniature version of OpenMC, worth looking at before OpenMC itself.  After @matthewc2003 has written the application.py and experiment.py, you could work on instrumenting with Caliper so we get to the measurements.

Giorgi's will be using this benchmark for his compiler work as well, so we will get the compiler info from him at a later time. 

@slabasan @michaelmckinsey1 FYI.

matthewc2003:
## Description

Add xsbench to benchpark


experiments/xsbench/experiment.py
```diff
@@ -0,0 +1,68 @@
...
+    )
+
+    variant(
+        "version",
+        default="20",
+        values=("20", "latest", "caliper-support", "mpi-fix"),
+        description="Which XSBench version to use.",
+    )
+
+    # variant(
```
pearce8, line 35: Hopefully this can just be removed

repos/spack_repo/benchpark/packages/xsbench/package.py
```diff
@@ -0,0 +1,122 @@
...
+
+class Xsbench(MakefilePackage, CudaPackage, ROCmPackage):
+    """XSBench is a mini-app representing a key computational
+    kernel of the Monte Carlo neutronics application OpenMC.
+    A full explanation of the theory and purpose of XSBench
+    is provided in docs/XSBench_Theory.pdf."""
+
+    homepage = "https://github.com/ANL-CESAR/XSBench/"
+    url = "https://github.com/ANL-CESAR/XSBench/archive/v13.tar.gz"
+    git = "https://github.com/daboehme/XSBench.git"
```
pearce8, line 21: @daboehme Do you have plans to:
1) merge the caliper-support branch into the master branch in your fork?
2) Merge your fork back to the original repo (or at least see if they would be interested)?

@RanivG After the questions above converge, please point at the right code here and test it to make sure it is still working.

daboehme: Merging the code into the master branch in my fork wouldn't make much of a difference I think.

Upstreaming the caliper-support branch will need some work to make Caliper support a compile-time option, the way it's currently written Caliper/Adiak are required. I also have no idea who maintains XSBench and if they would be interested in it.

I can update the caliper-support branch and put up a PR if requested, but either way it'll probably take some time until that is merged. For the short term I'd just keep it as-is and point to my branch when building with Caliper support.

PR #1436, opened 2026-07-07
added systems jcahpc-miyabi-g

yoshifuminakamura:
## Description

- Adds a Benchpark system configuration for JCAHPC Miyabi-G, the GH200 GPU partition of the Miyabi supercomputer.
- Includes PBS queue definitions, JCAHPC group accounting support, NVHPC/CUDA/HPC-X external package configuration, and compiler configuration for the Miyabi-G module environment.
- Updates PBS allocation handling so systems can request a site-specific per-node core count via `n_cores_per_node`; Miyabi-G uses this to request the full GH200 CPU cpuset and avoid MPI/hwloc issues with too-small `ncpus` values.
- Also comments out the PBS `n_gpus` chunk-resource emission in the allocation modifier; on Miyabi-G, GPU allocation is already represented by `n_nodes`/`select`, and adding a separate `gpus` chunk resource is unnecessary for this site configuration.
- Dependencies: existing Miyabi-G site modules and the existing Benchpark hardware description for NVIDIA GH200.
- Relevant issues: none.

- [x] Add/modify `systems/jcahpc-miyabi-g/system.py` file
- [x] Reuse existing `systems/all_hardware_descriptions/NVIDIA-neoverse-GH200-Infiniband/hardware_description.yaml`, which appears in the [docs catalogue](https://software.llnl.gov/benchpark/system-list.html)

- [ ] Update docs
- [ ] Update `.github/workflows` and `.gitlab/tests` unit tests (if needed)
- [ ] Ensure `rules.yml` resource_rules_* change rules match `shared_*_clusters.yml` run_test_* job matricies. (if needed)

### DIFF
--- a/modifiers/allocation/modifier.py
+++ b/modifiers/allocation/modifier.py
@@ -19,6 +19,8 @@ class AllocOpt(Enum):
     N_GPUS = 6
     N_CORES_PER_NODE = 7
     OMP_NUM_THREADS = 8
+    PBS_NCPUS_PER_NODE = 9
+    PBS_EMIT_GPUS = 10
 
     # Descriptions of resources available on systems
     SYS_GPUS_PER_NODE = 100
@@ -509,29 +511,20 @@ class Allocation(BasicModifier):
         if v.n_threads_per_proc and v.n_threads_per_proc != 1:
             node_spec.append(f"ompthreads={v.n_threads_per_proc}")
 
-        # A system may request a specific per-node core count via the
-        # `n_cores_per_node` variable (e.g. sites that allocate whole nodes and
-        # need the full cpuset). When it is not set we fall back to the cores
-        # implied by the rank/thread layout. This keeps site policy out of the
-        # shared modifier: a system that wants a different value (MIG slices,
-        # partial-node sharing, ...) simply sets `n_cores_per_node` accordingly.
-        n_cpus_per_node = v.n_cores_per_node or (
-            v.n_ranks_per_node * v.n_threads_per_proc
+        # Some PBS sites need a scheduler-specific ncpus floor that is larger
+        # than the rank/thread layout, for example to avoid a too-small cpuset.
+        layout_ncpus = v.n_ranks_per_node * v.n_threads_per_proc
+        n_cpus_per_node = max(
+            v.pbs_ncpus_per_node or 0,
+            v.n_cores_per_node or 0,
+            layout_ncpus,
         )
 
         node_spec.append(f"ncpus={n_cpus_per_node}")
 
-        # No GPU chunk sub-resource is emitted here.
-        # On Miyabi-G, GPU allocation is already reflected in n_nodes/select, so
-        # adding a separate GPU chunk resource is unnecessary.
-        #
-        # Avoid the PBS GPU chunk path here, as it is not needed for this site
-        # configuration and can fail when invoked.
-        #
-        #
-        # if v.n_gpus:
-        #    gpus_per_rank = self.gpus_as_gpus_per_rank(v.n_gpus)
-        #    node_spec.append(f"gpus={gpus_per_rank}")
+        if v.n_gpus and v.pbs_emit_gpus != 0:
+            gpus_per_rank = self.gpus_as_gpus_per_rank(v)
+            node_spec.append(f"gpus={gpus_per_rank}")
 
         if node_spec:
             batch_opts.append(f"-l {':'.join(node_spec)}")
--- a/modifiers/allocation/modifier.py
+++ b/modifiers/allocation/modifier.py
@@ -19,7 +19,7 @@ class AllocOpt(Enum):
     N_GPUS = 6
     N_CORES_PER_NODE = 7
     OMP_NUM_THREADS = 8
-    PBS_NCPUS_PER_NODE = 9
+    PBS_MIN_NCPUS_PER_NODE = 9
     PBS_EMIT_GPUS = 10
 
     # Descriptions of resources available on systems
@@ -511,11 +511,11 @@ class Allocation(BasicModifier):
         if v.n_threads_per_proc and v.n_threads_per_proc != 1:
             node_spec.append(f"ompthreads={v.n_threads_per_proc}")
 
-        # Some PBS sites need a scheduler-specific ncpus floor that is larger
-        # than the rank/thread layout, for example to avoid a too-small cpuset.
+        # Some PBS sites need a scheduler-specific ncpus floor, for example to
+        # avoid a too-small cpuset.
         layout_ncpus = v.n_ranks_per_node * v.n_threads_per_proc
         n_cpus_per_node = max(
-            v.pbs_ncpus_per_node or 0,
+            v.pbs_min_ncpus_per_node or 0,
             v.n_cores_per_node or 0,
             layout_ncpus,
         )
--- a/modifiers/allocation/modifier.py
+++ b/modifiers/allocation/modifier.py
@@ -529,7 +529,7 @@ class Allocation(BasicModifier):
         # configuration and can fail when invoked.
         #
         #
-        #if v.n_gpus:
+        # if v.n_gpus:
         #    gpus_per_rank = self.gpus_as_gpus_per_rank(v.n_gpus)
         #    node_spec.append(f"gpus={gpus_per_rank}")
 
--- a/modifiers/allocation/modifier.py
+++ b/modifiers/allocation/modifier.py
@@ -509,12 +509,29 @@ class Allocation(BasicModifier):
         if v.n_threads_per_proc and v.n_threads_per_proc != 1:
             node_spec.append(f"ompthreads={v.n_threads_per_proc}")
 
-        n_cpus_per_node = v.n_ranks_per_node * v.n_threads_per_proc
+        # A system may request a specific per-node core count via the
+        # `n_cores_per_node` variable (e.g. sites that allocate whole nodes and
+        # need the full cpuset). When it is not set we fall back to the cores
+        # implied by the rank/thread layout. This keeps site policy out of the
+        # shared modifier: a system that wants a different value (MIG slices,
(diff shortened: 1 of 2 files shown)


modifiers/allocation/modifier.py
```diff
@@ -509,12 +509,29 @@ def pbs_instructions(self, v):
...
+        )
+
         node_spec.append(f"ncpus={n_cpus_per_node}")
 
-        if v.n_gpus:
-            gpus_per_rank = self.gpus_as_gpus_per_rank(v.n_gpus)
-            node_spec.append(f"gpus={gpus_per_rank}")
+        # No GPU chunk sub-resource is emitted here.
+        # On Miyabi-G, GPU allocation is already reflected in n_nodes/select, so
+        # adding a separate GPU chunk resource is unnecessary.
```
michaelmckinsey1: Is this commented out because it breaks Miyabi-G? Should `gpus=` be unset for a "full node"?

I suppose since there is only 1 gpu the only option is 0 or 1.

modifiers/allocation/modifier.py
```diff
@@ -509,12 +509,29 @@ def pbs_instructions(self, v):
...
             node_spec.append(f"ompthreads={v.n_threads_per_proc}")
 
-        n_cpus_per_node = v.n_ranks_per_node * v.n_threads_per_proc
+        # A system may request a specific per-node core count via the
+        # `n_cores_per_node` variable (e.g. sites that allocate whole nodes and
+        # need the full cpuset). When it is not set we fall back to the cores
+        # implied by the rank/thread layout. This keeps site policy out of the
+        # shared modifier: a system that wants a different value (MIG slices,
+        # partial-node sharing, ...) simply sets `n_cores_per_node` accordingly.
+        n_cpus_per_node = v.n_cores_per_node or (
```
michaelmckinsey1: We avoid setting the number of cpus/tasks based on the number of cores on a node, because the number of tasks is set per-experiment.

This is essentially a full-node setting @pearce8, which we have avoided in the past.

@yoshifuminakamura Is it not possible to run using less than 72 cores on the system?

systems/jcahpc-miyabi-g/system.py
```diff
@@ -0,0 +1,381 @@
...
+    variant(
+        "group",
+        default="none",
+        multi=False,
+        description=(
+            "JCAHPC project/group code for accounting, emitted as "
+            "'#PBS -W group_list=<group>' (e.g. group=gz00). Required to "
+            "actually submit a job; 'none' omits the directive."
+        ),
+    )
```
michaelmckinsey1, line 120: Is a group different from a bank here? Would `-A <group>` fail?

systems/jcahpc-miyabi-g/system.py
```diff
@@ -0,0 +1,381 @@
...
+        "24.5": ("12.4", "4.1.7"),
+        "24.9": ("12.6", "4.1.7"),
+        "24.11": ("12.6", "4.1.7"),
+        "25.1": ("12.6", "4.1.7"),
+        "25.3": ("12.8", "4.1.7"),
+        "25.5": ("12.9", "4.1.9"),
+        "25.9": ("13.0", "4.1.9"),
+        "25.11": ("13.0", "4.1.9"),
+        "26.3": ("13.1", "4.1.9"),
+    }
```
michaelmckinsey1: @yoshifuminakamura Can we specify this in same way you did for `riken-gh200` for consistency? https://github.com/llnl/benchpark/blob/72ee531ef69b874020bc15cdab4e9cdaab82ee77/systems/riken-gh200/system.py#L66-L76

systems/jcahpc-miyabi-g/system.py
```diff
@@ -0,0 +1,381 @@
...
+    def system_specific_variables(self):
+        variables = {
+            "cuda_arch": "90",
+            # Miyabi-G nodes are allocated whole (one GH200 per node, no MIG
+            # sharing), so request the node's full core count. Otherwise the
+            # generated `ncpus` collapses to the rank/thread count (e.g. 1 for a
+            # single-rank job); PBS then pins the job to that narrow cpuset and
+            # HPC-X/Open MPI's hwloc segfaults building the topology on GH200.
+            "n_cores_per_node": self.sys_cores_per_node,
+        }
```
michaelmckinsey1, line 372: These variables should be set in the `id_to_resources` section. I didn't catch this with the riken-gh200 PR

michaelmckinsey1 (changes requested):
I think the allocation modifier changes could be breaking.

modifiers/allocation/modifier.py
yoshifuminakamura: Looking at this again, the issue is not necessarily that `gpus=` itself breaks Miyabi-G. The pre-existing PBS code path appears to call `self.gpus_as_gpus_per_rank(v.n_gpus)`, but `gpus_as_gpus_per_rank()` expects the allocation object `v`, not the integer `v.n_gpus`.

So this block may fail before we even test whether Miyabi-G accepts `gpus=1`. Since each Miyabi-G node has one GPU, `gpus=1` may be fine. I should not have commented this out globally; I will restore the PBS `gpus=` emission and fix the call to use `self.gpus_as_gpus_per_rank(v)`.

modifiers/allocation/modifier.py
yoshifuminakamura: It is possible to run with fewer than 72 cores on Miyabi-G. The issue I hit is more specific: for example, saxpy requests a single thread/rank, and the current PBS modifier then generates `ncpus=1`. On Miyabi-G that gives the job a one-CPU cpuset, and HPC-X/Open MPI's hwloc fails immediately at program startup.

So the requirement is not "always use all 72 cores"; it is "avoid a too-small cpuset such as `ncpus=1`". Allocating a larger CPU cpuset avoids the hwloc startup failure. I can rework this so it is expressed as a Miyabi-G-specific minimum/requested PBS `ncpus` value rather than framing it as a full-node setting.

systems/jcahpc-miyabi-g/system.py
yoshifuminakamura: `qsub` with `-A` fails on Miyabi.
I don't fully understand the meaning of “bank,” but in Miyabi, a “group” refers to the ID of a project group to which multiple people can belong. A budget (available node time) is linked to this group. Since a single user can belong to multiple project groups with the same uid, you must specify which project group the job belongs to using the `-W` option when submitting a job.

systems/jcahpc-miyabi-g/system.py
yoshifuminakamura: Yes, I can change this to match the `riken-gh200` style for consistency.

systems/jcahpc-miyabi-g/system.py
yoshifuminakamura: I will move variables into `id_to_resources`.

yoshifuminakamura:
I pushed updates addressing the review feedback:

- restored the PBS `gpus=` emission path and fixed the call to `gpus_as_gpus_per_rank(v)`
- reworked the Miyabi-G `ncpus` handling as a site-specific PBS request instead of changing the experiment task/thread layout
- kept Miyabi-G group submission through `-W group_list=...` because `qsub -A` fails on this system
- moved the relevant variables into `id_to_resources`
- updated the NVHPC/CUDA handling to follow the `riken-gh200` style more closely

The remaining CI failure appears to be unrelated to this PR: `yamlfix --check .` reports formatting changes in `.gitlab/tests/nightly.yml` and `.gitlab/tests/shared_slurm_clusters.yml`, while this PR only changes `modifiers/allocation/modifier.py` and `systems/jcahpc-miyabi-g/system.py`.

Could you take another look when you have a chance?

modifiers/allocation/modifier.py
```diff
@@ -509,11 +511,19 @@ def pbs_instructions(self, v):
         if v.n_threads_per_proc and v.n_threads_per_proc != 1:
             node_spec.append(f"ompthreads={v.n_threads_per_proc}")
 
-        n_cpus_per_node = v.n_ranks_per_node * v.n_threads_per_proc
+        # Some PBS sites need a scheduler-specific ncpus floor that is larger
+        # than the rank/thread layout, for example to avoid a too-small cpuset.
+        layout_ncpus = v.n_ranks_per_node * v.n_threads_per_proc
+        n_cpus_per_node = max(
+            v.pbs_ncpus_per_node or 0,
+            v.n_cores_per_node or 0,
```
michaelmckinsey1: If `pbs_ncpus_per_node` = `n_cores_per_node` on Miyabi why do we need both? Can we remove `pbs_ncpus_per_node` and just use the value for `n_cores_per_node`?

If this is to avoid this issue https://github.com/llnl/benchpark/pull/1436#discussion_r3548615946 could we rewrite it as `pbs_min_ncpus_per_node` and then make the value in Miyabi 2, so 1 CPU allocations are not generated (if I am understanding the problem correctly).

michaelmckinsey1 (changes requested):
@yoshifuminakamura Sorry for the delay, I intended to rereview sooner. I think I just have 1 more outstanding question.

modifiers/allocation/modifier.py
yoshifuminakamura: Yes, that makes sense. The intent is only to avoid generating too-small PBS cpusets such as `ncpus=1`, not to require a full node. I renamed this to `pbs_min_ncpus_per_node` and set the Miyabi-G value to 2 in the latest push, while keeping the final PBS `ncpus` as the max of the experiment layout and that site-specific minimum.

michaelmckinsey1 (approved):
`v.pbs_min_ncpus_per_node` and `v.pbs_emit_gpus` will evaluate to `None` for other pbs systems, which is handled by the new changes in `modifiers/allocation/modifier.py`, so the changes there will not affect the existing `aws-pcluster scheduler=pbs` system.

pearce8 (approved):

PR #1444, opened 2026-07-08
Fix sparta-snl test size

michaelmckinsey1:
```
ERROR on proc 2: Failed to reallocate 4194304 bytes for array grid:cells
```
Fixed by lowering x and y min/max


michaelmckinsey1:
sparta unit tests passing

pearce8 (approved):

PR #1447, opened 2026-07-08
Update outdated scripts

michaelmckinsey1:
## Description

- [x] These scripts are not tested, so they have become out of date. I have been using `lib/scripts/diffExperimentBuilds.py` recently, so I applied similar updates to all of the scripts.


pearce8:
@michaelmckinsey1 @slabasan This PR broke lint in develop.  I opened #1468 to try to fix.

PR #1459, opened 2026-07-14
Added externals back to metadata collection

joshuamvining:
## Description
- Githash now lists externals


slabasan:
@joshuamvining Please fix the lint issues

PR #1479, opened 2026-08-05
Gitlab pipeline metadata infrastructure, end of pipeline table generation, and nightly push to benchpark perf repo

joshuamvining:
## Description

- Reworks the metadata infrastructure to generate per host per tests .json files describing each test
- Generates an aggregate .json file describing the whole pipeline
- Generates a end of pipeline summary table describing failures and changes
- Pushes nightly test metadata to [https://github.com/llnl/benchpark-performance](https://github.com/llnl/benchpark-performance )


.gitlab/utils/generate_test_summary_csv.py
michaelmckinsey1, line 1: Was this moved to `generate_test_status_table.py`? Or what was this functionality and why was it deprecated?

.gitlab-ci.yml
```diff
@@ -92,6 +92,7 @@ stages:
   - test
   - release-resources
   - summarize
+  - deploy-performance
```
michaelmckinsey1: Would this more aptly be named `publish-performance`?

.gitlab/utils/fetch-job-artifact.sh
michaelmckinsey1, line 1: In fetch-job-artifact.sh, fetch_job_artifact() handles optional only around the unzip. If curl --fail fails (expired artifacts, 404, transient 5xx) the function returns nonzero, and in stage mode it's called bare inside a while read loop under set -euo pipefail.

^ claude

.gitlab/utils/write-test-metadata.sh
```diff
@@ -0,0 +1,158 @@
...
+          available: false,
+          reason: $reason,
+          regressed: false
+        }' > "${merged_performance_json}"
+fi
+
+printf '[]\n' > "${changed_files_json}"
+diff_base="origin/${BASELINE_REF:-develop}"
+if git rev-parse --verify --quiet "${diff_base}^{commit}" >/dev/null; then
+    git diff --name-only "${diff_base}" HEAD 2>/dev/null \
```
michaelmckinsey1: Two-dot diff reports every file that differs between develop's tip and the PR head, including files changed on develop after the branch point. So application_spec_changed / experiment_spec_changed / system_spec_changed will fire for commits the PR never touched, and get worse the longer the branch is open. Use "${diff_base}...HEAD" or an explicit git merge-base

^ claude

.gitlab/utils/performance/push-to-perf-repo.sh
```diff
@@ -0,0 +1,64 @@
...
+
+if [[ ! -d "${summary_dir}" ]]; then
+    echo "Missing nightly metadata directory: ${summary_dir}" >&2
+    exit 1
+fi
+
+if [[ -f "${BENCHPARK_PERF_DEPLOY_TOKEN}" ]]; then
+    cp "${BENCHPARK_PERF_DEPLOY_TOKEN}" "${ssh_key_file}"
+else
+    printf '%s\n' "${BENCHPARK_PERF_DEPLOY_TOKEN}" | tr -d '\r' > "${ssh_key_file}"
```
michaelmckinsey1, line 29: This is sketchy that this token would be readable in /tmp, but I checked and it should write to `/var/tmp/benchpark/` so only readable to the benchpark user. So shouldn't be an issue in that regard.

michaelmckinsey1 (changes requested):
I'm wondering if we should restructure the `utils/` dir with all of these new files... Perhaps files related to (1) job control (2) githash (3) publishing to the repo should be separated into different directories.

.gitlab/utils/generate_test_summary_csv.py
joshuamvining: This script was deprecated because it generated a CSV that was used to generate a very simple table describing only passing and failure based on a CSV file. The new script uses a .json file to generate a much more descriptive table.

.gitlab-ci.yml
joshuamvining: addressed in recent commits

.gitlab/utils/fetch-job-artifact.sh
joshuamvining: addressed in recent commits

.gitlab/utils/write-test-metadata.sh
joshuamvining: addressed in recent commits

michaelmckinsey1:
@joshuamvining lint File '/home/runner/work/benchpark/benchpark/docs/ci-developer-guide.rst' could be reformatted.

michaelmckinsey1:
~I think I'm ok with the changes now, but I want @slabasan opinion on https://github.com/llnl/benchpark/pull/1479#pullrequestreview-5216995435 in regards to the structure of the files in the utils directory. It has grown to be a lot of files, some unrelated.~

Looks like @joshuamvining proposed a new structure, we talked about it and I think the structure makes sense.

slabasan (approved):

PR #1484, opened 2026-08-12
define missing execute-gen in ramble

slabasan:
## Description

Earlier gromacs PR (https://github.com/llnl/benchpark/pull/835) switched the default experiment for gromacs to water_gmx50. It uses a ramble executable execute-gen, which is not defined in the application.py. The default configuration of the gromacs experiment won't build, because ramble will fail.

This PR defines an executable for execute-gen.


pearce8 (approved):

pearce8 (changes requested):
Gitlab CI seems to not be running again...

pearce8 (approved):

PR #1486, opened 2026-08-17
Update System Docs Page Update

matthewc2003:
Original content of this page is from the `add-a-system-config` page. So ignore left side diff, this is essentially a new file.

Update body text defining how to use the system software stack repair aspect of HPCTESTS paper


docs/update-system-config.rst
```diff
@@ -8,535 +8,93 @@
...
-create a ``system.py`` are:
-
-- :ref:`creating-sys-class`
-- :ref:`class-init-and-resources` - At least one cluster must be defined.
-- :ref:`compiler-def` - At least one compiler must be defined.
-- :ref:`software-section`
-
-.. _creating-sys-class:
+***************************
+ Check an Existing System
```
michaelmckinsey1:
```suggestion
 1. Compare the Existing Software Specification to the Current Environment
```

docs/update-system-config.rst
```diff
@@ -8,535 +8,93 @@
...
-    need to define a virtual package ``"mpi": {"buildable": False},``, followed by a
-    definition of at least one provider of this package (see the provider definition for
-    ``openmpi`` in our example). This is to ensure Spack uses the provider we specified,
-    and does not try to build another MPI package. See a similar example for ``blas``,
-    ``lapack``, and their provider ``atlas``.
+    ./bin/benchpark system external llnl-elcapitan cluster=tioga --propose
 
-::
+************************
+ Save Proposed Changes
```
michaelmckinsey1:
```suggestion
 2. Save Proposed Changes to a Patch File
```

docs/update-system-config.rst
```diff
@@ -8,535 +8,93 @@
...
+    benchpark system external llnl-elcapitan cluster=tioga --write-proposed-system=system.py.proposed
 
-And then whether or not you have packages, run ``benchpark system external <system>
-cluster=<cluster>``:
+These options generate proposed changes without modifying the existing
+Benchpark system definition.
 
-::
+*********************
+ Apply Local Changes
```
michaelmckinsey1:
```suggestion
 3. Apply Local Changes to the System
```

docs/update-system-config.rst
```diff
@@ -8,535 +8,93 @@
...
 
-    variant(
-        "queue",
-        default="none",
-        values=("none", "pbatch", "pdebug"),
-        multi=False,
-        description="Submit to named queue",
-    )
+************************
+ Create a Pull Request
```
michaelmckinsey1:
```suggestion
 4. Upstream Changes to the Benchpark Repository
```

docs/update-system-config.rst
```diff
@@ -8,535 +8,93 @@
...
 ###################
 
-This guide is intended for those who would like to add a new system to benchpark, such
-as vendors, system administrators, or application developers. Benchpark provides an API
-for representing system specifications as objects and options to customize the
-specification on the command line. System specifications are defined in ``system.py``
-files located in the systems directory: ``benchpark/systems/<system>/``.
+These steps show how to use ``benchpark system external`` to compare the
+external packages defined by a Benchpark system against packages detected on
+the current machine.
```
michaelmckinsey1:
```suggestion
the current machine. For adding a new system, please see :ref:`add-a-system-config`

.. note::

    You must run the following commands on the target system. Software is detected for the current environment.

```

docs/update-system-config.rst
```diff
@@ -8,535 +8,93 @@
...
-instances that share this same hardware and software specification using the
-``instance_type`` variant.
-
-.. note::
-
-    Most system classes in Benchpark have a similar concept, but often they refer to
-    physical (named) clusters with very-similar configs, and so they typically use the
-    term "cluster" rather than "instance_type".
+Run ``benchpark system external`` with the same system variants that would normally be
+passed to ``benchpar system init``. For example, to check the Tioga configuration of
```
michaelmckinsey1:
```suggestion
passed to ``benchpark system init``. For example, to check the Tioga configuration of
```

docs/update-system-config.rst
```diff
@@ -8,535 +8,93 @@
  Updating a System
```
michaelmckinsey1:
```suggestion
 Updating the System Software Specification
```

docs/update-system-config.rst
```diff
@@ -8,535 +8,93 @@
...
 ###################
 
-This guide is intended for those who would like to add a new system to benchpark, such
-as vendors, system administrators, or application developers. Benchpark provides an API
-for representing system specifications as objects and options to customize the
-specification on the command line. System specifications are defined in ``system.py``
-files located in the systems directory: ``benchpark/systems/<system>/``.
+These steps show how to use ``benchpark system external`` to compare the
```
michaelmckinsey1:
```suggestion
These steps show how to use ``benchpark system external`` to compare and update the
```

docs/update-system-config.rst
```diff
@@ -8,535 +8,93 @@
...
+    benchpark system external llnl-elcapitan cluster=tioga
 
-When defining ``__init__()`` for our system, we invoke the parent class
-``System::__init__()``, and set important system attributes using the
-``id_to_resources`` dictionary, which contains information for each ``cluster`` or
-``instance_type``. We define common attributes a single time for all ``instance_type``'s
-inside the ``__init__()`` function:
+Benchpark processes the supplied system specification, collects the expected
+package configuration from the system definition, and runs Spack external
+detection for those packages. The detected configuration is then compared
```
michaelmckinsey1:
```suggestion
Benchpark collects the current system software specification, and runs ``spack external find`` to detect new versions for each package.
The detected configuration is then compared
```

docs/update-system-config.rst
```diff
@@ -8,535 +8,93 @@
...
+against the configuration defined by Benchpark.
```
michaelmckinsey1:
```suggestion
against the current configuration defined by Benchpark.
```

docs/update-system-config.rst
```diff
@@ -8,535 +8,93 @@
...
-
-.. note::
-
-    For packages that declare virtual dependencies, e.g., ``depends_on("mpi")``, you
-    need to define a virtual package ``"mpi": {"buildable": False},``, followed by a
-    definition of at least one provider of this package (see the provider definition for
-    ``openmpi`` in our example). This is to ensure Spack uses the provider we specified,
-    and does not try to build another MPI package. See a similar example for ``blas``,
-    ``lapack``, and their provider ``atlas``.
+    ./bin/benchpark system external llnl-elcapitan cluster=tioga --propose
```
michaelmckinsey1:
```suggestion
    benchpark system external llnl-elcapitan cluster=tioga --propose
```

michaelmckinsey1: Also the line above this says "additional variants can be supplied", but this line is identical to the previous benchpark system external line.

docs/update-system-config.rst
```diff
@@ -8,535 +8,93 @@
...
 
-.. _software-section:
-
-4. Add a Software Section
-=========================
-
-Here we define the ``compute_software_section()``, where at minimum we must define the
-``default-compiler`` for Ramble. This is trivial for the single compiler that we have,
-``gcc@11.4.0``.
+Additional system variants can be supplied in the same way. For example:
```
michaelmckinsey1:
```suggestion
Additional system variants can be specified in the same way. For example:
```

docs/update-system-config.rst
michaelmckinsey1: More detail would be nice after this like

```
This proposed changes will be written to standard output. If you want to save these changes to a file, proceed to the next step.
```

docs/update-system-config.rst
```diff
@@ -8,535 +8,93 @@
...
-an empty dictionary:
+Alternatively, write a complete proposed copy of the updated ``system.py``:
 
 ::
 
-    def compute_packages_section(self):
-        return {
-            "packages": {}
-        }
+    benchpark system external llnl-elcapitan cluster=tioga --write-proposed-system=system.py.proposed
```
michaelmckinsey1: So `--patch-file` creates a git patch file, where `--write-proposed-system` creates a patch file and applies it in one step and saves that to a new copy of my `system.py`, named `system.py.proposed`? Can you write and explain in this detail, because this is the type of question people will have.

docs/update-system-config.rst
```diff
@@ -8,535 +8,93 @@
...
-modules:
+After reviewing the proposed updates, use ``--apply`` to apply changes
+directly to the system's ``system.py``:
 
 ::
 
-    [dane6:~]$ module display gcc/12.1.1
-    ...
-    prepend_path("PATH","/usr/tce/packages/gcc/gcc-12.1.1/bin")
+    benchpark system external llnl-elcapitan cluster=tioga --apply
```
michaelmckinsey1, line 61: Does `--apply` apply the patch file I just made with `--patch-file=tioga-externals.patch`? What if I have generated multiple different patch files with different names? If it creates a new file and applies it, it seems redundant to `--write-proposed-system`.

If I am mistaken about the redundancy, then the difference between the "apply" and "save" sections should be elaborated.

docs/update-system-config.rst
```diff
@@ -8,535 +8,93 @@
...
-    "queues": [JobQueue("pdebug", 60, 1), JobQueue("pbatch", 1440, 28)],
+    benchpark system external llnl-elcapitan cluster=tioga --apply --pr
 
-where for each available entry in your variant, you must add a ``JobQueue`` object where
-you specify (1) the name, (2) the maximum time limit of that queue, and (3) the maximum
-number of nodes you can allocate at one time in that queue.
+Benchpark first applies the safe updates to the system definition. It then
+creates a Git branch for the update, commits the modified ``system.py``, pushes
+the branch to the configured remote, and opens a pull request against the
+Benchpark repository.
```
michaelmckinsey1: Unless this is already implemented, I think it would be fine to omit this feature and expect the used can create a git branch and make a pull request. IMO.

michaelmckinsey1 (changes requested):
I think this may be too concise in places, i.e. maybe confusing for someone who isn't familiar with Benchpark. See the "Adding a System" page as a reference. This doesn't need that level of detail, but assume someone reading this page may know little about Benchpark, so we must describe things that we may think are obvious.

slabasan:
@matthewc2003 Why is the ics.rst file in this commit? I think that may have been committed in error? If so, can you please remove it?

michaelmckinsey1:
> @matthewc2003 Why is the ics.rst file in this commit? I think that may have been committed in error? If so, can you please remove it?

I think that file was breaking lint at some point, and it was excluded from lint instead of fixing the linting https://github.com/llnl/benchpark/commit/18739b131020ece0e14f2e4b78e72fa74b0ea58c. So when matthew ran lint manually without `--exclude docs/ics.rst` it tried to fix lint, which is the changes you see here.

matthewc2003:
> @matthewc2003 Why is the ics.rst file in this commit? I think that may have been committed in error? If so, can you please remove it?

@slabasan yes the file was breaking lint, so I committed it, should I remove it?

slabasan:
Yes please remove the file from this PR as it is outside the scope. If you want to fix the lint in ics.rst, let's do it in a separate PR and update the lint command to not ignore the file (lint is run automatically as part of the github workflows).

slabasan (approved):

PR #1455, opened 2026-07-10
System External Functionality Update

matthewc2003:
## Description

extend the benchpark system external command functionality for issue #1440

PR #1494, opened 2026-08-25
add extracted system software stack repair

matthewc2003:
## Description

Adds the ability to discover the system software stack and update declared system spec.


slabasan:
Supersedes #1455

